PR #91, opened 2018-06-15
Font alignment selector

dnmeid:
Made a picker for the alignment (based on colorpicker)


lib/bank.js
```diff
@@ -20,7 +20,13 @@ function bank(system) {
 				type: 'textinput',
 				id: 'text',
 				label: 'Button text',
-				width: 12
+				width: 10
+			},
+			{
+				type: 'alignmentcontrol',
```
haakonnessjoen: I think the better solution would be to have a type: 'select' or 'dropdown'. And then give the options as options: [ { id: xx, value: 'Left' }, ... ]

PR #419, opened 2018-12-27
Minor: Action delays can be absolute or relative. Can also be reordered.

SteveTherrien:
Added a new option for all buttons whether to make the delay timing absolute or relative. The default is still absolute. When delays are relative, a preceding actions' delay is added to all subsequent actions' delays.

Actions can also be reordered by dragging an actions' label up/down in the table.

Example
- Action 1. Delay 1000 ms
- Action 2. Delay 2000 ms
- Action 3. Delay 0 ms
- Action 4. Delay 1000 ms

With absolute delays:
- Action 3 runs immediately.
- After 1000 ms, actions 1 and 4 run.
- After another 1000 ms, action 2 runs.

With relative delays:
- Action 1 runs after a 1000 ms delay.
- Action 2 runs 2000 ms after after 1.
- Action 3 runs with action 2.
- Action 4 runs 1000 ms after actions 2 and 3.


willosof:
William Viker [3:36 AM]
@Steve Therrien the thing that happens when you start moving cant happen
the extra row over there has to go 🙂 
nope
it makes no sense what so ever
and, maybe we should have a drag-burger in the left side of the table?
and, you should use the up/down cursor, not the up/down/left/right 🙂 
Where do I decide if the delays are going to be absolute or relative?
i would want to do decide that for each action 🙂 
and we should restrict the motion horitzontally, so we're only able to move it vertical

SteveTherrien:
I didn't want to add a new library to Companion so I reused the `touch-dnd.js` library it came with. The extra row is a side effect of that library. It can be disabled, but dragging a row wasn't as intuitive then.

I've replaced the library with a proper row sorter which is a cleaner experience.

I've added a grip next to the labels to show there's something there to drag around. I'm not sure if there's a better way to display the grip. I didn't want to create a new column and the "Delete" event is bound to that table cell, so I settled next to the label. I'm trying to keep my PR as lite as possible.

I added some CSS to clean up the row reordering. This library also restricts the row dragging to the Y axis.

relative-2

Each button has a "Relative Delays" checkbox at the top which toggles between relative and absolute mode. Absolute is the default (as it was before).

relative-3

----

If everything looks good, I'll add a note describing the feature to `gettingstarted.md` before it's merged.

krocheck:
One way I've seen drag handles in other things is to have a 'hamburger' column (attached image I found).  The style of the handle I think is appropriate, but giving it its own column would clean it up IMO.

image

SteveTherrien:
I didn't want to reuse the [bars](https://fontawesome.com/icons/bars?style=solid) icon since it represents the menu. [grip-horizontal](https://fontawesome.com/icons/grip-horizontal?style=solid) didn't feel right, and [grip-lines](https://fontawesome.com/icons/grip-lines?style=solid), which I wanted to use, isn't available in our font-awesome pack.

I wanted to avoid adding a new column if it wasn't needed (the UI already feels tight on my screen), but...

relative-4

How does this look?

krocheck:
I think it works better then where it was for the time being.  Good work!

SteveTherrien:
It does look cleaner on the side, but at the expense of losing 32px of horizontal space. It's probably not a concern for most people though.

willosof:
I like the one with arrows, but they should be on the left i think :)

SteveTherrien:
Sure thing:

relative-5

SteveTherrien:
Awesome! Thanks

PR #421, opened 2018-12-31
Interface instead of IP and port

Mariu5:
I think it is more consistent to use the interface instead of IP. it's more like in the GUI version and it's DHCP compatible.

### DIFF
--- a/headless.js
+++ b/headless.js
@@ -23,13 +23,11 @@ if (process.argv.length < 3) {
     console.log("Available Interfaces:");
 
     Object.keys(ifaces).forEach(function (ifname) {
-
         ifaces[ifname].forEach(function (iface) {
             if ('IPv4' !== iface.family) {
                 // skip over non-ipv4 addresses for now
                 return;
             }
-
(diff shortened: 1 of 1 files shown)


krocheck:
How, if at all, does this affect interfaces with multiple IP addresses bound on them?  I've done this in the past on Win machines but not necessarily with Companion.

Mariu5:
> How, if at all, does this affect interfaces with multiple IP addresses bound on them? I've done this in the past on Win machines but not necessarily with Companion.

For now, interfaces with multiple IPs are not considered. what's the use case for this? What does the GUI when an interface has more than one IP?

krocheck:
The use-case would be where there's 2 control networks (1 DHCP and the other static OR both static) where you're not VLAN'ing the switch but simply using Layer 3 subnets to segment the networks.  In that case the control computer can be given an IP in both segments on a single interface.

But I checked and the secondary IP binds are currently ignored by interface dropdown anyway, so ignore my post.

PR #432, opened 2019-01-08
Issue 430 Patch

krocheck:
Bypasses the 'Bitfocus Companion' module from being added to modules_name array and therefore searchable in 'Add by search' Interface tab.

Tested.  Properly inhibits the module from 'Add by search' and did not see any errors in the normal execution or background load of the module/instance.


willosof:
👍

PR #506, opened 2019-02-24
Add ability to control button style with osc.

DrewLX:
Along side pressing the button, this adds /style/bgcolor, /style/color and /style/text...

Pretty simple really, but I needed for an event so that QLab could update button labels based on very specific things. It could be argued this should be done as feedback on the OSC module, but I figured this was simpler.

### DIFF
--- a/lib/osc.js
+++ b/lib/osc.js
@@ -62,8 +62,49 @@ function osc(system) {
 				}
 
 			}
+		} else if (message.address.match(/^\/style\/bgcolor\/\d+\/\d+$/)) {
+			if (message.args.length > 2) {
+				var r = message.args[0].value;
+				var g = message.args[1].value;
+				var b = message.args[2].value;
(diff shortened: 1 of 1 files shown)


haakonnessjoen:
If this feature is going to be added, I think that this should be added via a feedback from the internal module.

willosof:
CLA Pending.

DrewLX:
CLA is signed. If I follow the link it shows as being completed.

haakonnessjoen:
Have you added the email used in the commit to your github user account?

DrewLX:
As far as I can see, its all from the same account / email etc.

haakonnessjoen:
The commit was done by "Drew Perry <drew@event.engineering>"

DrewLX:
Ahhh! Sorry, my bad. Should be clear now

PR #508, opened 2019-02-26
Add Dockerfile

LesterCovax:
A Dockerfile and corresponding .dockerignore file are included in this PR.  The resulting docker image has been tested by myself and several people in Slack (https://hub.docker.com/r/lestercovax/companion).    This has also been tested running on a Synology NAS, and an Unraid instance.  I'll eventually create a public Unraid template so it shows up in the Unraid app list.

I can set up auto-builds on my docker hub, or the Bitfocus team can create their own hub account (e.g. bitfocus/companion).


willosof:
Brilliant! :)

Issue #569, opened 2019-03-27
Enhancement: Pin Code unlocks all surfaces option


MrLazerface:
**Describe the feature**
Add a toggle option to the settings page that would allow pin code unlocking to apply to all connected stream decks, or if not toggled on would allow individual stream deck locks (as it currently functions).

**Usecases**
For companion users using multiple stream decks, having a single lock button and using a single stream deck to unlock all connected stream decks would make them operate more like a traditional lighting/audio console lock function.

mfoulks3200:
Feature added in pull #572 , [Pull #4](https://github.com/bitfocus/companion-module-bitfocus-companion/pull/4) in the internal module to add related actions.

PR #572, opened 2019-03-27
Updated lockouts to be linkable

mfoulks3200:
Also added system event listeners to selectively lockout devices, or lockout everything

### DIFF
--- a/lib/device.js
+++ b/lib/device.js
@@ -104,17 +104,21 @@ function device(_system, panel) {
 		}
 	};
 
-	system.on('lockoutall', function () {
+	self.lockoutAll = function() {
 		self.lockedout = true;
 		self.drawPage();
-	});
+	};
 
-	system.on('unlockoutall', function () {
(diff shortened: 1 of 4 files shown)


mfoulks3200:
In reference to issue #569

lib/device.js
```diff
@@ -104,10 +104,26 @@ function device(_system, panel) {
 		}
 	};
 
+	system.on('lockoutall', function () {
```
haakonnessjoen: Everytime you add a listener on system, you need to make sure the function is a reference, and that you unload that reference in the function unload(). If not, this callback will continue to try to run, even after the device is removed. This is the reason for MrLazerface's problems. If you remove a stream deck, and add a new, your callbacks will try to modify a device that should have been removed and garbage-collected. Look at how "elgato_click" has been implemented in this file.

mfoulks3200:
Removed listeners on unload

PR #594, opened 2019-04-06
Build feature update

isaiahwilliams:
Through this pull request the following changes are being made 

1. Added the build definitions for Azure Pipelines. This will make it simpler for those using Azure DevOps to build Companion.
2. Expanded the .gitignore file based on [Node.gitignore](https://raw.githubusercontent.com/github/gitignore/master/Node.gitignore)


JeffreyDavidsz:
This is one for the masters @haakonnessjoen @willosof

willosof:
What's up with the massive .gitignore change?

isaiahwilliams:
I am using Visual Studio for development, and it adds a bunch of random files that should not be committed. So, I updated the .gitingore to ensure that nothing gets add that should not.

PR #685, opened 2019-06-08
New action controls (checkbox and number/range input)

SteveTherrien:
Two new controls available for actions and feedbacks:

new-action-options

## Checkbox
A checkbox with a boolean value of `true` when checked and `false` when unchecked. The `default` property must be a boolean.

The value returned will be a boolean.

```
{
  type: 'checkbox',
  label: 'Invert',
  id: 'invert'
  tooltip: 'Inverts the layer',
  default: false
}
```

## Number
Creates a [number input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) element which will only accept integers between `min` and `max` (inclusive). The `default` property must be a number (or `""` if not required). If the `required` property is `true` then the field can't be left empty.

If the `range` property is `true` then a [range input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) will also be created. Adding a range input will always make the field required since a range always has a value.

The value returned will always be a number (example: `50`, not `"50"`), unless the field isn't required and was left empty, in which case it will be an empty string.

```
{
  type: 'number',
  label: 'Opacity',
  id: 'opacity',
  tooltip: 'Sets the opacity percent (0-100)',
  min: 0,
  max: 100,
  default: 50,
  required: true,
  range: false
}
```
----

PR also includes a bug fix to allow the default color to be set (`action.js` and `release_actions.js`). Default color was previously ignored in favor of the saved value, which didn't yet exist for new actions.

### DIFF
--- a/public/action.js
+++ b/public/action.js
@@ -247,7 +247,7 @@ $(function() {
 
 								$opt_input.spectrum({
 
-									color: int2hex(action.options[f_oid]),
+									color: int2hex($opt_input.val()),
 									preferredFormat: "rgb",
 									showInput: true,
 									showPalette: true,
--- a/public/action.js
+++ b/public/action.js
@@ -60,6 +60,44 @@ $(function() {
 		socket.emit('bank_update_action_option', page, bank,  $(this).data('action-id'), $(this).data('option-id'), $(this).val() );
 	});
 
+	$('#bankActions').on('change', '.action-checkbox', function() {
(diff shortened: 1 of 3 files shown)


SteveTherrien:
Background:

I'm adding some new actions to a module. Having checkbox and range inputs would make it much friendlier to use versus the normal dropdowns and textboxes.

interface

I'm sure others would find these new controls helpful in their modules too.

PR #798, opened 2019-08-31
Minor: add release_action and feedback support to addUpgradeScript

krocheck:
addUpgradeScript does not currently support release_actions or feedback, meaning it is functionally broken.  These should be the necessary fixes, however, I have not had an opportunity to test.  Someone else may be able to get to that before I can and this can be merged once that is done.

### DIFF
--- a/lib/feedback.js
+++ b/lib/feedback.js
@@ -111,7 +111,7 @@ function feedback(system) {
 				for (var i in self.feedbacks[page][bank]) {
 					var feedback = self.feedbacks[page][bank][i];
 					if (feedback.instance == instance_id) {
-						fbs.push(action);
+						fbs.push(feedback);
 					}
 				}
 			}
--- a/lib/feedback.js
+++ b/lib/feedback.js
@@ -104,6 +104,21 @@ function feedback(system) {
 		cb(self.feedbacks);
 	});
 
+	system.on('feedbacks_for_instance', function (instance_id, cb) {
+		var fbs = [];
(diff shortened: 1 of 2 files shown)


lib/feedback.js
```diff
@@ -104,6 +104,21 @@ function feedback(system) {
...
 	});
 
+	system.on('feedbacks_for_instance', function (instance_id, cb) {
+		var fbs = [];
+		for (var page in self.feedbacks) {
+			for (var bank in self.feedbacks[page]) {
+				for (var i in self.feedbacks[page][bank]) {
+					var feedback = self.feedbacks[page][bank][i];
+					if (feedback.instance == instance_id) {
+						fbs.push(action);
```
haakonnessjoen: push action?

lib/instance_skel.js
```diff
@@ -126,10 +126,18 @@ instance.prototype.upgradeConfig = function () {
 		self.system.emit('actions_for_instance', self.id, function (_actions) {
 			actions = _actions;
 		});
+		var release_actions = [];
+		self.system.emit('actions_for_instance', self.id, function (_release_actions) {
```
haakonnessjoen: same list? actions_for_instance

lib/feedback.js
krocheck: Yep ... that one I missed.

lib/instance_skel.js
krocheck: That one too

Issue #838, opened 2019-09-25
Backport (1.4) needs Linux usb support update


istnv:
backport has the same issue fixed in #677  
This affects all streamdeck models in linux
After upgrading from 1.3 to 1.4, companion lost access to the Stream decks.

josephdadams:
Can we close this? Looks like the fix was added to backport.

PR #839, opened 2019-09-27
libusb update for linux on backport

istnv:
2 line fix for #838


willosof:
This should probably go to master instead of backport :) 1.4.0 is released.

istnv:
This is a 'backport' of the fix that is already in main. Backport doesn't work on linux without it.

PR #844, opened 2019-10-01
Add OSC Commands to doc

cxselph:
OSC remote control commands


haakonnessjoen:
When creating stuff in your own repo, you should always use a branch. Not work on master. We don't want all the unnecessary merges etc.

The best way is to keep your master a clean copy of our master. And each time you work on something new, create a new branch for it, and pull from that branch.

cxselph:
That makes sense. Then do the pull request from that branch straight to the upstream?

haakonnessjoen:
Yes. That way, you can also have several pull requests up. And continue to do changes in the repo, that will also automatically update the pull request. In case people ask for changes to that pull request. 👍

PR #850, opened 2019-10-03
Minor: Add autocomplete for variables

peschuster:
This adds autocomplete for variables to the button text field using the "[tributejs](https://github.com/zurb/tribute)" package.

Preview:
variables_autocomplete

Summary:
- opens list of variables after entering `$(`
- list is filtered during typing
- use suggestions with `Enter`or by clicking on the item in the list

### DIFF
--- a/public/bank.js
+++ b/public/bank.js
@@ -82,18 +82,46 @@ $(function() {
 		},
 	});
 
+	function build_autocomplete_item(instance, item) {
+		var variable_name = instance + ':' + item.name;
+		return { key: variable_name + ')', value: variable_name, label: item.label };
+	}
+
+	// Listen on initial variable definitions broadcast
 	socket.on('variable_instance_definitions_get:result', function (err, data) {
 		if (data) {
 			var auto_complete_list = [];
 			for (var instance in data) {
 				for (var index in data[instance]) {
-					var item = data[instance][index];
-					var variable_name = instance + ':' + item.name;
-					auto_complete_list.push({ key: variable_name + ')', value: variable_name, label: item.label });
+					var item = build_autocomplete_item(instance, data[instance][index]);
+					auto_complete_list.push(item);
 				}
 			}
 
 			variables_autocomplete.append(0, auto_complete_list, true);
+		}
+	});
+
+	// Listen on instance enable/disable events
+	socket.on('variable_instance_definitions_set', function (instance, data) {
+		if (instance && data) {
+			var auto_complete_list = variables_autocomplete.collection[0].values;
+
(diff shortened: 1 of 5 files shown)


peschuster:
The tribute library does not support multi-character triggers right now. There is a pull request pending (https://github.com/zurb/tribute/pull/230), but I am not sure if/when this is getting merged.

I got it working with some additional modifications, but to use this I would have to commit the files from the library directly to the companion repository without using the package as dependency and we would need to switch to the official release at some point later. I could do that, if it is ok with you.

haakonnessjoen:
Instead of including the livrary source inside companion, you could just add the module by your github fork, instead of from npm. Type; yarn add github:user/repo.

What do you think? And when/if they merge the pull and publish, we can change back to the npm one.

peschuster:
Ok, just did that. Now the suggestions menu open after typing "$(" and not with just "$".

haakonnessjoen:
I've tested it and it works great. Except that if I add a new instance, I need to reload the browser to get the new variable definitions.

peschuster:
Good catch. It should be fixed now.

public/bank.js
```diff
@@ -82,21 +82,49 @@ $(function() {
...
 	});
 
+	// Listen on instance enable/disable events
+	socket.on('variable_instance_definitions_set', function (instance, data) {
+		if (instance && data) {
+			var auto_complete_list = variables_autocomplete.collection[0].values;
+
+			// remove existing variables of instance
+			var index = 0;
+			while (index < auto_complete_list.length) {
```
haakonnessjoen, line 112: Weird way of writing for(var index = 0; index < autocoplete_list.length; ++i) 😝

haakonnessjoen:
Thanks! <3

public/bank.js
peschuster: I am not sure how a list in javascript behaves, when deleting items while iterating over it. That's my solution to be on the safe side.

haakonnessjoen: the for loop I used as an example does exactly the same as your version. Remember that a `for` instruction is just 3 expressions.

So I would have done:
```
for (var i = 0; i < auto_complete_list.length; ++i) {
  if (auto_complete_list[i].value.startsWith(instance + ':')) {
    auto_complete_list.splice(i--, 1);
  }
}
```

In regards to how js behaves, the while() and the for() will both read auto_complete_list.length for each iteration, so it is just up to you to keep the index variable correct. So we both make sure that the variable is not increased when you delete :)

Just a tip!

willosof:
Thanks dude!! :D

Issue #836, opened 2019-09-25
Backport - Launch uncaught error - windows


McHauge:
**Describe the bug**
when trying to compile the backport branch, it compiles with no errors, but when booting it up this error shows up, and if you press OK, then there is no modules available in the interface.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to the backport branch
2. Compile on Windows via WSL
3. Open the newly build companion.exe
5. Open the control interface, and find that you can't add modules

**Screenshots**
![image](https://user-images.githubusercontent.com/31018035/65607564-0e136880-dfad-11e9-90e8-d113783193cf.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - WSL 

**Additional context**
also, this doesn't seem right:
![image (1)](https://user-images.githubusercontent.com/31018035/65607893-96920900-dfad-11e9-9bce-fc0a5bb787c6.png)

McHauge:
Hmm, gets a new error massage now, but still no modules. 
![image](https://user-images.githubusercontent.com/31018035/66298199-848b6100-e8f1-11e9-9db8-2b10d68144fd.png)

istnv:
That new error is specific to an update from me. My development 1.4.0 branch seems to have newer node libraries than a bare 1.4.0 build. I added a pull request to wrap that function with a check that function exists.
Interesting that the first error also refers to the node-hid library.

McHauge:
Nice, but now I got this error instead:
![image](https://user-images.githubusercontent.com/31018035/66467929-956cdb80-ea85-11e9-8454-c7aad8df164e.png)
![image](https://user-images.githubusercontent.com/31018035/66468866-2abc9f80-ea87-11e9-8f15-50fd3fd7a761.png)

But I will just try deleting and re-cloning it to test with a clean setup and see if that might help in on my side.

McHauge:
Nope still happening, at least for me: 
![image](https://user-images.githubusercontent.com/31018035/66475294-49c12e80-ea93-11e9-8578-d3b1cf466711.png)

willosof:
Can you guys please check if this is still the case on 2.0? http://builds.bitfocus.io

McHauge:
I got no problems compiling and running V2.0 only the 1.4 backport

willosof:
Aha, so this isn't a problem with the official 1.4 release, only when you make your own 1.4 builds?

McHauge:
Correct

PR #854, opened 2019-10-07
Check if HID function is available before using.

istnv:
This should fix the latest windows error mentioned in #836 
My development 1.4.0 branch seems to have newer node libraries than a bare 1.4.0 build.
linux needs the first patch to run with newer libraries, this one checks if the function exists first.
That function is available only in newer libraries that need it.

### DIFF
--- a/lib/elgato_dm.js
+++ b/lib/elgato_dm.js
@@ -26,7 +26,9 @@ var satelliteDevice = require('./satellite_device');
 var preview = require('./preview');
 var system;
 
-HID.setDriverType('libusb');
+if (typeof HID.setDriverType === 'function'){
+	HID.setDriverType('libusb');
+}
 
 debug("module required");
 


JeffreyDavidsz:
Do we need this patch in 2.0 also?

istnv:
It's not necessary in 2.0 since the newer node-hid library is included/required in 2.0.
My working environment seems to be 1.4 with 'newer' libraries. Most likely an npm update while in my 1.4 workspace. My earlier patch copied the patch from 2.0, but since 1.4 doesn't require the new library, it broke the backport build.
This patch allows 1.4 to work with a fresh clone or after a node library update.
Major apologies for that. Still learning what's in which release.
I now have a clean 1.4 directory to test for any excess dependencies before suggesting fixes.

JeffreyDavidsz:
no problem, tnx!

PR #926, opened 2019-12-13
fix: upgradeConfig run at every startup. run with no feedbacks

Julusian:
A couple of bugs here with the upgrade config scripts.

1) No feedback was being passed into the scripts, as feedback properties are named differently to actions, and the loop to find them was using the wrong name. 

1) Every restart of companion was running the same scripts over again, as the change to _configIdx was not explicitly being saved

### DIFF
--- a/lib/feedback.js
+++ b/lib/feedback.js
@@ -110,7 +110,7 @@ function feedback(system) {
 			for (var bank in self.feedbacks[page]) {
 				for (var i in self.feedbacks[page][bank]) {
 					var feedback = self.feedbacks[page][bank][i];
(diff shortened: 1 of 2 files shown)


krocheck:
@Julusian I assume there's testing completed on this?

Julusian:
@krocheck I wrote this fix while writing an upgrade script for the atem module, so I have tested it there.
I could also see the `bitfocus-companion` module stuck in the same loop of running a script

krocheck:
Oh fun!  Sounds good.

Julusian:
Thanks :)

Issue #595, opened 2019-04-08
"Set surface to page" action triggers "up action" on wrong button


SteveTherrien:
**Describe the bug**
I don't know if this is related to #587, but `internal: Set surface to page` triggers "key up" on the destination page's button.

**To Reproduce**
- Create a button (1.1) with the down action of `internal: Set surface to page` of the current device to go to page 2.
- Press button 1.1
- The log will show that button 1.1 is pressed, but 2.1 is released
```
07. 18:49:57 device(...): Button 2.1 released
07. 18:49:57 device(...): Button 1.1 pressed
```

This will trigger the key up action of button 2.1 **and** also the key up action of button 1.1 even though 1.1's up action isn't logged.

**Expected behavior**
Only the key up action of button pressed (1.1) will be triggered.

**Desktop (please complete the following information):**
 - OS: macOS 10.14.4
 - Browser: Chrome
 - Version 73

**Additional context**
Problems happens both with a Stream Deck attached and just with the emulator, but not in the tablet view.

See attached config: [bug.companionconfig.txt](https://github.com/bitfocus/companion/files/3052400/bug.companionconfig.txt)

- Button 1.1
  - Down action: `Set surface to page 2`
  - Up action: Set button 2.2 BG color to red
- Button 2.1
  - Up action: Set button 2.3 BG color to blue

Pressing 1.1 sets 2.2 to red and 2.3 to blue.

officialmattsnyder:
I'm also experiencing this. Although, I'm not experiencing the "key up" action on the original button that Steve is.
 So...
If button 1.1's down action is "Set surface to page 2"
Pressing that button results in: 

button 1.1 pressed
button 2.1 released

and button 1.1 shows as being "latched" and cannot be unlatched.

Basically, the "set surface to page" renders that button un-releasable.

SteveTherrien:
That does sound related @officialmattsnyder.

It turns out this problem was foreseen when the `set_page` action was [first developed](https://github.com/bitfocus/companion-module-bitfocus-companion/blob/d291bdd50817a5b03ffd97186122ea29d619a139/index.js#L443):
```
 // TODO: Somehow handle the futile "action_release" of the same button on the new page 
```

`set_page` knows the default key-up event will be run against the button on the wrong page, so it creates a second key-up event against the original button. Both key-up events then run.

The cleanest way I've been able to work around this is to make the "set_page" action issue a new event which tells the surface to disregard the next specific key press (bank/key/surface/direction combination).

[...]

Issue #596, opened 2019-04-08
Button may be drawn on incorrect page after "Set surface to page" action


SteveTherrien:
**Describe the bug**
Setting a button's action to change to a different page and also change a button's style (background or text color) in the current page causes that button to be drawn on the wrong page.

**To Reproduce**
1. Set two actions for button 1.1:
   - `internal: Set surface to page` to page 2
   - `internal: Button Background Color` of 1.2 to red
2. Press button 1.1

Page will change to page 2, but button 1.2 will be drawn in place of button 2.2.

Reordering the button's actions works as expected.

**Expected behavior**
Button 2.2 will be drawn normally.

**Screenshots**
![button actions](https://user-images.githubusercontent.com/9618303/55695982-da497580-596f-11e9-9344-3969e08fb035.png)

![bug](https://user-images.githubusercontent.com/9618303/55696149-89864c80-5970-11e9-81ec-b08ad11608df.gif)

**Desktop (please complete the following information):**
 - OS: macOS 10.14.4
 - Browser Chrome
 - Version 73

josephdadams:
@SteveTherrien are you still having this problem? (Just trying to rule out stale bugs.) Thanks!

Julusian:
@josephdadams I suspect this is a duplicate of https://github.com/bitfocus/companion/issues/729 but with a different STR

josephdadams:
#596 (this one, #729, #595, and #769 all sound a lot alike...

istnv:
> #596 (this one, #729, #595, and #769 all sound a lot alike...

They are. The **Internal: Set surface to page** should be fixed and I have a patch coming for the other internal page change actions.

SteveTherrien:
@josephdadams, yes, this problem is still present in HEAD (2020-01-10).

I tried it with @istnv's [PR 953](https://github.com/bitfocus/companion/pull/953) but it didn't make a difference. I'm using the web emulator for my tests (no Stream Deck available at the moment).

Here's a [backup config](https://github.com/bitfocus/companion/files/4049868/issue-596.companionconfig.txt) you can restore to test this locally.

istnv:
That one's different than the others I fixed. The page gets changed before the color change happens. 
If you notice the button number, it is a 'stale' picture issue similar to the others.
Quick fix: swap the order to change background before changing pages.
I shouldn't happen though. I'll try to look at it later today.

Issue #769, opened 2019-08-02
"Set Surface to Page" not functioning correctly. (In both 1.4 and 2.0)


officialmattsnyder:
**Describe the bug**
A "Set Surface to Page" action on key down will trigger the key up action of the corresponding button on the destination page. Set surface to page on key up will result in a Java error which crashes the program.

**To Reproduce**
1. Assign a button (e.g. 1.3) to perform a "Set Surface to Page" action on key down.
2. Assign the same button (e.g. 2.3) on the destination page a key up action.
3. Click on button 1.3 to set the surface to a new page
4. Button 2.3's key up action will trigger.

Alternatively:
1. Assign a button (e.g. 1.3) to perform a "Set Surface to Page" action on key up.
2. Click on button 1.3 to set the surface to a new page.
3. A Java error will occur.

**Expected behavior**
I would expect either of these to result in the surface being set to the specified page with no actions being triggered on the new page.

**Screenshots**
<img width="423" alt="Screen Shot 2019-08-01 at 12 38 08 PM" src="https://user-images.githubusercontent.com/50146619/62389972-304db480-b52f-11e9-9728-a891f341d9db.PNG">

**Desktop (please complete the following information):**
 - OS: MacOS 10.14.5 (also on 10.14.6)
 - Browser: Chrome

**Additional context**
In previous versions (most 1.3 releases) Set Surface to Page on Key Up would function properly. But Set Surface to Page on Key down would leave the example button 1.3 latched and also trigger the up action on the new page. The latching seems to have been fixed, but the triggering on the new page has not.

dnmeid:
I think the solution can be to add a new marker to all banks/buttons. After a page change all buttons are marked as fresh. A key up event for a fresh putton is not operated. A key down event is operated and the fresh marker is cleared. Care should be taken to distinguish between key down/up and bank press/release. Only key up of physical buttons should be ignored, not bank release commands received by OSC or TCP or whatsoever.

josephdadams:
@istnv can we close this also?

istnv:
Yes.

Issue #729, opened 2019-07-04
Buttons aren't redrawn when using "set surface to page" internal action


VanSorgenAudio:
Compamion 2 v1.3.0 (1a7592b-1203)

**Describe the bug**
I made a button (we call this Button 'A1'):
 1ste Action: QLab Start Cue #
 2nd Action: delay 250ms
                     internal: Set surface to page 97

Button as PNG with a custom PNG

Button 'A2': (same position on the desk as 'A1')
 Action: internal: Set surface to page 98

Action Button '1A' works fine. When it goes to the new page the image of Button 'A1' doesn't change to the info of Button 'A2'). Even on a empty button, or on a programmed button. Button 'A2' works fine only the background is not refreshed. It doesn't matter if the new button style is None, Text of PNG.

 - OS: OSX 10.13.6
 - Browser: chrome
 - Versie 75.0.3770.100 (Officiële build) (64-bits)

With regards,
Gijs van Sorgen

haakonnessjoen:
Related or duplicate of #595

josephdadams:
This issue is friends with #595 and #769 :)

PadaKwaak:
I'm also having this issue with 2.0.0-8fe2222-1983 and Stream Deck XL where it does not update the button that was pressed when the button set the surface page.

The same happens for actions:
internal: Increment page number
internal: Decrement page number

Also, is it normal behaviour for the "pagenum" style button to take you to Page 1? In my opinion this button shouldn't be clickable or its actions should be configurable like a regular button

Julusian:
I have noticed this occasionally too.
It feels like a race condition in the ordering of redraws. I would guess this is only reported with the XL due to it having more draw and so the timings behave differently.
This is on a pi4 with a modified version of node-hid and the streamdeck library which improves draw speed (PR incoming once node-hid does a release) so it is less frequent, but still happens.

I can see the following flow of redraws occuring when I press and hold:
On press:
 * Flash of source button (pressed variant)
 * Target button (normal variant)
 * Source button just after the last button has redrawn (normal variant)
On release:
 * Target button (normal variant)

If I dont hold and very quickly release then I think it is simply missing the last redraw.

istnv:
> Also, is it normal behaviour for the "pagenum" style button to take you to Page 1? In my opinion this button shouldn't be clickable or its actions should be configurable like a regular button
The PAGE button has always been goto page 1, like a HOME key.

josephdadams:
[...]

PR #953, opened 2020-01-11
Issue #596 Prevent stale page button graphics

istnv:
This is part of the fix for #596 and friends.
There is another patch for the submodule bitfocus-companion.

Fixed a potential (small) memory creep.


istnv:
I think this also fixes #595 #769 #729

PR #957, opened 2020-01-12
Button gfx timing adjust

### LINKED ISSUE #596: Button may be drawn on incorrect page after "Set surface to page" action

istnv:
Allow button graphics update without drawing on the new page.
Fixes #596

### DIFF
--- a/lib/device.js
+++ b/lib/device.js
@@ -367,9 +367,10 @@ device.prototype.updateBank = function(page, bank) {
 		var img = graphics.getBank(page, bank);
 
 		if (page == self.page) {
-			setImmediate(function () {
-				self.panel.draw(bank - 1, img.buffer);
-			});
+			// 12-Jan-2020: was wrapped with setImmediate
+			// which delays the button draw.
+			// if the page changed, the old button gfx gets put on the new page
+			self.panel.draw(bank - 1, img.buffer);
 		}
 	} else {
 		self.datap = graphics.getImagesForPincode(self.currentpin);


Julusian:
I am still able to reproduce #729 with this applied

istnv:
@Julusian All work fine here (regular streamdeck on linux).
I'm using a test config that is pretty much all overlapping up/down/go/latch/page/internal buttons.
There are other patches that fix button/graphics timings in other routines. All need to be included.

Julusian:
What I am seeing is due to this call into sharp. https://github.com/bitfocus/companion/blob/master/lib/usb/elgato-xl.js#L150

The writes are queued in the correct order (even with one for the previous page discarded as it is queued halfway through the page draw), but sharp is completing the promises in a non-deterministic order causing the trigger button to sometimes be completed after that button from the new page.

Ill open this as a new issue as it is a separate bug, and sharp is already planned to be replaced

PR #988, opened 2020-02-16
Feat/upgrade streamdeck lib

Julusian:
This upgrades the streamdeck connection library, which can bring a noticeable draw performance boost, especially for the xl when changing page.

The library now exposes jpeg encoding options, so I have set some that will have some negative performance impact, but hopefully not much and make the buttons look slightly better.

Finally, I ran [yarn-deduplicate](https://github.com/atlassian/yarn-deduplicate) and there were a bunch of unnecessary dependency versions installed. I generally do this after installing or updating something as it avoid having multiple versions of the same package installed when it is not needed.
eg lodash v4.17.11 and v4.17.15 were both installed when everything would be happy with v4.17.15

I shall test this on my pi4 on monday to confirm it is also better on lower power machines.

### DIFF
--- a/lib/usb/elgato-v2.js
+++ b/lib/usb/elgato-v2.js
@@ -47,7 +47,12 @@ function elgatov2(system, devicepath) {
 	debug('Adding Elgato Streamdeck USB device', devicepath);
 
 	self.devicepath = devicepath;
-	self.streamDeck = new StreamDeck(devicepath);
+	self.streamDeck = new StreamDeck(devicepath, {
+		jpegOptions: {
+			quality: 95,
+			subsampling: 1 // 422
+		}
+	});
 	self.buttonState = [];
 
 	self.info.serialnumber = self.serialnumber = self.streamDeck.device.getDeviceInfo().serialNumber;
--- a/lib/usb/elgato-xl.js
+++ b/lib/usb/elgato-xl.js
@@ -52,7 +52,12 @@ function elgato_xl(system, devicepath) {
 	debug('Adding Elgato Streamdeck XL USB device', devicepath);
 
 	self.devicepath = devicepath;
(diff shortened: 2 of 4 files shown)


haakonnessjoen:
Great stuff! Does the new library still compile on mac, linux and windows?

Julusian:
The streamdeck and jpeg library both have unit tests that run on all 3 platforms and are happy. But I have only done manual testing on linux currently

Julusian:
I have tested on windows and linux with an xl and original, and a pi4 with an xl.
This runs happily and fast on all of those. I don't have easy access to a mac to verify it there, but unit tests say it should run fine.

sakhaUAC:
Getting drastically reduced performance now on page transitions.

build companion-2.0.0-5e2c920-2061
Stream Deck XL 32-key
ATEM Firmware 8.1, ATEM 1 M/E Production Studio 4K
Windows 10 Pro, 1903 (Build 18362.657)

Page transitions used to be essentially instant, now each button gets individually redrawn in sequence and it takes 1-2 seconds to fully transition pages. On previous build companion-2.0.0-8b6d53c-2055 performance was fine.

First page has an image on every button used. The second page has no images on any of the buttons.

performance

Julusian:
Interestingly, I have just checked with my XL, Windows 10 Pro 1809 and the same 2061 build downloaded from the site and drawing is near instant..

sakhaUAC:
I just tried on a different computer with the same pages and have the same issue. Could you try importing these pages to see if there's a hit for you? Both are laptops, one a Thinkpad T470 and the second a T430.

Windows 10 Pro Version 1809 (OS Build 17763.973)

[ExampleCompanionConfig.zip](https://github.com/bitfocus/companion/files/4231612/ExampleCompanionConfig.zip)

Issue #827, opened 2019-09-19
Bug on toggles


darkalester80:
Setting up a toggle to mute channels on my behringer xr18 air using an elgato stream deck. My current issue is when I press the key on the stream deck it isn’t acting as a toggle just a normal key on press and key off on release, but the function works fine on the web layout and on the test function when setting up the button. It even works on the stream deck emulator just fine it toggles on and off with each press just doesn’t do it on an actual stream deck.

josephdadams:
Are you talking about latches?

Also, is the Elgato stream deck software running?

darkalester80:
Yes latches l, and yes my stream deck app is running with the plugin for companion installed on it.

mwsimmonite:
Hi all,

I'm having the same issue, works in the web layout and emulator, but doesn't toggle on the streamdeck itself.

In the debug log, when I push on the emulator it shows the button being pressed and released once, when I push on the stream deck it counts 4 'pressed' followed by 2 'released'??

If there's anything else I can supply to help figure it out let me know.

Thanks
<img width="333" alt="Screenshot 2020-02-18 at 16 19 44" src="https://user-images.githubusercontent.com/44989573/74755072-7ae41380-526a-11ea-8d9c-51ce5ca33a1e.png">

mwsimmonite:
I've tested this some more, and the latch/toggle issue isn't limited to Behringer Xair buttons. Even a button with no function other than to 'latch' doesn't work on the physical stream deck, with the log showing 4 presses and 2 releases. 

I'm currently testing the latest version of 2.0

mwsimmonite:
This works as expected with stream deck XL on current V1.4, with only one button press and release. I'm on MacOSX 10.15.3.

<img width="807" alt="Screenshot 2020-02-18 at 16 38 22" src="https://user-images.githubusercontent.com/44989573/74756812-14acc000-526d-11ea-9cf0-c77237b44374.png">

istnv:
The elgato driver has issues with 2.0. 
That's probably where the multiple press inputs are coming from.
My patches were tested with mutes on my X18, #995 fixes a coding error that slipped through the updates I submitted last month.

haakonnessjoen:
This should be fixed by #995 .
If not, reopen this issue please.

PR #995, opened 2020-02-18
fix: latch button fires down action on second press

Julusian:
the code was almost right, but the return was inside a callback, and needed to be bubbled up a layer

### DIFF
--- a/lib/action.js
+++ b/lib/action.js
@@ -272,6 +272,7 @@ function action(system) {
 				}
 			}
 
+			let reject = false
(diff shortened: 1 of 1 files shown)


mwsimmonite:
Hi Julusian - I'm afraid I'm not experienced in code, but I was wondering if this might be a fix for this issue: https://github.com/bitfocus/companion/issues/827 ? As I say I've no experience but as a shot in the dark what you've described sounded similar to me (and would be great timing!).

Thanks

PR #1018, opened 2020-03-04
Feat/streamdeck sharp scaling queue

### LINKED ISSUE #958: [BUG] Buttons occasionally drawn incorrect on XL and Mini
Related: #729, #596, #595, and #769

**Describe the bug**
Page jump buttons are sometimes not redrawing properly, and are instead staying with the pressed image from the previous page. This case is specific to the XL and mini due to the image scaling, but isnt specific to the page jump case.

I have narrowed the problem down to being caused by the call into sharp at https://github.com/bitfocus/companion/blob/master/lib/usb/elgato-xl.js#L150
The promise results from this are fulfilled in a non-deterministic order, meaning the calls into sharp:
```
sending 1-2 #445

sending 2-1 #446-p
sending 2-2 #447-p
sending 2-3 #448-p
sending 2-4 #449-p
sending 2-5 #450-p
sending 2-6 #451-p

```
can become
```
writing 2-2#447-p

writing 1-2#445

writing 2-4#449-p
writing 2-3#448-p
writing 2-1#446-p
writing 2-5#450-p
writing 2-8#453-p
writing 2-6#451-p

```
and cause the button to show the outdated image.

**To Reproduce**
1) Set button 1.1 to jump to page 2
2) Set button 2.1 to jump to page 1
3) Tap the button and sometimes it will not redraw with the rest

**Expected behavior**

**Screenshots**
See #596

 - OS: Ubuntu 19.10
 - Browser: Chrome
 - Companion Version: 2.0-afa5d3af

**Additional context**

Julusian:
This is the combination of a few improvements to the streamdeck handling code. It might be easier to read through a commit at a time, as I made sure to do it in stages.

I have tested this on an original(v1) and xl. I tried to be careful when refactoring to minimise risk of breaking anything, but it would be good for them to have a quick test before merging.

1) This refactors all the elgato device types to share as much code as possible. It could be done a bit further, but I think this is far enough for now.
This was not strictly necessary, but it made it simpler to do the next step without a lot of copy and paste.
There was also a little bit of tidying some dead code.
Some of the key order logic has been removed as that can now be handled by the streamdeck library, and that keeps the code here a bit cleaner

2) Adds some better error handling. Both by registering an unhandledRejection handler, and by catching errors a bit better in the usb processes. Some are ignored (eg streamDeck.setBrightness), but some will kill the device more consistently (eg streamDeck.fillImage on the xl)
This was prompted by the error @thedist reported having on slack, and hopefully will cause it to remove the device instead of it erroring due to it being in a dead state.

3) For the xl and mini, when the images are scaled by sharp they now have a small work queue. This should fix #958, by ensuring that we are not scaling two images for a single key at a time. In doing so it is also limiting the size of the queue (max 1 per key in the queue), which should be a nice safety net to avoid flooding sharp with more than it can handle


willosof:
Awesome work, man! 👍

PR #1019, opened 2020-03-06
Minor: Add module typescript definitions

Julusian:
This moves the typescript definitions used by the atem module into core. I really should have done this a while ago, but there was no need to as only the atem module consumed them.

It will have no impact on any other module.

These have been left to stabilise for a few months, so should be pretty accurate but there could still be some small errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,7 +75,3 @@ typings/
 
 # next.js build output
 .next
-
-# Ignore atem-hack
-lib/instance_skel.d.ts
-lib/instance_skel_types.d.ts
--- a/lib/instance_skel.d.ts
+++ b/lib/instance_skel.d.ts
@@ -1,0 +1,98 @@
+import {
+  CompanionActions,
+  CompanionInputField,
+  CompanionFeedbacks,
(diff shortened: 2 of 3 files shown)


MeestorX:
Does this mean that it’ll be easier to write other modules in TS now?

Julusian:
@MeestorX Well, it doesn't really change anything as the atem module has been injecting these files at build time (when that module had dependencies installed) for months. So this has been accessible to other modules for ages, this is just a 'tidying' pr to correct the ownership of these files (core instead of atem)

MeestorX:
Ok, maybe someone can do a little "how to" on using TS for modules one day!

Julusian:
@MeestorX yeah, writing up something or creating a boilerplate would probably be a good idea.

Until then, I would make a copy of https://github.com/bitfocus/companion-module-sonos or https://github.com/bitfocus/companion-module-home-assistant (or even https://github.com/bitfocus/companion-module-atem) and work from the framework/structure they provide. The sonos and home-assistant ones were based off the atem module, but are smaller/simpler.

PR #1020, opened 2020-03-06
Minor: Sort instance category/manufacturer lists by name

Julusian:
When adding an instance, sort the category and manufacturer lists by name to make it easier to find things


Julusian:
Hold off, this needs to be made case insensitive

Julusian:
Case sensitivity is fixed.
And the child lists of products is sorted now too.

PR #1021, opened 2020-03-07
Ability to jump to pages by typing page number

bryce-seifert:
-Tested in Chrome, Firefox, and Safari

Open to any suggestions or feedback

### DIFF
--- a/lib/osc.js
+++ b/lib/osc.js
@@ -107,10 +107,12 @@ function osc(system) {
 	};
 
 	self.system.on('osc_send', function(host, port, path, args) {
+	
 		self.listener.send({
 			address: path,
 			args: args
 		}, host, port);
+	
 	});
 
 	/*
--- a/public/bank.js
+++ b/public/bank.js
@@ -875,7 +875,7 @@ $(function() {
 			$(this).val(page);
 		}).keyup(function(e){
 			if(e.keyCode == 13) {
-				var value = $(this).val();
+				var value = parseInt($(this).val(), 10);
 				if (value > 0 && value <100) { 
 					page = value
 					changePage(value);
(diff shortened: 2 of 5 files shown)


krocheck:
What testing has been completed for this PR?

bryce-seifert:
I've tested the following:
–That the page_num input is clickable and editable
–The page_num block shows the correct page number, whether it's toggled to with the buttons or by typing
–The page_num input clears the current page number when clicks to allow for typing a new number
–The page_num input resets to the current page number if a user clicks on the input, but then clicks away without enter a number
–The page_num input warns the user if they enter an invalid page number (not 1-99) or something that's not a number
-The page_num input ignores leading zeros

krocheck:
Did you test that the number updates to current page when clicking the existing navigation buttons?  I think that's what you mean by this:
> –The page_num block shows the correct page number, whether it's toggled to with the buttons or by typing  

There is one reference to .page_curr you missed [here](https://github.com/bitfocus/companion/blob/b18cd11cfb26e39b686ab3e2ca7c506eaadbfe4e/public/import.js#L52) which I believe serves that function.

@willosof any thoughts here?

bryce-seifert:
> Did you test that the number updates to current page when clicking the existing navigation buttons?
Indeed. It does!

> There is one reference to .page_curr you missed here which I believe serves that function.
It seems that what that broke was the page navigation on the Import section, which I forgot about. With the latest commit, I ported the jump to page to that section as well for parity.

krocheck:
Yeah, I don't venture into the `public` directory too much, so I can make educated guesses on what things are doing over there.  I think these changes make sense.

I'm going to let @willosof or @haakonnessjoen give final word on this.

bryce-seifert:
Going to clean this up with proper branching / naming and resubmit.

PR #1027, opened 2020-03-08
Minor: Tidy admin GUI UI elements

bryce-seifert:


JeffreyDavidsz:
Thanks @bryce-seifert

PR #1036, opened 2020-03-16
Added getVariable wrapper to instance_skel

bicheichane:
Saw that there was a setVariable but no getVariable, and other modules just went around the issue by managing their own copy of the dynamic variable states. So I thought why not.

### DIFF
--- a/lib/instance_skel.d.ts
+++ b/lib/instance_skel.d.ts
@@ -67,7 +67,7 @@ declare abstract class InstanceSkel<TConfig> {
   setPresetDefinitions(presets: CompanionPreset[]): void
 
   setVariable(variableId: string, value: string): void
-  getVariable(variableId: string, cb: function): void
+  getVariable(variableId: string, cb: (value: string) => void): void
   checkFeedbacks(feedbackId?: string): void
 
   status(level: null | 0 | 1 | 2, message?: string): void
--- a/lib/instance_skel.d.ts
+++ b/lib/instance_skel.d.ts
@@ -67,6 +67,7 @@ declare abstract class InstanceSkel<TConfig> {
   setPresetDefinitions(presets: CompanionPreset[]): void
 
   setVariable(variableId: string, value: string): void
+  getVariable(variableId: string, cb: function): void
   checkFeedbacks(feedbackId?: string): void
 
   status(level: null | 0 | 1 | 2, message?: string): void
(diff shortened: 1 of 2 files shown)


krocheck:
@Julusian can you validate this for the TypeScript definition.  It looks good otherwise.

Julusian:
hmm, yeah that isnt valid. This will work:
```
getVariable(variableId: string, cb: (value: string) => void): void
```

However, I think it would be better if getVariable simply returned the string instead of using a callback. What do you think @krocheck?

So make the changes be (be warned, untested code)
```
getVariable(variableId: string): string

instance.prototype.getVariable = function(variable) {
	var self = this;

	var result = undefined;
	self.system.emit('variable_get', self.label, variable, function(_result) {
		result = _result;
	});
	return result;
};
```
as that will give the user/module developer a nicer api, and hides the excessive events from them completely

krocheck:
I've not seen direct return from `system.emit`, if thought that was all callback to return since its effectively async.

bicheichane:
I can change no problem, I am a C# dev and have almost 0 experience with js/node. With that said, what do you mean the code isn't valid? It has been tested to work on a module I'm making, so I'm just confused about that.

And I'll definitely test that new code of yours, would much rather avoid the callback if I could, but like krocheck said, didn't think it would be possible to return a value from an emit

Julusian:
@bicheichane the typescript definitions (*.d.ts files) wont cause errors except in the few typescript modules. The module you are making is highly unlikely to be using typescript (I am the only one to have done so currently), so it is expected for you to not see any breakage there. If you are using vscode then you might get some warnings about it, as `function` simply is not valid in that context in the language.

@krocheck While you are right about emit calls being async, scattered throughout the core code is some abuse of knowing that the handler doesnt do anything async.

As an example: https://github.com/bitfocus/companion/blob/master/lib/action.js#L254
which does exactly this flow.

It is horrible, flaky, an abuse of a lot of undocumented behaviour, a bit dangerous and I hate it. But it is a flow which is already being done, so for now I am willing to embrace it, as this case is just as safe/unsafe as the rest of them.

krocheck:
@Julusian I think sticking with the proper method is better, and @bicheichane already tested what he needs against that?

In that scenario @bicheichane the code in `instance_skel.d.ts` should be:

`getVariable(variableId: string, cb: (value: string) => void): void`

PR #1044, opened 2020-03-25
Patch: Clone the dragged preset to show it's being copied not moved.

### LINKED ISSUE #1043: [BUG] Dragging Preset to buttons is visually confusing
**Describe the bug**
When you successfully drag a preset to a button, it "flies back" which in any other program would mean that the item did not drop successfully. A successful drop should "land" on the button and not fly back when the mouse is released. Only if the drop were unsuccessful for some reason should the preset "fly back" to it's start.

**To Reproduce**
Drag a preset to a button. When release your mouse button where you want to drop the preset.

**Expected behavior**
If the drop is successful, the dragging preset image should disappear. It doesn't. It "flies back" as though the drop was unsuccessful.

**Screenshots**
Looks like screen recording can't capture a dragging preset

 - OS: MacOS Mojave, Windows 7
 - Browser: Safari, Firefox
 - Companion Version: 2.0

SteveTherrien:
Clones the preset that's being dragged so it doesn't need to animate back to its starting position after being dropped. Fixes #1043 

issue-1043

### DIFF
--- a/public/preset.js
+++ b/public/preset.js
@@ -73,7 +73,27 @@ $(function() {
 		}
 
 		$presets.append('<br style="clear: both;" />');
-		$presets.find('.presetbank').draggable({});
+		$presets.find('.presetbank').draggable({
+			clone: function() {
+				// Clone the preset that's being dragged instead of dragging the preset itself.
+				// This prevents the preset from animating back to its start position after a drop.
+				// See #1043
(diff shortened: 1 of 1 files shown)


krocheck:
Fixes #1043

MeestorX:
That's fantastic! This has bugged me since I started with companion. Great work!

PR #1045, opened 2020-03-25
Patch: Updated gettingstarted.md for Companion 2.0

SteveTherrien:
Updated gettingstarted.md for Companion 2.0:
- Screenshots from v2
- Updated references to links/buttons that have changed
- Added/updated sections for new things
- Removed the outdated list of modules/versions and instead linked to https://bitfocus.io/support


krocheck:
@SteveTherrien is there an opportunity to make a global help popup button somewhere that calls this ... similar to the module helps?  Drop it top right?

SteveTherrien:
Sure, I could probably work something out.

SteveTherrien:
I've added a new item to the sidebar, "Getting Started", which opens `http://[ip]:8000/help.html` in a new tab and renders the contents of `documentation/gettingstarted.md` using `marked`.

help

I opted for a new tab instead of a modal because I think most people will want to switch back and forth between the interface and the help document. A modal with lots of scrolling makes this difficult.

I've added `lib/help.js` which listens for client messages emitted to `get_help` containing the Markdown filename to-be rendered.

I did this so we have the option on expanding the help system if desired. Example:
- Add `documentation/presets.md`
- Add a help icon to the Presets panel
- Add a click listener for that icon

```js
$presets.on('click', '#preset-help', function() {
	// Will return the contents of documentation/presets.md as HTML
	socket.emit('get_help', { 'file':'presets.md' });
	socket.once('get_help:result', function(resp) {
		// Show modal with contents of resp['html']
	});
});
```

krocheck:
I’m good with this approach. @willsof any thoughts?

krocheck:
I'm good with this approach.  @willosof any thoughts?

willosof:
I haven’t seen it live yet, but if it’s anything like the screenshot I love it :) go!

willosof:
if the menu disappears when entering help, it should
A) Be opened as target="_new"
B) Have a "Back to GUI" button position:fixed or something

SteveTherrien:
@willosof, it opens with target="_new", like the Emulator and Web Buttons.

I just tried the latest MacOS and Windows build versions and gettingstarted.md doesn't load (at least my error handling works 😄). [package.json (lines 40 + 42)](https://github.com/bitfocus/companion/blob/4843defda01eda9c41d90603ae32e17c8de18bbc/package.json#L40) excludes those files during the build process.

Does anyone have concerns about removing those lines from package.json?

PR #1057, opened 2020-03-30
Instance order  - do not merge

JeffreyDavidsz:
Start of instance ordering, DOES NOT WORK YET

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -562,8 +562,8 @@ instance.prototype.connect = function (client) {
 		self.system.emit('instance_delete', id, self.active[id] !== undefined ? self.active[id].label : undefined);
 	});
 
-	client.on('instance_order', function(page, bank, old_index, new_index) {
-		var instances = self.instances[page][bank];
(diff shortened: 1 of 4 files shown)


lib/instance.js
```diff
@@ -563,8 +564,9 @@ instance.prototype.connect = function (client) {
 	});
 
 	client.on('instance_order', function(old_index, new_index) {
-		var instances = self.instances;
+		var instances;
 		if (instances !== undefined) {
```
haakonnessjoen, line 568: This if will never go through, since you just defined instances as undefined :P

krocheck: Funny ... I was just looking through this code last night.  Doing this ends up being more difficult than I expected.  The drag order is using a numeric index for the sort order but ultimately the keys in the array its sorting are the instance IDs.  I haven't found an elegant way to make that all work yet.

JeffreyDavidsz: the same reason why my time shifted to something else ;-)

krocheck: I'm going with punt to v3

PR #1058, opened 2020-04-01
Minor: move option default population to add phase

krocheck:
For actions, release_actions, and feedback.  Removes an async population & save of option defaults from the GUI.  Instead this is done in the add phase in `lib`.

Done for efficiency ... remove extra saves from the overall add phase since a save would be fired with each option as the defaults are loaded.

Tested OK.

### DIFF
--- a/lib/action.js
+++ b/lib/action.js
@@ -501,13 +501,26 @@ function action(system) {
 				if (self.bank_actions[page][bank] === undefined) self.bank_actions[page][bank] = [];
 				var s = action.split(/:/);
 
-				self.bank_actions[page][bank].push({
+				var action = {
 					'id': shortid.generate(),
 					'label': action,
 					'instance': s[0],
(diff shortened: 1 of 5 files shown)


lib/action.js
```diff
@@ -501,13 +501,26 @@ function action(system) {
...
 					'id': shortid.generate(),
 					'label': action,
 					'instance': s[0],
-					'action': s[1]
-				});
+					'action': s[1],
+					'options': {}
+				};
 
+				if (self.actions[s[0]]+':'+[s[1]] !== undefined) {
```
haakonnessjoen: ```':' + [s[1]]```
A string + array of one?

haakonnessjoen: I see it more places down too

haakonnessjoen: I guess these were fixed in the next commit?

PR #1062, opened 2020-04-02
Minor: add feedback hooks (subscribe, unsubscribe)

krocheck:
This PR adds 2 callbacks to the feedback definitions:
- `subscribe`: allows the feedback to send a `SUBSCRIBE` or `NOTIFY` command if it must request to receive updates for the associated data; or a simple `GET` can be sent to warm up the data should the device not have a catch-up method for its feedback data
- `unsubscribe`: allows the feedback to send a, `UNSUBSCRIBE` command should update subscription be required and the device recommends unsubscribing when certain updates are no longer required

These callbacks are invoked via two methods in an instance:
1. Manually via direct calls (outlined below) in the instance class.
2. Automatic `subscribe` and `unsubscribe` firing based on events such as:
    - New feedback (subscribe)
    - Edit feedback (unsubscribe old; subscribe new)
    - Delete feedback (unsubscribe)
    - Preset drop (unsubscribe old; subscribe new)
    - Import page/config (unsubscribe old; subscribe new)

This PR also exposes a new `getAllFeedbacks()` function in `instance_skel` to request all of the active feedbacks from the core.

## Feedback Definition
_ES6 example_
```javascript
initFeedbacks() {
	var feedbacks = {
		input_bg: {
			label: 'Change background color by output',
			description: 'If the input specified is in use by the output specified, change background color of the bank',
			options: [
				this.FG_COLOR_FIELD(this.rgb(0,0,0)),
				this.BG_COLOR_FIELD(this.rgb(255,255,0)),
				this.INPUT_FIELD,
				this.OUTPUT_FIELD
			],
			callback: (feedback, bank) => {
				if (this.getOutput(feedback.options.output).route == feedback.options.input) {
					return {
						color: feedback.options.fg,
						bgcolor: feedback.options.bg
					};
				}
			},
			subscribe: (feedback) => {
				if (this.socket !== undefined && this.socket.connected) {
					this.sendCommand(`SUBSCRIBE Out${feedback.options.output} In${feedback.options.input}`);
				}
			},
			unsubscribe: (feedback) => {
				if (this.socket !== undefined && this.socket.connected) {
					this.sendCommand(`UNSUBSCRIBE Out${feedback.options.output} In${feedback.options.input}`);
				}
			}
		}
	};

	this.setFeedbackDefinitions(feedbacks);
}
```

## Instance Calls
Within a module, calls have been setup to execute these callbacks either across all feedbacks that have them or for a particular type via text parameter:
_ES6 examples_
```javascript
this.subscribeFeedbacks();
this.subscribeFeedbacks('input_bg');

this.unsubscribeFeedbacks();
this.unsubscribeFeedbacks('input_bg');
```
You can also get an array of all user feedbacks in order to index them or fire custom commands:
_ES6 example_
```javascript
processFeedbacks() {
	let feedbacks = this.getAllFeedbacks();

	if (feedbacks.length > 0) {
		for (var x in feedbacks) {
			let feedback = feedbacks[x];
			...
		}
	}
}
```
## Use Cases
A typical module requiring this would execute the subscribe commands after connecting to the device:
_ES6 example_
```javascript
this.socket.on('connect', () => {
	...
	this.subscribeFeedbacks();
	...
});
```

Unsubscribe might be necessary for some devices prior to Companion shutting down or the instance being disabled:
_ES6 example_
```javascript
destroy() {
	if (this.socket !== undefined) {
		this.unsubscribeFeedbacks();
		this.socket.destroy();
	}

	this.debug("destroy", this.id);;
}
```

### DIFF
--- a/lib/feedback.js
+++ b/lib/feedback.js
@@ -181,75 +181,6 @@ function feedback(system) {
 		}
 	});
 
-	system.on('feedback_instance_refresh', function (instance, type) {
-		//debug('Instance ' + instance.label + ' wants us to check banks (' + type + ')');
-		for (var page in self.feedbacks) {
-			for (var bank in self.feedbacks[page]) {
-
-				// Iterate through feedbacks on this bank
-				if (self.feedbacks[page][bank] !== undefined) {
-					for (var i in self.feedbacks[page][bank]) {
-						var feedback = self.feedbacks[page][bank][i];
-
-						if (type !== undefined && feedback.type != type) {
-							continue;
-						}
-
-						if (feedback.instance_id == instance.id) {
-							self.refreshFeedback(feedback);
-						}
-					}
-				}
-			}
-		}
-	});
-
-	system.on('feedback_instance_subscribe', function (instance, type) {
-		//debug('Instance ' + instance.label + ' wants us to check banks (' + type + ')');
-		for (var page in self.feedbacks) {
-			for (var bank in self.feedbacks[page]) {
-
-				// Iterate through feedbacks on this bank
-				if (self.feedbacks[page][bank] !== undefined) {
-					for (var i in self.feedbacks[page][bank]) {
-						var feedback = self.feedbacks[page][bank][i];
-
-						if (type !== undefined && feedback.type != type) {
-							continue;
-						}
-
-						if (feedback.instance_id == instance.id) {
-							self.subscribeFeedback(feedback);
-						}
-					}
-				}
-			}
-		}
-	});
-
-	system.on('feedback_instance_unsubscribe', function (instance, type) {
-		//debug('Instance ' + instance.label + ' wants us to check banks (' + type + ')');
-		for (var page in self.feedbacks) {
-			for (var bank in self.feedbacks[page]) {
-
-				// Iterate through feedbacks on this bank
-				if (self.feedbacks[page][bank] !== undefined) {
-					for (var i in self.feedbacks[page][bank]) {
-						var feedback = self.feedbacks[page][bank][i];
-
-						if (type !== undefined && feedback.type != type) {
-							continue;
-						}
-
-						if (feedback.instance_id == instance.id) {
-							self.unsubscribeFeedback(feedback);
-						}
-					}
-				}
-			}
-		}
-	});
-
 	system.on('feedback_delete', function (page, bank, id) {
 		if(self.feedbacks[page] !== undefined && self.feedbacks[page][bank] !== undefined && self.feedbacks[page][bank][id]!== undefined) {
 			self.unsubscribeFeedback(self.feedbacks[page][bank][id])
@@ -407,20 +338,6 @@ function feedback(system) {
 
 }
 
-feedback.prototype.refreshFeedback = function(feedback) {
-	var self = this;
-
-	if (feedback.type !== undefined && feedback.instance_id !== undefined) {
-		if (self.feedback_definitions[feedback.instance_id] !== undefined && self.feedback_definitions[feedback.instance_id][feedback.type] !== undefined) {
-			let definition = self.feedback_definitions[feedback.instance_id][feedback.type];
-			// Run the subscribe function if needed
-			if (definition !== undefined && definition.subscribe !== undefined && typeof definition.subscribe == 'function') {
-				definition.refresh(feedback);
-			}
-		}
-	}
-};
-
 feedback.prototype.subscribeFeedback = function(feedback) {
 	var self = this;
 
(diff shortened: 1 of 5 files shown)


krocheck:
I can merge, just want other eyes on this.  Thx!

lib/feedback.js
```diff
@@ -311,6 +407,48 @@ function feedback(system) {
...
 }
 
+feedback.prototype.refreshFeedback = function(feedback) {
+	var self = this;
+
+	if (feedback.type !== undefined && feedback.instance_id !== undefined) {
+		if (self.feedback_definitions[feedback.instance_id] !== undefined && self.feedback_definitions[feedback.instance_id][feedback.type] !== undefined) {
+			let definition = self.feedback_definitions[feedback.instance_id][feedback.type];
+			// Run the subscribe function if needed
+			if (definition !== undefined && definition.subscribe !== undefined && typeof definition.subscribe == 'function') {
```
Julusian: Both `definition.subscribe` should be `definition.refresh`

krocheck: Yep ... didn't test with only a refresh definition

Julusian:
Im not sure about the refresh callback. It is only triggered by the module calling `refreshFeedbacks()`, so I think it would be nicer to be done as a simple `this.getAllFeedbacks(type)` that the modules can call. As this will better allow for processing of the feedbacks before doing the refresh and make it easier to do different types of task.
eg, if there are 2 identical feedbacks, the module can more easily do a single `this.sendCommand('GET ...')` call instead of sending the same call twice.

MeestorX:
Let me know if I can help by testing any of this.

krocheck:
Are you suggesting leaving the `refresh` callback, but getting the `getAllFeedbacks(type)` call added to allow more precise execution of it?

Julusian:
I was thinking of removing the `refresh` callback, but I am open to having them both

krocheck:
@Julusian  Reality is, the same could be said for `subscribeFeedbacks(type)` in terms of double-execution.  I could look into adding dup detection for the instance calls?

I see what you're saying about `refresh` since there's nothing in the core doing async fire, its relatively inefficient.

krocheck:
As I'm thinking about it, blanket dup detection would be difficult because how do you define which of the options matching constitutes a dup.  e.g. (button 1) mute on; (button 2) mute off.  The mute parameter could have to be flagged as 'irrelevant' for detection since its the same parameter just two outputs ... where-as an input for 'channel' would need to be flagged as 'important'.  Can of worms.

Julusian:
True, maybe the subscribeFeedbacks and unsubscribeFeedbacks methods on instance_skel shouldnt exist then? It will be very easy for the user to do it manually via getAllFeedbacks.
I dont think doing duplicate detection in core is a good idea, as things like colours will be included in that check, so it will easily miss cases.

krocheck:
One of the main reasons those hooks are in the core is so that as feedbacks are made and changed by the user the core will let those subscribe/unsubscribe events auto-fire.  Otherwise the instances need to hook into feedback save and do their own work to figure out what changed.

krocheck:
Middle-ground, what if `instance_skel` does the `getAllFeebacks()` and runs the firing local?

Julusian:
I am happy with the subscribe and unsubscribe callbacks passed to setFeedbackDefinitions. I think they are good lifecycle notifications for the module.

But its the 'instance calls'
```
this.subscribeFeedbacks();
this.subscribeFeedbacks('input_bg');

this.unsubscribeFeedbacks();
this.unsubscribeFeedbacks('input_bg');
```
that I dont think are needed. As they are called by a module to re-run all the subscriptions. They (and all bits of the refresh) could be replaced by the proposed `getAllFeedbacks` instance call.

krocheck:
Reality is that `unsubscribe` as a general structure is probably not needed ... I more did it for a sense of completeness to the build but I don't know of a piece of gear that needs it.

`this.subscribeFeedbacks()` is not necessarily to 're-run' them as when the module loads a connection will not have been established yet.  This is why my example shows it within a socket on connection call.  This is also why the definition example shows a test for `this.socket.conencted`.  Its certainly a quick and dirty approach, but simpler for most developers to adopt and use vs. a `getAllFeedbacks()` approach and them writing dup detection.  But I think adding that call is needed, for sure, so more complex modules can take that more heavy-handed approach.

We could add an `important: true` flag for the option definitions in order to write dup detection into `instance_skel` ... which tags those as being the parameters to match.  If no important flags are present then you bypass detection for that type.

Julusian:
Good point. With that then I am happy to leave all the bits of subscribe/unsubscribe as is, and suggest to remove all the refresh bits and add the getAllFeedbacks.

Im still not keen on doing duplicate detection internally. I can see it causing problems if getAllFeedbacks de-duplicates but not calls to subscribe and unsubscribe. Or it might just cause confusing bugs if an important flag is not set correctly

krocheck:
Got it.

To clarify, I'm not suggesting `getAllFeedbacks()` do dup detection.  Rather I would move the processing for `subscribeFeedbacks()` into `instance_skel` and it would call `getAllFeedbacks()` to get the data from the core.  We would need to store the feedback definitions locally when `setFeedbackDefinitions(feedbacks)` is called, but probably should already be doing that anyway.

Do I really want to go through adding dup detection at this time?  No, I'd rather wait for v3.  So we can probably just table that.  But I'll remove refresh, add `getAllFeedbacks()`, and try out the firing in `instance_skel`.  Thx!

krocheck:
Lemme know when you're good @Julusian

PR #1063, opened 2020-04-02
Display status text from UNKNOWN

istnv:
This will display the text argument when using `status(STATUS_UNKNOWN,'Something');`.
Before, the text message would be discarded.
It is useful as a progress message: 'Connecting', 'Initializing', etc. aren't errors, but the module is also not in an OK state, either.

### DIFF
--- a/public/instance.js
+++ b/public/instance.js
@@ -74,6 +74,11 @@ $(function() {
 				$("#instance_status_"+x).html("ERROR").attr('title', s[1]).removeClass('instance-status-ok').removeClass('instance-status-warn').removeClass('instance-status-disabled').addClass('instance-status-error')
 			}
 
+			// unknown
+            else if (s[0] === null) {
+                $("#instance_status_"+x).html(""+s[1]).attr('title', s[1]).removeClass('instance-status-ok').removeClass('instance-status-error').removeClass('instance-status-disabled').removeClass('instance-status-warn')
+            }
+
 		}
 	}
 
@@ -696,7 +701,7 @@ $(function() {
 
 				$sm.append($inp);
 			}
-			
+
 			else if (field.type == 'checkbox') {
 				var $opt_checkbox = $("<input type='checkbox' class='form-control instanceConfigField'>");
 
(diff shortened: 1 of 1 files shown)


Julusian:
From what I gather, connecting should be done with the warning status. So I think unknown is only supposed to be used when the instance is disabled.

Reference: https://github.com/bitfocus/companion-module-bmd-atem/issues/48 and https://github.com/bitfocus/companion/blob/master/lib/tcp.js#L41

istnv:
I think using warning is misleading, because the instance hasn't reached a point to determine if there is an error or warning, yet. Adding another 'in-progress' state would be nice.
I used unknown during the connection and startup phase in qlab-advance because until it connects I don't 'know' if there is a need to show a warning or error.
For example, QLab has a startup screen to select a workspace. At that point, the connection will succeed but most of the actions won't do anything, so a 'warning' of "No Workspaces Open" gets displayed.

MeestorX:
@Julusian The main issue is that status(STATUS_UNKNOWN, 'message') doesn't do ANYTHING currently, in terms of displaying something on the instances page. Is there another hook in the core for this message?

krocheck:
Just going to leave here my search for instances of `STATUS_UNKNOWN` being used in the wild in modules.  Some have text associated, but hard to say the uses with the original intent for the state.
image
There are some using the older `STATE_UNKNOWN` constant as well.
image

MeestorX:
Thanks, Keith. But the question remains... does core companion do ANYTHING currently when ```this.status()``` is called, other than show it on the instances tab?

JeffreyDavidsz:
I believe warning in combination with a message will display connecting. Or other stuff. As far as I know, no other module uses the info

willosof:
Companion doesn't do anything with this information other than showing it in the UI. 
About **connecting** not being a WARNING, I must disagree on so many levels. As a practical example; When you're in the middle of a show and the device you're trying to control is "connecting" and not "connected" - that's something that should give you a pretty serious warning in advance :) Everything not connected or connected is a SERIOUS condition, and should be treated and displayed as such. UNKNOWN is intended for devices unable to provide a connection status.

istnv:
I was using this for a connection in progress indicator, since it really is not an error at that point. 
Like an old dial-up modem where 'Dialing' is not an error until the other side does not answer.
We just noticed that the message gets discarded and this would show it.
I can close this and use Warning instead.

willosof:
I just want us all to be on the same page. Unknown should be able to show status as well:)

haakonnessjoen:
It does show a red box in buttons using the instance if the status is error though. Don’t remember if we did orange box on warning, but maybe we should?

MeestorX:
I think Warning is yellow. I use that when waiting for something from the device or user.

PR #1101, opened 2020-05-08
Added YouTube module providing two actions - for setting youtube broadcast live and finished

melimat:
<p>
Right now the YouTube module can do two actions - setting the YouTube broadcast state to live and finished. As far as I know these features are highly requested. 
</p>
<h2>
Brief introduction into usage (what the plugin does)
</h2>
<p>
The user simply creates himself access credentials for YouTube data API for his account in Google API console, downloads JSON file with them from there. Then, when creating instance of the module he provides a path to directory where the JSON is stored. When he applies changes to configuration he is asked to login to his account (OAuth 2 login) in order to access the API. After that, the retrieved access information are stored in token.json file in same directory as previously mentioned JSON. This operation is done in order to not having the user perform the OAuth login every time he Starts companion.
</p>
<p>
After successful  login, the module gets titles and IDs of all YouTube broadcasts and fills the options drop-down menus. User simply selects the title of broadcast on which he wants to perform state update (start/ stop).
</p>
<p>
This is for me the state of minimum viable product, so I would like to continue in development of the module in a future
</p>


josephdadams:
@melimat thanks for the PR! We actually handle modules a little differently than contributions to Companion itself. We set each module up as its own repository and then link it as a sub module within Companion. So we will get that set up for you and ask you to contribute your module code to that repository directly. As the creator and maintainer of that module, you’ll be able to push to it anytime you have an update.

melimat:
Ok, thanks for that. Please accept an apology for this, I should have read the documentatiton more carefully.

josephdadams:
@melimat https://github.com/bitfocus/companion-module-youtube-live
Contribute your module code to this repo directly. Thank you!

PR #1105, opened 2020-05-14
Minor: add actions callbacks to match feedbacks (callback, subscribe, unsubscribe)

Julusian:
Related: https://github.com/bitfocus/companion/pull/1062

This PR adds 3 callbacks to the action definitions, identical to those already on feedback definitions to allow for similar use cases:
- `callback`: execute the action. Takes priority over the instance level `action()` handler if this exists
- `subscribe`: allows the action to send a `SUBSCRIBE` or `NOTIFY` command if it must request to receive updates for the associated data; or a simple `GET` can be sent to warm up the data should the device not have a catch-up method for its action data
- `unsubscribe`: allows the action to send a, `UNSUBSCRIBE` command should update subscription be required and the device recommends unsubscribing when certain updates are no longer required

Note: for these, actions and release_actions are being treated the same. There is no distinction being made, and the difference is hidden from the modules.

These callbacks are invoked via two methods in an instance:
1. Manually via direct calls (outlined below) in the instance class.
2. Automatic `subscribe` and `unsubscribe` firing based on events such as:
    - New action (subscribe)
    - Edit action (unsubscribe old; subscribe new)
    - Delete action (unsubscribe)
    - Preset drop (unsubscribe old; subscribe new)
    - Import page/config (unsubscribe old; subscribe new)

This PR also exposes a new `getAllActions()` function in `instance_skel` to request all of the active action from the core.

## Action Definition
_ES6 example_
```javascript
initActions() {
	var actions = {
		input_bg: {
			label: 'Set output to input',
			options: [
				this.INPUT_FIELD,
				this.OUTPUT_FIELD
			],
			callback: (action, bank) => {
				this.setOutput(action.options.output, action.options.input)
			},
			subscribe: (action) => {
				if (this.socket !== undefined && this.socket.connected) {
					this.sendCommand(`SUBSCRIBE Out${action.options.output} In${action.options.input}`);
				}
			},
			unsubscribe: (action) => {
				if (this.socket !== undefined && this.socket.connected) {
					this.sendCommand(`UNSUBSCRIBE Out${action.options.output} In${action.options.input}`);
				}
			}
		}
	};

	this.setActions(actions);
}
```

## Instance Calls
Within a module, calls have been setup to execute these callbacks either across all actions that have them or for a particular type via text parameter:
_ES6 examples_
```javascript
this.subscribeActions();
this.subscribeActions('input_bg');

this.unsubscribeActions();
this.unsubscribeActions('input_bg');
```
You can also get an array of all user actions in order to index them or fire custom commands:
_ES6 example_
```javascript
processActions() {
	let actions = this.getAllActions();

	if (actions.length > 0) {
		for (var x in actions) {
			let action = actions[x];
			...
		}
	}
}
```

## Use Cases
This will have less use than the feedback versions will, but this allows for some actions to operate based on the current state. Such as toggle, increment or decrement.

A good example, is in the behringer-x32, where the mute action can toggle. This means that if there is a mute set to toggle, the module needs to ensure it knows the state of the mute at all times. 
The x32 doesnt send any data upon connection, any wanted state has to be asked for. one field at a time.
Rather than having to load in the state of anything that might have a toggle action, this allows it to instead only load the ones that are needed

### DIFF
--- a/lib/action.js
+++ b/lib/action.js
@@ -79,7 +79,6 @@ function action(system) {
 	self.system.on('instance_save', function() {
 		setImmediate(function() {
 			self.io.emit('actions', self.actions);
-			self.io.emit('release_actions', self.release_actions);
 		});
 	});
 
@@ -432,10 +431,6 @@ function action(system) {
 				client.emit('actions', self.actions);
 			});
 
-			client.on('release_actions_get', function() {
-				client.emit('release_actions', self.release_actions);
-			});
-
 			client.on('bank_update_action_delay', function(page,bank,action,value) {
 				var bp = self.bank_actions[page][bank];
 				if (bp !== undefined) {
@@ -648,13 +643,6 @@ function action(system) {
 			}
 		}
 		self.system.emit('actions_update');
-		for (var n in self.release_actions) {
-			var x = n.split(/:/);
-			if (x[0] == id) {
-				delete self.release_actions[n];
-			}
-		}
-		self.system.emit('release_actions_update');
 	});
 
 	self.system.on('bank_reset', function (page, bank) {
@@ -680,11 +668,6 @@ function action(system) {
 		self.io.emit('actions', self.actions);
 	});
 
-	self.system.on('release_actions_update', function() {
-		debug('release_actions_update:', self.release_actions);
-		self.io.emit('release_actions', self.release_actions);
-	});
-
 	self.system.on('instance_actions', function(id, actions) {
 		for (var n in actions) {
 			var a = actions[n];
@@ -692,15 +675,6 @@ function action(system) {
 			debug('adding action', id+':'+n);
 		}
 		self.io.emit('actions', self.actions);
-	});
-
-	self.system.on('instance_release_actions', function(id, actions) {
-		for (var n in actions) {
-			var a = actions[n];
-			self.release_actions[id+':'+n] = a;
-			debug('adding release_action', id+':'+n);
-		}
-		self.io.emit('release_actions', self.release_actions);
 	});
 
 	return self;
--- a/lib/action.js
+++ b/lib/action.js
@@ -91,6 +91,7 @@ function action(system) {
 
 						if (action.instance == id) {
 							debug('Deleting action ' + i + ' from button ' + page + '.' + bank);
+							self.unsubscribeAction(self.bank_actions[page][bank][i])
 							self.bank_actions[page][bank].splice(i, 1);
 							self.system.emit('instance_status_check_bank', page, bank);
 							i--;
@@ -107,6 +108,7 @@ function action(system) {
 						var action = self.bank_release_actions[page][bank][i];
 						if (action.instance == id) {
 							debug('Deleting release action ' + i + ' from button ' + page + '.' + bank);
+							self.unsubscribeAction(self.bank_release_actions[page][bank][i])
 							self.bank_release_actions[page][bank].splice(i, 1);
 							self.system.emit('instance_status_check_bank', page, bank);
 							i--;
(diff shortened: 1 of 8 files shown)


krocheck:
The only thing I'm not sure about here is the combining of actions and release actions, not necessarily that its bad for them to be under one call.  I'm wondering if a developer is going to want different conditions for when the action is applied in a release case.  Should there be a "type" field injected in the objects to differentiate?

Julusian:
Im not sure what use case there is for knowing if it is a press or release action. Or if the module should even know.
When an action is being executed there is no way of knowing, so if we were to add something in subscribe then it should probably be added there too, but for what purpose? 
Perhaps adding that type property is something that can be done later when/if there is a use case for it

haakonnessjoen:
I don’t think modules should know if the action was press or rekease. That would just make things less configurable, and the user will not be able to know exactly what happens when an action is called. I am against modules doing different stuff on behalf of a button being pressed or released. It should be up to the user or the preset.

lib/action.js
```diff
@@ -648,17 +677,43 @@ function action(system) {
...
+			for (var i in self.bank_release_actions[page][bank]) {
+				self.subscribeAction(self.bank_release_actions[page][bank][i]);
 			}
 		}
-		self.system.emit('release_actions_update');
 	});
 
+	system.on('action_unsubscribe_bank', function (page, bank) {
+		if (self.bank_actions[page] !== undefined && self.bank_actions[page][bank] !== undefined) {
+			// find all instance-ids in feedbacks for bank
```
haakonnessjoen: feedbacks? :P

lib/action.js
```diff
@@ -648,17 +677,43 @@ function action(system) {
...
+			for (var i in self.bank_actions[page][bank]) {
+				self.unsubscribeAction(self.bank_actions[page][bank][i]);
+			}
+		}
+		if (self.bank_release_actions[page] !== undefined && self.bank_release_actions[page][bank] !== undefined) {
+			// find all instance-ids in feedbacks for bank
```
haakonnessjoen: --- || ----

lib/action.js
```diff
@@ -79,7 +79,6 @@ function action(system) {
 	self.system.on('instance_save', function() {
 		setImmediate(function() {
 			self.io.emit('actions', self.actions);
-			self.io.emit('release_actions', self.release_actions);
```
haakonnessjoen, line 82: Is this not needed?

public/release_action.js
```diff
@@ -30,8 +30,6 @@ function hex2int(hex) {
 }
 
 $(function() {
-	socket.emit('release_actions_get');
```
haakonnessjoen, line 33: Are these sent somewhere else?

lib/action.js
Julusian: Not at all obvious I copied this from the implementation done for feedback :P

lib/action.js
Julusian: My best guess is that originally there was a difference in what could be run between actions and release_actions.
All of the client or io emits removed here had no handlers I could find. I have added,removed and executed release_actions and havent found anything broken from this too

public/release_action.js
krocheck: Technically this should be a `socket.emit('get_actions');` but its hooking by luck the same call from `public/actions.js`.

lib/action.js
krocheck: I came across this when I was dealing with #1058, as I was noticing `self.release_actions` was never initialized like others in the file nor was it ever populated.  Makes sense because the actions and release actions use the same pool.  Mostly seems like the full boar copy/paste ended up with rouge code that was never cleaned up.

PR #1116, opened 2020-06-02
Improved tablet.html using Tailwind CSS

bmeredyk:
This is a new re-work of the tablet.html that I shared through the Facebook group last night (https://www.facebook.com/groups/companion/permalink/2592156324336035/). This time I have rebuilt the page using Tailwind CSS to allow for simpler resizing of buttons at various device sizes. Because I'm adding in a dependency on Tailwind (via CDN link) I am offering this as a _different_ file rather than replacing the current tablet.html but maintainers feel free to merge how you see fit.

I have also improved upon the deck_size=6|15 option so that the number of button columns changes accordingly thus allowing larger buttons and better use of screen space on smaller devices.

Finally, I've updated the way the HTML is built within the Javascript to use template literals making the code far more readable and hopefully easier for others to understand.


JeffreyDavidsz:
Can you download the file and include it in the pull request? Using CDN will not work without internet. Lot of production machines are off the net, so this upgrade won't work for them

bmeredyk:
@JeffreyDavidsz Sure, I'd debated whether or not to include it but decided since it was specifically for use on mobile devices which must then have WiFi they'd generally have internet access too but I suppose it's possible that that isn't 100% true.  I've included tailwind.min.css now. It's a shame to bring in such a large file then only a tiny fraction is actually needed but I'm not familiar enough with the overall build process to try integrating the PostCSS processing in right now. I know the main web UI uses CoreUI (based on Bootstrap 4.1.1) and it might be possible to accomplish something similar with that but implementing with Tailwind was just so simple.

JeffreyDavidsz:
could it be that i'm not seeing the css file in this pull request?

bmeredyk:
@JeffreyDavidsz The css file should be there. It is in my repo and the auto generated github feedback messages on this pull request are mentioning the commits to my forked repo that added it. Unless it is just being misleading and the pull request itself is still for the commit I initiated it at rather than my most current ones?

Issue #474, opened 2019-02-02
Support for RossTalk triggers


monoxane:
**Describe the feature**
To be able to trigger Companion with RossTalk CC bank triggers and have an assignable Next trigger to use with the RollClip button on carbonite surfaces

**Is this platform dependent (windows, mac, ..)?**
No

**If documentation is required to implement, do you know where to find it?**
http://help.rossvideo.com/carbonite-device/Topics/Devices/RossTalk/RossTalk.html#RossTalk

**Usecases**
So people can fire anything from a carbonite surface (Wanna roll QLab off the carbonite surface? Sure!)

JeffreyDavidsz:
I don't have a Ross switcher at hand, but I'm seeing a custom option in the manual (Ross talk controll). Did you try that already? You can send a string TCP to companion to trigger a button. See below the actions at any button for the right syntax

monoxane:
the custom control in the manual is the RossTaslk standard CC string, not an arbitrary tcp string that can be set by the user, its `CC <bank>:<control>` and nothing else, it would be easily worked into companion to trigger the page and button numbers

JeffreyDavidsz:
@monoxane all right, so say you want page 3 button 4 pressed, what is the exact TCP string is being send?

krocheck:
@monoxane you might not be able to answer this, but the 'Generic Device' protocol is stating that custom control is the full string "Custom Control" where-as for all the switchers the actual command string is "CC".  Any sense which is correct here?

krocheck:
Nevermind all ... Ross Support said its the "CC b:cc" format.

monoxane:
> @monoxane all right, so say you want page 3 button 4 pressed, what is the exact TCP string is being send?

for that the exact string being sent from the carbonite would be `CC 3:4` from what I can tell

krocheck:
I had a thread with a Ross tech today and he said it is "CC xyy" where x is the bank (page for Companion) and yy is the number (bank # for Companion).  He said there isn't a semicolon.  The above example would be CC 304.

JeffreyDavidsz:
Ok my idea was to add it to server_tcp so no extra listener connection is needed, agree?

monoxane:
> I had a thread with a Ross tech today and he said it is "CC xyy" where x is the bank (page for Companion) and yy is the number (bank # for Companion). He said there isn't a semicolon. The above example would be CC 304.

Interesting, just a tad different to the way they do the rest of the commands.

[...]

PR #1176, opened 2020-07-24
Rosstalk Listener

onfire4g05:
This allows Companion to listen to RossTalk commands and will recall buttons in Companion. This will resolve issue #474.

The help in the config explains, but basically:

`CC 4:6`

Will press/depress `4.6` in Companion.


krocheck:
I'm OK with this.  It handles press/release and it is off by default.  Appears to follow RossTalk specifications.  Thoughts @haakonnessjoen or @Julusian?

haakonnessjoen:
Maybe our other services aren’t better. But there seems to be a lack of error handling here. (Looking with my mobile phone :p. please arrest me if I am wrong)

That should be added first. But nice idea :)

lib/rosstalk.js
```diff
@@ -0,0 +1,111 @@
...
+				try {
+					this.listen();
+				} catch(e) {
+					console.log('Error listening for artnet', e);
+				}
+			}
+		});
+
+		this.system.on('set_userconfig_key', (key, val) => {
+			if(key !== 'rosstalk_enabled') return;
```
krocheck: This should have curly braces

lib/rosstalk.js
```diff
@@ -0,0 +1,111 @@
...
+			this.config['rosstalk_enabled'] = val;
+
+			try {
+				if(val === true) this.listen();
```
krocheck: Ditto

lib/rosstalk.js
```diff
@@ -0,0 +1,111 @@
...
+
+	process_incomming(data) {
+		// Type, bank/page, CC/bnt number
+		const match = data.match(/(CC) ([0-9]*)\:([0-9]*)/);
+		if(match === null) {
+			this.system.emit('log', 'rosstalk', 'warn', `Invalid incomming RossTalk command: ${data}`);
+			return;
+		}
+
+		if(match[1] === 'CC') this.press_button(match[2], match[3]);
```
krocheck: Ditto

lib/rosstalk.js
```diff
@@ -0,0 +1,111 @@
...
+		this.system.emit('bank_pressed', bank, button, true);
+
+		setTimeout(() => {
+			this.system.emit('bank_pressed', bank, button, false);
+			this.system.emit('log', 'rosstalk', 'info', `Release button ${bank}.${button}`);
+		}, this.release_time);
+	}
+
+	listen() {
+		if(this.active) return; // already listening
```
krocheck: Ditto

lib/rosstalk.js
haakonnessjoen: There should be a space after the «if» codeword also

lib/rosstalk.js
```diff
@@ -0,0 +1,121 @@
...
+ *
+ */
+
+let net = require('net');
+
+class rosstalk {
+	constructor(_system) {
+		this.system = _system;
+		this.active = false;
+		this.socket = this.createSocket();
```
krocheck: Should this have a try/catch?

onfire4g05: It didn't look like `createServer` threw exceptions. It looks like errors are passed to the `error` event. I did just add some error logging, which will then close to port, if there's a problem creating the socket or while it's open. I'm not sure if more is needed.

PR #1193, opened 2020-08-11
Admin Stops Working

### LINKED ISSUE #567: Web interface fails after several days
**Describe the bug**
I leave Companion running 24/7. After about a week, the web interface looks like there is no CSS. When I try to refresh the page, it gives the following error. This has happened on just about every version that has been available.

Error: ENOENT: no such file or directory, open '/var/folders/80/rgd9416d71sbkwzd_yhmgkl80000gn/T/.companion.bitfocus.no.sBN3AJ'

**To Reproduce**
1. Open Companion
2. Wait about a week
3. Try to go to admin interface

**Expected behavior**
I expect it to run forever. :)

**Screenshots**

 - OS: OSX 10.14.3 - Same issue with Sierra and High Sierra
 - Browser: Chrome Version 73.0.3683.86 (Official Build) (64-bit), Safari 12.0.3

onfire4g05:
This seems to have fixed #567 for me.

This keeps the static resources needed by express outside of the asar, which have issues after Companion runs for about a week. Before fix I was able to consistently replicate the issue, and after this change cannot.

### DIFF
--- a/lib/server_http.js
+++ b/lib/server_http.js
@@ -32,7 +32,7 @@ express.use(function(req, res, next) {
 	next();
 });
 
-express.use(Express.static(__dirname + '/../public', {
+express.use(Express.static(path.join(__dirname, '/../public'), {
 	dotfiles: 'ignore',
 	etag: true,
 	extensions: [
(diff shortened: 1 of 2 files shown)


josephdadams:
This bug plagues me weekly so when I am back in the office I can test it. If it’s not merged by then :)

Issue #1197, opened 2020-08-16
punctuation characters are not properly aligned on the button


istnv:
**Describe the bug**
Certain punctuation characters are not properly aligned on the button.

**To Reproduce**
1. Enter this text on a button: `"Test" `
2. Button actually displays `„Test„`

**Expected behavior**
 Expect button to display `"Test"`

**Desktop (please complete the following information):**
N/A

**Additional context**

PR #1206, opened 2020-08-24
Fix quote glyph alignment

istnv:
Adjust character position for single (') and double (") quotes for better alignment with text.


istnv:
Fixes #1197

PR #1234, opened 2020-09-24
Minor: add parseVariables in instance_skel

krocheck:
@Julusian do you want to update the TS definitions for this PR?


Julusian:
@krocheck done :)

PR #1256, opened 2020-10-20
Module: Upgraded h2r-graphics - v1.1.0 

iamjohnbarker:
Adding all the currently missing commands in the h2r-graphics modules for Companion.

This is my first whack at a companion module update so here's hoping I didn't mess things up too much!


josephdadams:
@iamjohnbarker no PR against the core repo needed. Just add your commits to the hr2-graphics module repository directly (since you are a maintainer) and then ask for it to be updated in the core over in the slack group. One of the devs will do that. We use submodule links against repositories rather than PRs into the core repo.

josephdadams:
Also convert your spaces to tabs :)

iamjohnbarker:
Good to know @josephdadams - I was sure I read otherwise on the Wiki.

Tabs coming up!

PR #1277, opened 2020-11-10
Patch: Scheduler updates

onfire4g05:
- Fixes bug with variables
- Added babel support so public files can be compiled on build to work with old browsers


millsemillse:
fantastic, great tool! 
one point: <> in the condition does not work with variable values providing additional chars , so -30 dBFS will not work with < or >. Maybe a sorting out by the user in the frontend ( a field with a given pattern abcVALUExyz) could solve this issue.

krocheck:
FYI, the full commit stack is coming through because I did a squash and merge originally.  The same should happen to merge this in if it look OK.

onfire4g05:
Hah! I was wondering why the full stack showed but the changes tab showed correctly. I had to merge this way, since we're currently using the scheduler, so basically I had to fork the dev branch and merge those changes back in so I can still use the master branch. 😬 There will probably be other changes like this in the future; hope that's ok.

As for the issue @millsemillse brought up, if anyone has feedback on the best way for that... I'm thinking the best thing is just doing a `parseFloat()` on the variable if using `<` or `>`.

package.json
```diff
@@ -81,15 +81,16 @@
 			"!font/*",
 			"!tools/*",
 			"!*.md",
-			"!node_modules/sharp/vendor/lib"
+			"!node_modules/sharp/vendor/lib",
+			"!public/js/es6"
```
Julusian: I think this should be `public/js/dist`

package.json
```diff
@@ -81,15 +81,16 @@
...
 		]
 	},
 	"scripts": {
 		"prod": "./tools/build_writefile.sh && ./bitfocus-skeleton/start.sh",
-		"dev": "./tools/build_writefile.sh && DEVELOPER=1 ./bitfocus-skeleton/start.sh",
+		"dev": "rm -rf ./public/js/dist ; ./tools/build_writefile.sh && DEVELOPER=1 ./bitfocus-skeleton/start.sh",
 		"update": "./tools/update.sh",
 		"pack": "electron-builder --dir",
-		"dist": "rm -rf electron-output && electron-builder",
+		"dist": "./node_modules/.bin/babel ./public/js/es6 --out-dir ./public/js/dist ; rm -rf electron-output && electron-builder",
```
Julusian: I think this should this be using `public/js/babel` not `public/js/es6`?

public/js/babel/schedule.js
```diff
@@ -0,0 +1,541 @@
+class schedule_frontend {
```
Julusian: The old public/schedule.js should be deleted as this replaces it

Julusian (changes requested):
It looks like a few paths have gotten messed up.

I am wondering if using babel like this is the way to go or if we should use webpack instead which will itself use babel, but will output us a single combined js file and can do hot-reloading.
@haakonnessjoen what do you think?

public/js/babel/schedule.js
onfire4g05: Weird, I renamed the file in commit 460bb72, but it came back in 8b05c5d merge. Not sure how or why that happened.

willosof:
Where are we on this? :)

onfire4g05:
Should all be resolved.

PR #1290, opened 2020-11-28
Refactor: ES6 first pass

krocheck:
@Julusian and @haakonnessjoen this is general refactoring of `lib` to ES6 classes.  It does not include `lib/tcp`, `lib/telnet`, `lib/udp`, and `lib/instance_skel` since those are module touchpoints and I didn't want to break anything in that space out of the gate.  As I noted to @haakonnessjoen, the purpose of this pass was to implement ES6 classes only, so in that respect its a partial refactoring.  Much more work to replace `var` with `const` and `let` declarations and other structural changes will need to be built off this.

The classes are known to compile and load.  They've been 'high-level' tested, meaning various functions have been tested in each module but a full top-to-bottom test strategy was not completed.

One known bug I haven't been able to track down is that the USB device handlers do not destroy on a USB disconnect and if the same device it reconnected it will double (or triple, etc. based on disconnect/reconnects) execute key presses.

I would hold on a merge into `develop` until after v2.1.1 is released and the standing Node/electron updates can be merged to `master`.  That is why I have this in draft.


willosof:
ahrg, I managed to press "ready for review" by a mistake. How do I undo that? ;)

willosof:
And what should be done here? Any reasons to not pull this into master? :) We should ofcourse refactor everything, but this is a good start?

Julusian:
I have no objection to this being merged, and if we are to do so then we should do it soon before this becomes stale or full of merge conflicts.
It looks like the 2.x codebase has a good deal of life left in it before a 3.0 prototype is ready to consider using, so getting it into a better shape sounds like a good idea and will make it easier to review what wants porting and what wants rewriting.

krocheck:
Conflicts to date are handled

krocheck:
@haakonnessjoen any thoughts here?

Issue #1274, opened 2020-11-05
Rescan USB - Keyboard Shortcut?


DanDiamond:
It would be very helpful to have a keyboard shortcut to have Companion rescan USB so I don't have to reopen the GUI everytime I switch locations.

josephdadams:
I'm not sure a global hotkey will be possible across all platforms to perform a rescan, or if this would really be desired by everyone. Can you elaborate more on what it looks like when you switch locations? Are you physically disconnecting your deck via USB and then reconnecting it?

What if a button was added to the system tray popup to rescan?

DanDiamond:
I'm changing locations and networks. When I come back to the studio I often need to rescan to connect. A system tray popup button would be great.

Thanks,

Dan

*( http://www.dandiamondmd.com )
*

*EMAIL:* dan@dandiamondmd.com ( dan@dandiamondmd.com )

*WEBSITE:* www.dandiamondmd.com ( https://www.dandiamondmd.com/ )

*PODCAST:* www.trenchtested.com ( http://www.trenchtested.com )

*PHONE:* 1–360-620-5000

( https://player.vimeo.com/video/279736043 )

People perform best when they believe they have the power to make a difference and a conviction to put other people first.

~ Dan Diamond MD

Dalesjo:
Would it not make more sense to make Companion do this automatically. Either rescan after "system(power): Resuming" or try to rescan if it detects that a usb device has disappeared? I have the same issue when starting up my computer (Windows 10) from hibernation. Pressing "Rescan USB" detects the streamdeck again.

Looking into the logs, Companion detect both returning from hibernation and that the streamdeck disconnected

```
22. 15:01:32 device(AL37G1A03428<): Clear deck failed: Error: Cannot write to HID device
22. 15:01:32 device(AL37G1A03428<): Elgato Streamdeck device disconnected
22. 17:08:39 system(power): Resuming
```

After pressing "Rescan USB"
```
22. 15:08:58 device(AL37G1A03428): Elgato Streamdeck detected
```

DanDiamond:
Samuel, I agree. Automating it is a great idea!

*( http://www.dandiamondmd.com )
*

*EMAIL:* dan@dandiamondmd.com ( dan@dandiamondmd.com )

*WEBSITE:* www.dandiamondmd.com ( https://www.dandiamondmd.com/ )

*PODCAST:* www.trenchtested.com ( http://www.trenchtested.com )

*PHONE:* 1–360-620-5000

[...]

PR #1303, opened 2020-12-07
Minor: Add context-menu to tray icon

Julusian:
Adds a small context-menu to the tray-icon, which is shown on right-click (existing behaviour on other clicks)

tempsnip
Since the screenshot, I also added a 'Launch GUI' option, which behaves the same as shown in the skeleton window

This is the easy part of #1274


yurividal:
@Julusian can i suggest you add a "Exit" or "Quit" button on the context menu too?

Issue #1319, opened 2020-12-15
Hide List of Dynamic Variables when they're all programmatically removed


estilles:
**Describe the feature**
Automatically hide the List of Dynamic Variables if/when they're all programmatically removed

**Is this platform dependent (windows, mac, ..)?**
No

**If documentation is required to implement, do you know where to find it?**
No documentation required. I'll be submitting a pull request for this issue.

**Usecases**
I'm adding a user configurable polling feature to the elgato-keylight module (see example below). There will be a checkbox that will allow the user to turn polling on or off. When polling is on several dynamic variables will be populated. If the user turns polling off, the dynamic variables will no longer be available and will programmatically removed.

I would like to hide the List of Dynamic Variables once they no longer exist.

**EXAMPLE**
> When polling is enabled and variables exist
![image](https://user-images.githubusercontent.com/5514878/102217560-b9447080-3f17-11eb-8549-b5150639c47a.png)

>When polling is disabled, variables are programmatically removed and they automatically be hidden
![image](https://user-images.githubusercontent.com/5514878/102217814-16d8bd00-3f18-11eb-9d65-8e4756e5f76d.png)

krocheck:
I don't know.  I'd prefer the table be emptied rather then hidden.

PR #1320, opened 2020-12-15
Patch: hide the dynamic variables section on empty table

estilles:
Adds #1319 
When `instance_variables[]` are programmatically removed the `#instanceConfigVariables` element should be hidden.

### DIFF
--- a/public/instance.js
+++ b/public/instance.js
@@ -491,7 +491,8 @@ $(function() {
 				var variable = instance_variables[current_instance][i];
 				$icvl.append('<tr data-id="' + current_instance + ':' + variable.name + '"><td>$(' + current_instance + ':' + variable.name + ')</td><td>' + variable.label + '</td><td>' + instance_variabledata[current_instance + ':' + variable.name] + '</td></tr>');
 			}
-
+		} else {
+			$icv.hide();
 		}
 	}
 


krocheck:
I don't know.  I'd prefer the table be emptied rather than hidden.

estilles:
> I don't know. I'd prefer the table be emptied rather than hidden.
Fair enough. I'll update my PR. Thanks for the comments!

estilles:
@krocheck ... I just remembered why I chose to hide instead to empty to table. It was because When I emptied the table, the title and table heading were still visible.

Would you be open with doing both hiding and emptying the table?

krocheck:
The title and headers showing is correct.  For modules that don't expose any variables it keeps people asking where are the variables because they can see the empty table and [hopefully] intuitively understand that module doesn't offer any.

estilles:
> The title and headers showing is correct. For modules that don't expose any variables it keeps people from asking where are the variables because they can see the empty table and [hopefully] intuitively understand that module doesn't offer any.

Understood. However, currently modules that don't have variables typically don't show the title and table headings already, so I was just trying to keep it consistent.

`$icv.show();` doesn't execute when `instance_variables[current_instance]` is undefined or has length 0.

krocheck:
Oh yeah, how about that!  I wonder if that changed at some point and I wasn't aware.  Well then yeah, this is fine.

estilles:
> Oh yeah, how about that! I wonder if that changed at some point and I wasn't aware. Well then yeah, this is fine.

Awesome! I'll update the PR. Thanks again!

estilles:
> Oh yeah, how about that! I wonder if that changed at some point and I wasn't aware. Well then yeah, this is fine.

Hold on. I'm confused. Do you mean you're okay with the PR as is, or should I update the PR to both hide and empty the table?

krocheck:
It's good

PR #1328, opened 2020-12-16
Merge pull request #1 from bitfocus/master

rodrigocaravaca:
updating my fork


Julusian:
It looks like you are trying to do something wrong.
In this PR you are trying to merge nothing into companion, which is pointless

PR #1333, opened 2020-12-17
Change wording in window to 'Exit' and 'Minimize'

### LINKED ISSUE #1331: [BUG] [MINOR] "Close" should be renamed to "QUIT" or "EXIT"
The "close" action button on the electron window can be a little bit confusing. I suggest "exit" or "quit", as it is more usually adopted when a button terminates the program.

Maybe, the "Hide" could also be renamed to "Minimize" or "Minimize to Tray".

NoahCallaway:
Fixes #1331 

Change 'Hide' and 'Close' to 'Minimize'

Screenshot 2020-12-17 at 18 38 31

### DIFF
--- a/window.html
+++ b/window.html
@@ -61,14 +61,14 @@
 				border: 1px solid white;
 			}
 
-			#hide {
+			#minimize {
 				background-color: rgba(255,255,255,0.1);
 				color: white;
 				margin-right: 5px;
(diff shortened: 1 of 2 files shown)


NoahCallaway:
Ahh! Been fixed just as I was doing this!

PR #1336, opened 2020-12-20
Minor: Change module load to just-in-time

krocheck:
This would change execution from loading the main module scripts at launch to a just-in-time load during a validation check at instance add/activation.  With over 200 modules now it doesn't seem necessary to load all module code into RAM at launch.  Though there aren't major benefits either, so the usefulness of a change like this needs review.

Pros:
- ~5MB less RAM use
- Shorter launch speeds.

The RAM drop was from 381MB to 376MB on Win, so a 1% drop or effectively nothing.  The shorter launch speeds were perceivable on Windows, especially since the skeleton will normally hang for a brief period on launch.  This hang goes away completely.

Cons:
- A module add/activate may fail where-as an error related to the module may have shown in the logs at launch instead
- Such an error won't crash the program but will show a module error in the log at that time

This would be required long-term anyway if a shift to running instances in child processes is pursued.

This should be a `Squash and Merge`

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -31,6 +31,7 @@ function instance(system) {
 	self.modules_manufacturer = {};
 	self.modules_category = {};
 	self.modules_name = {};
+	self.modules_main = {};
 	self.package_info = {};
 	self.status = {};
 
@@ -161,7 +162,7 @@ function instance(system) {
 
 				var packagefile = fs.readFileSync(path +'/'+ folder + '/package.json');
 				var moduleconfig = JSON.parse(packagefile);
-				var mod = require(path + '/' + folder + '/' + moduleconfig.main);
+				var main = path + '/' + folder + '/' + moduleconfig.main;
 				if (fs.existsSync(path +'/'+ folder + '/HELP.md')) {
 					moduleconfig.help = true;
 				} else {
@@ -169,7 +170,7 @@ function instance(system) {
 				}
 
 				self.store.module.push(moduleconfig);
-				self.modules[folder] = mod;
+				self.modules_main[folder] = main;
 				self.package_info[folder] = moduleconfig;
 
 				// Generate keywords list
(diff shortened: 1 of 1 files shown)


Julusian (approved):
Yeah I like the idea here. I have noticed the startup times, especially when developing a module.

In the `checkModuleLoaded()` function, I think the `folder` parameter should be called `instance_type`, as that better matches the usages.

Also, it would be nice (but not essential) to have an environment variable that forces all the modules to be loaded at startup. The reason being that while fiddling with electron/dependency versions, sometimes the modules will throw an error which could easily slip undetected with this change.

krocheck:
I'll made the change to the parameter name @Julusian.

I can add this back to the initial loop:

krocheck:
I'll made the change to the parameter name @Julusian.

I can add this back to the initial loop, which effectively restores what was there:

```javascript
if (process.env.DEVELOPER !== undefined) {
	var mod = require(path + '/' + folder + '/' + moduleconfig.main);
	self.modules[folder] = mod;
}
```

Do you want this or is there a way to address a flag, such as: `yarn dev --load-all-modules`

Julusian:
@krocheck I think it should be separate to the DEVELOPER variable, as the lazy loading will be a nice speed boost when developing a module.

We could do it via a flag, but thatll require a library to help parse those, and that may take some fiddling to work both headless and in electron. I would be happy with it checking a variable called `DISABLED_MODULE_LAZY_LOAD` or similar.

krocheck:
Is `DISABLE_` the correct terminology?  What about `LOAD_ALL_MODULES`?  Rather then enabling or disabling a behavior, you're telling the program exactly what to do.

Julusian:
I was thinking that lazy loading modules is a new feature so having an option to disable that new feature would make sense.
But `LOAD_ALL_MODULES` is fine with me. I dont really mind very much what it is called

PR #1357, opened 2021-01-02
Minor: 'abort_one' action

istnv:
Core/internal listener for the `abort_one` action.
My pretty-formatter also stripped some trailing whitespace :)

### DIFF
--- a/lib/action.js
+++ b/lib/action.js
@@ -247,7 +247,7 @@ function action(system) {
 	});
 
 	// If a user wants to abort a single button actions
-	self.system.on('action_abort_bank', function(page, bank) {
+	self.system.on('action_abort_bank', function(page, bank, unlatch) {
 
 		var bid = page + '_' + bank;
 		var cleared = 0;
@@ -264,6 +264,11 @@ function action(system) {
 				cleared += 1;
 			}
 		})
+
+		// if requested, reset and skip up-actions
+		if (unlatch) {
+			self.system.emit('graphics_indicate_push', page, bank, false);
+		}
 
 		if (cleared > 0) {
 			self.system.emit('graphics_bank_invalidate', page, bank);
--- a/lib/action.js
+++ b/lib/action.js
@@ -247,25 +247,25 @@ function action(system) {
 	});
 
 	// If a user wants to abort a single button actions
-	self.system.on('action_abort_one', function(bank) {
+	self.system.on('action_abort_bank', function(page, bank) {
 
-		var bid = bank[0] + '_' + bank[1];
-		var c = 0;
+		var bid = page + '_' + bank;
+		var cleared = 0;
 
 		self.actions_running.delete(bid);
 
 		self.timers_running.forEach((timerId, timer) => {
-			if (timerId === bid) { 
-				if (c == 0) {
-					debug("Aborting button ",bank[0],",",bank[1]);
+			if (timerId === bid) {
+				if (cleared == 0) {
+					debug("Aborting button ",page,",",bank");
 				}
 				clearTimeout(timer);
-				c += 1;
+				cleared += 1;
 			}
 		})
 
-		if (c > 0) {
-			self.system.emit('graphics_bank_invalidate', bank[0], bank[1]);
+		if (cleared > 0) {
+			self.system.emit('graphics_bank_invalidate', page, bank);
 		}
 	});
 
--- a/lib/action.js
+++ b/lib/action.js
@@ -245,6 +245,28 @@ function action(system) {
 
 	});
 
+	// If a user wants to abort a single button actions
+	self.system.on('action_abort_one', function(bank) {
+
+		var bid = bank[0] + '_' + bank[1];
+		var b;
+		var c = 0;
+
+		while ((b = self.actions_running.indexOf(bid)) >= 0) {
+			if (c == 0) {
+				debug("Aborting button ",bank[0],",",bank[1]);
+			}
+			self.actions_running.splice(b,1);
+			clearTimeout(self.timers_running[b]);
+			self.timers_running.splice(b,1);
+			c += 1;
+		}
+
+		if (c > 0) {
+			self.system.emit('graphics_bank_invalidate', bank[0], bank[1]);
+		}
+	});
+
 	// skipNext needed for 'bank_pressed' callback
 	var skipNext = {};
 	var pageStyles = ['pageup','pagenum','pagedown'];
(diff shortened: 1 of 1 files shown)


lib/action.js
```diff
@@ -245,6 +245,28 @@ function action(system) {
 
 	});
 
+	// If a user wants to abort a single button actions
+	self.system.on('action_abort_one', function(bank) {
```
Julusian: instead of having to use bank[0] and bank[1] it would be nicer to have two parameters, page and bank.

I would prefer a name that refers to 'bank' rather than 'one' (eg `action_delayed_abort_bank`), but that will require a change in the module using this too

lib/action.js
```diff
@@ -245,6 +245,28 @@ function action(system) {
...
+
+		var bid = bank[0] + '_' + bank[1];
+		var b;
+		var c = 0;
+
+		while ((b = self.actions_running.indexOf(bid)) >= 0) {
+			if (c == 0) {
+				debug("Aborting button ",bank[0],",",bank[1]);
+			}
+			self.actions_running.splice(b,1);
```
Julusian: I think this is making a bad assumption about the relationship between `self.actions_running` and `self.timers_running`.

For each bank press with delayed actions, one entry will be added to actions_running on https://github.com/bitfocus/companion/blob/master/lib/action.js#L369, and multiple entries into timers_running https://github.com/bitfocus/companion/blob/master/lib/action.js#L409. The exact number depends on how many delayed actions there are.

For this to work, either the two arrays should be combined (perhaps using a Map of arrays?) or timers_running should contain pairs of values. The timer and the bank id.

lib/action.js
```diff
@@ -245,6 +245,28 @@ function action(system) {
 
 	});
 
+	// If a user wants to abort a single button actions
+	self.system.on('action_abort_one', function(bank) {
+
+		var bid = bank[0] + '_' + bank[1];
+		var b;
+		var c = 0;
```
Julusian: It would be nice to have a more descriptive name for `c` (eg `changed`)

Julusian (changes requested):
This needs some larger changes to how delayed actions are stored to work properly.

lib/action.js
krocheck: This name and parameter is defined in the bitfocus-companion module already, though that could be changed to suit.  It appears the ["Abort actions on button"](https://github.com/bitfocus/companion-module-bitfocus-companion/blob/7a84df70064e8a3907ecbff5ea66359885d4a58b/index.js#L717) action in internal has existed for a while without code to execute having been implemented.  Though I agree that `(page, bank)` as the parameters follows conventions elsewhere in the software.

istnv: The action 'request' in the internal module came about while adding an option for 'this button' on button modifiers. The actual execution was in this module. At the moment it is a do-nothing action since the listener is in this patch.

lib/action.js
istnv: With a single delayed action, the actions<->timers arrays do map out. It works great to create a single-button press-repeat, cancel on release, without aborting everything else.
While working on a demonstration of how this works with multiple actions... As @Julusian pointed out it does not map properly with more than one delayed action. I was thinking about adding a bank/page id as part of the timers_running object to locate the timers involved.

Julusian: @istnv yeah I am currently thinking about a Map with keys being the timer, and the value being the bankid. You can then iterate over the map contents to find the items.
Im going to have a play with that quickly, and also refactor actions_running a bit to be more durable/simpler

lib/action.js
istnv: I'll change that to `cleared`

lib/action.js
istnv: The only module using this is the `bitfocus-companion` internal module.

lib/action.js
Julusian: @istnv I have pushed a change to https://github.com/bitfocus/companion/tree/feat/abort_one that switches out these structures and some of the flow here to resolve my concerns.
Unless @krocheck has any concerns about my use of Map and Set instead of normal objects and arrays, I suggest using that as a fix for this comment thread

istnv: @Julusian So I should close this PR and use your update as a base for the other changes suggested.

Julusian: You dont need to close the pr, you can pull the changes into your branch. 
`git fetch --all` then `git merge origin/feat/abort_one` (might not be called origin, depends on how your git remotes are setup) then continue making changes. there might be a conflict to resolve, but to resolve that you just need to make and commit the appropriate changes.

istnv: I'm working against a fork, how do I pull a branch without rebuilding my repo?

Julusian: good point. make that
* `git remote add upstream https://github.com/bitfocus/companion.git` 
* `git fetch --all`
* `git merge upstream/feat/abort_one`

istnv: Still learning git. Added a remote :)

krocheck: @Julusian if I knew what Map and Set were I'm sure I'd take issue for some reason, but since I don't I trust your judgement :)  You know I still suck at JS in certain areas.

Julusian: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

The same thing could be done with plain objects, but I generally prefer using these when possible. So its more of a style choice than anything

istnv:
One more question about this feature. The abort procedure for a toggle button that is 'down'... Should it be un-latched and the up-actions aborted/skipped as well? I think abort-toggle is currently in a feature request and this is probably the place to do it. (It is 10 lines of code)

lib/action.js
```diff
@@ -222,29 +222,53 @@ function action(system) {
...
 
+	// If a user wants to abort a single button actions
+	self.system.on('action_abort_bank', function(page, bank) {
+
+		var bid = page + '_' + bank;
+		var cleared = 0;
+
+		self.actions_running.delete(bid);
+
+		self.timers_running.forEach((timerId, timer) => {
```
Julusian, line 257: oops, I forgot to delete the timers from self.timers_running. This block of code is only calling clearTimeout on them, so will be leaking the refernces...

istnv: Fixed.

Julusian:
@istnv I think that should be a separate action. To me that would be a weird side effect and undesired at times

lib/action.js
Julusian: Perfect :)

Julusian (approved):

istnv:
@Julusian That would depend on how 'abort' is defined. Since it is a single button action, should it only apply to things in progress or does it do a full button reset? One could also assume that it would abort any pending 'up' actions.

istnv:
@Julusian Maybe add an 'Unlatch' option in the module action that initiates the abort? Then it would be users choice.

PR #1364, opened 2021-01-08
Patch: fixed reference to button config during page/bank config init

estilles:
I was in the middle of doing some research and I encountered this error. I'm not certain if it's causing any issues, but I figured we should fix it anyway.


Julusian:
Doesn't look like it will cause any issues, but it is good to fix it anyway :)

PR #1401, opened 2021-01-28
fix: cross-platform building

Julusian:
Doing a `yarn windist` on mac or `yarn macdist` on windows fails when trying to reinstall the app dependencies.

This gets us closer to restoring that working. The identified issues:
* [x] electron-builder tells prebuild-install to builds for the wrong runtime https://github.com/develar/app-builder/pull/47
* [x] @serialport/bindings@8.0.8 not available for electron 9+. The version is specified by osc, and trying 9.0.x caused crashes on windows. The crashing is resolved with electron 10.
* [x] sharp does some extra magic to install dlls and things as well as prebuild-install. This needs to be triggered to get the full sharp dependencies


MeestorX:
`rpidist` is also broken (cannot be built on Win10 or MacOS 10.15+), which makes me thing that building a linux version might also be broken. All platforms could be built on MacOS 10.14.

Julusian:
@MeestorX yes, this will be the same for all cross-platform builds not just the ones I mentioned

Julusian:
Getting closer. from linux I can now do a `yarn windist` and get a working build.
`yarn rpidist` is still broken, I need to get a pi running a non arm64 os to figure out what is going on

Julusian:
Its not a perfect test as it is armv6l not armv7l, but based on the pi-zero I have easily accessible, it had to compile sharp from scratch itself. I see no reason for armv7l to be different, based on the prebuilds available on github

I dont know how to progress from here, so Im going to leave rpidist broken for now and revisit it later

PR #1412, opened 2021-01-30
Added styling over TCP, updated settings tab info

Haavard15:
Amateur programmer and first actual pull request.

I needed to be able to style buttons over TCP so I added changing text and color/bgcolor.

I also noticed that the HTTP API had some un-documented features so I added those to the settings tab along the new things.

Let me know if there is anything I should change.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -560,9 +560,9 @@ href="https://join.slack.com/t/bitfocusio/shared_invite/enQtODk4NTYzNTkzMjU1LTMz
 											<ul>
 												<li><code>/press/bank/</code> &lt;page&gt;<code>/</code>&lt;bank&gt;<br>
 												<i>Press and release a button (run both down and up actions)</i></li>
-												<li><code>/style/bgcolor/</code>&lt;page&gt;<code>/</code>&lt;bank&gt;<code>?text=</code>&lt;bgcolor HEX&gt;<br>
+												<li><code>/style/bank/</code>&lt;page&gt;<code>/</code>&lt;bank&gt;<code>?bgcolor=</code>&lt;bgcolor HEX&gt;<br>
 												<i>Change background color of button</i></li>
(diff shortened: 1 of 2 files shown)


public/index.html
```diff
@@ -548,6 +554,37 @@ <h4>TCP/UDP Remote control</h4>
 										<p>Press page 1 bank 2<br>
 											<code>BANK-PRESS 1 2</code></p>
 
+										<h4>HTTP Remote control</h4>
+										<p>Remote triggering can be done by sending HTTP Requests to the same IP and port Companion is running on.</p>
+										<p><strong>Commands:</strong>
+											<ul>
+												<li><code>/press/bank/</code> &lt;page&gt;<code>/</code>&lt;bank&gt;<br>
+												<i>Press and release a button (run both down and up actions)</i></li>
+												<li><code>/style/bgcolor/</code>&lt;page&gt;<code>/</code>&lt;bank&gt;<code>?text=</code>&lt;bgcolor HEX&gt;<br>
```
Julusian: This does not look correct to me.
from a skim of the code it looks like it should be `/style/bank/<page>/<bank>?bgcolor=<hex>` 
etc

Haavard15: Good catch! I committed to my own repo, does it look correct now or do I have to do anything more to make it appear here?

public/index.html
```diff
@@ -548,6 +554,37 @@ <h4>TCP/UDP Remote control</h4>
...
+												<i>Change text size on a button (between the predefined values)</i></li>
+											</ul>
+										</p>
+
+										<p>
+											<strong>Examples</strong>
+										</p>
+
+										<p>Press page 1 bank 2<br>
+											<code>/style/bank/1/2</code></p>
```
Julusian: should be `/press/bank/1/2`

public/index.html
Julusian: It shows up here too :)

public/index.html
Haavard15: Thanks :-)

PR #1415, opened 2021-01-31
Changed buttons tab to be selected by default

murphytodd4:
Title is pretty self explanatory. This is in reference to a thread on Slack (https://bitfocusio.slack.com/archives/CFG7HAN5N/p1611578947020100). I think I got everything changed that is needed. The web interface appears to be loading correctly with the buttons tab selected by default.


bryce-seifert:
It looks like this got reverted to Instances being selected first with https://github.com/bitfocus/companion/commit/9b87377d4e941b24d0fcecba6278e929e8afeb3b Was that intentional?

JeffreyDavidsz:
I don't think so. changed it back

PR #1425, opened 2021-02-05
Adding an instance_status variable which gets emitted with instance e…

bkpeterson:
Added a variable to keep track of the status of each instance, which also gets emitted with instance_errors in calculateErrors.  This will be used in the bitfocus-companion module to provide instance-specific status feedback.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "companion",
-	"version": "2.2.0",
+	"version": "2.3.0",
 	"description": "Companion",
 	"main": "electron.js",
 	"build": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "companion",
-	"version": "2.2.0",
```
Julusian: Please don't change the version of core in this pr

Julusian (changes requested):

package.json
bkpeterson: Sorry, reverted.

bkpeterson:
I am waiting on this for an install - any thoughts from the reviewers?  Any further changes or can this be merged?

krocheck:
I have implemented this an alternate way.  There was no need to add `self.instance_status` since `self.status` already exists with the same array.

PR #1451, opened 2021-02-25
Module: Upgraded depili-clock-8001 to 5.1.1

depili:
This fixes a small bug in the module where at least the sync time action was failing.


depili:
Why was this rejected?

JeffreyDavidsz:
Updated via module instead of this pull. See slack

PR #1468, opened 2021-03-03
chore: apply prettier

krocheck:
Somebody else can hit the go button on this one.  I have a plan so it won't be merge mess with my other stuff.  But let's be quick about approving so that doesn't fall apart.

The code compiled and ran fine.  I did have to clean some things up, as I did scrub every file top to bottom in a split view.


krocheck:
Also ... squash and merge this for simplicity sake

Julusian:
Ill do so now.
Im going to move the prettierrc to the root of the repo so that we cover the files in other folders too

PR #1476, opened 2021-03-04
feat: Implement 'remove top bar' setting in React UI

NoahCallaway:
Implements the remove top bar feature in the new react UI.
There is also a change to update surfaces after the change has taken place. This prevents users needing to change page in order to see the updated look and eliminates the need to refresh web GUI.

### DIFF
--- a/lib/graphics.js
+++ b/lib/graphics.js
@@ -166,13 +166,7 @@ graphics.prototype.generate = function (invalidate) {
 	var self = this
 
 	for (var p = 1; p <= 99; p++) {
-			self.drawPage(p)
-			//Invalidate all on page if required
-			if (invalidate) {
-				for (var b = 1; b <= 32; b++) {
-					system.emit('graphics_bank_invalidated', p, b)
-				}
-			}
+		self.drawPage(p, invalidate)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


Julusian (changes requested):
I think it would be better to make generate do the emit instead.
By adding a parameter to make `generate(invalidate)`, this can be kept optional. This will avoid extra amounts of looping.

Other than that, looks good

lib/graphics.js
```diff
@@ -162,11 +162,17 @@ graphics.prototype.isPushed = function (page, bank, cb) {
...
+graphics.prototype.generate = function (invalidate) {
 	var self = this
 
 	for (var p = 1; p <= 99; p++) {
-		self.drawPage(p)
+			self.drawPage(p)
+			//Invalidate all on page if required
+			if (invalidate) {
+				for (var b = 1; b <= 32; b++) {
+					system.emit('graphics_bank_invalidated', p, b)
```
Julusian: Sorry, I meant to say to pass the invalidate parameter onto drawPage, and do the check below the `self.drawBank(page, bank)` line

NoahCallaway: Ahh I see.
Will do

PR #1501, opened 2021-03-14
Minor: Tablet/ Emulator CSS tweaks

bryce-seifert:
–Center emulator on page
–Align page name and buttons to edge of buttons in tablet view

### DIFF
--- a/webui/src/App.scss
+++ b/webui/src/App.scss
@@ -67,9 +67,10 @@
 
 	> div {
 		margin: 1em auto;
-		max-width: 860px;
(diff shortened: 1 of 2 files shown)


Julusian:
Thanks. I did a few more tweaks on this and included it as https://github.com/bitfocus/companion/commit/b9002c12348cc9b94178c71daa29f5de3e71d59f.

`.page-white` was also being used for the getting-started page, and this change broke that a bit. I split the class usages up instead for simplicity, which made this easier

PR #1509, opened 2021-03-20
Improved Windows build compatibility

astellingwerf:
When I ran the build on Windows 10 in the Git Bash console, at some point, it attempted to start a cmd prompt, where it didn't recognize the syntax of setting the variable at the start of the line.


Julusian (changes requested):
Unfortunately, this does not work. If I try running `yarn dist:prepare:win` on a non-windows machine, then I end up with non-windows sharp binaries.

I have pushed a change to master that I think will work for you, please test and let me know https://github.com/bitfocus/companion/commit/85db6be5666c01ae76a3cd8d59c31088fd7b9283

krocheck:
For the same reason I typically modify the same script locally to be `set npm_config_platform=win32 && set npm_config_arch=x64 && yarn dist:prepare:sharp`.  The alternate change using `cross-env` worked for me.

astellingwerf:
`cross-env` seems to work for me. I guess we can close this PR with that.

Issue #1496, opened 2021-03-12
disable key-up action


fdebeer:
make an internal action to disable key-up action

**Usecases**
i have a button with a delayed action on key-down. when i release the button before the delay. key-up gives a aboard all delayed action and preforms another action on key-up.  i would like the button not to do this when the action on the key-down is preformed. so when i release the button no action is taken for key-up. 
this 'internal:disable key-up' has to be put after the delay and the action in key-down
it makes it possible to make 2 commands on 1 button. 
shortpress command 1, longpress command 2

i use this now for ptz-cam. 
key-down 1000ms delay and save preset
key-up abort delayed action and recall preset
but no need for recall preset when preset is just saved. saves a bit of network action

Julusian:
I would rather do #925 instead of this. While this will be a lot simpler to implement, it will be something that will need to be maintained going forward even once there is a better solution

istnv:
I think the `unlatch` option for the `Abort specific button` action cancels the UP actions, too.

fdebeer:
> I think the `unlatch` option for the `Abort specific button` action cancels the UP actions, too.

no it does not what i mean with cancel key-up

@Julusian; I would rather do #925 instead of this
this is a discussion about what i already do with save/recall preset PTZ, and it works well.
key-up recalls preset which is no problem after save preset
What i like to have is a cancel key-up so you can put 2 commands on 1 button
like key-up starts recording, key-down with 2000ms delay stops recording. in this setup you dont want key-up to go after longpress keydown

istnv:
@fdebeer That is what `unlatch` does. It aborts the `key-up` commands.

istnv:
@fdebeer  Try this on a button:

key-down:
1. 2000ms delay, Stop recording
2. 2010ms delay, `Abort actions on one button`:  Select `This button` and `Unlatch?`

key-up:
1. `Abort actions on one button`:  Select `This button`
2. start recording

fdebeer:
> @fdebeer Try this on a button:
> 
> key-down:
> 
> 1. 2000ms delay, Stop recording
> 2. 2010ms delay, `Abort actions on one button`:  Select `This button` and `Unlatch?`
> 
> key-up:
> 
> 1. `Abort actions on one button`:  Select `This button`
> 2. start recording

tried it. doesn't work. still performs key-up action
did it with color background change:
key down: 2000 delay; bgcolor change red; 500 ms bgcolor black; abort action on button with unlatch
[...]

PR #1516, opened 2021-03-23
Add ability for non-latch button to abort up acts

istnv:
Allows `Unlatch?` option for `Abort actions on button` to also skip up-actions on non-latch buttons.
Provides feature requested in #1496

### DIFF
--- a/lib/action.js
+++ b/lib/action.js
@@ -426,22 +426,22 @@ function action(system) {
 					// is this a timedelayed action?
 					if (delay > 0) {
 						has_delayed = true
-						;(function (action, delay_time) {
-							var timer = setTimeout(function () {
-								self.system.emit('action_run', action, { deviceid: deviceid, page: page, bank: bank })
+							(function (action, delay_time) {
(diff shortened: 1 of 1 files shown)


istnv:
Interesting. Prettier is failing here:
Line 417:` 				;(function (action, delay_time) {`

Julusian:
@istnv that semi colon is weird but essential. otherwise this happens https://github.com/bitfocus/companion/commit/8cafcb7934a94042c4cc838a3512115b5bfbf618#diff-28a32a1c22ae1422b41c1b8707c7a80f792fedee48368e16136590238791428aR427

Because js doesnt know that the bracket starting on the next line is a new statement, and not a continuation of the line before.
As prettier has been told to not have semi colon line endings, it instead puts them where necessary at the start of lines to ensure the meaning is correct

istnv:
@Julusian Now I need to unlearn months of semi-colons :)  Most of my programming has not needed them so js has been a small struggle.

Issue #1545, opened 2021-04-05
[FEATURE] Automatically resize button images that exceed maximum dimensions


estilles:
**Describe the feature**
At the moment, when uploading a button image that is larger than 72x72 we get an error message indicating;
`Image must have the following dimensions: 72x72`.

I propose automatically resizing such images the maximum dimensions if any of the following conditions are met:
- image width is greater than max width, or
- image height is greater than max height

**Usecases**
Not a necessity, just a time saver. I recently acquired a stream deck icon pack, but all the icons are 288x288. 

---

PR to follow, but I do have a question. Should I be submitting the PR against the new React WebUI, the legacy UI or both?

Julusian:
Implemented in #1546

kjetilpp:
I have a similar issue, use icons from materialdesignicons.com that are 48x48, on the Stream deck Elgato software it works fine, but migrating to Companion we have a problem with "Image dimensions must be at least 72x58"
Maybe auto resizing small icons too?

estilles:
Hey @kjetilpp, that's an interesting use case. I never thought of upscaling smaller images/icons. Does the native Stream Deck software display icon as-is at 48x48 (padding the image to fill) or does it upscale it?

kjetilpp:
I think it’s using 48x48 as-is, the icons
in \AppData\Roaming\Elgato\StreamDeck\ProfilesV2\XX.sdProfile\Profiles\XXX.sdProfile\X.X\CustomImages
is still at 48x48, if i change the icon directly in the folder, is changes
on the stream deck.

estilles:
Okay. I'll look into this. This issues was closed months ago. I suggest you create a new ticket with a feature request so we can track the progress.

kjetilpp:
Done #1774

PR #1546, opened 2021-04-05
feat: Resize uploaded button images to max dimensions when too large

estilles:
Implements #1545 

PR includes changes to both legacy and React WebUI

### DIFF
--- a/public/js/preview.js
+++ b/public/js/preview.js
@@ -17,6 +17,34 @@
 		return imageData;
 	}
 
+	function imageResize(img, maxW, maxH, cb) {
+		var canvas = document.createElement('canvas');
+		var width = img.width;
+		var height = img.height;
+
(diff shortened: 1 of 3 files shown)


Julusian (changes requested):
I dont think we should enforce the icons being square, as 72x58 is the correct size for images when the header is shown. Instead we should resize if either the width or height is above their max allowed

eg an image of 100x40 would become 72x29. this could be something designed to be aligned to the bottom of the button with text drawn in the space above

Apart from that, I am happy with the code :)

estilles:
@Julusian ... Understood.

I guess I got confused because I saw a comment indicating a change in default height from 58 to 72 here:
https://github.com/bitfocus/companion/blob/master/lib/bank.js#L66-L77. But what you say makes perfect sense. I'll update the PR accordingly.

PR #1548, opened 2021-04-05
Added Companion Bank Types

thedist:
Added type interfaces for Companion Banks, and updated the `CompanionFeedback` interface as the `callback` function has a 2nd parameter for the feedbacks bank.

### DIFF
--- a/instance_skel_types.d.ts
+++ b/instance_skel_types.d.ts
@@ -5,6 +5,33 @@ export interface CompanionSystem extends EventEmitter {}
 
 export type InputValue = number | string | boolean
 
+export type CompanionBank = CompanionBankPage | CompanionBankPNG | CompanionBankPreset
+
+export interface CompanionBankPage {
+	style: 'pageup' | 'pagedown' | 'pagenum'
+}
+
+export interface CompanionBankPNG {
+	style: 'png'
+	text: string
+	size: 'auto' | '7' | '14' | '18' | '24' | '30' | '44'
+	alignment: 'left:top' | 'center:top' | 'right:top' | 'left:center' | 'center:center' | 'right:center' | 'left:bottom' | 'center:bottom' | 'right:bottom'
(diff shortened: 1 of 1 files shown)


instance_skel_types.d.ts
```diff
@@ -5,6 +5,51 @@ export interface CompanionSystem extends EventEmitter {}
...
+
+export interface CompanionBankPage {
+	style: 'pageup' | 'pagedown' | 'pagenum'
+}
+
+export interface CompanionBankPNG {
+	style: 'png'
+	text: string
+	size: 'auto' | '7' | '14' | '18' | '24' | '30' | '44'
+	alignment:
```
Julusian, line 18: Please make a CompanionBankAlignment type for these to reduce the duplication

instance_skel_types.d.ts
```diff
@@ -5,6 +5,51 @@ export interface CompanionSystem extends EventEmitter {}
...
+		| 'center:bottom'
+		| 'right:bottom'
+	png64?: string
+	color: number
+	bgcolor: number
+	latch: boolean
+	relative_delay: boolean
+}
+
+export interface CompanionBankPreset {
```
Julusian, line 45: I'm not keen on CompanionBankPreset and CompanionBankPNG being separate types, as they are the same thing, but as the preset version wants a lot of the properties to be optional then it becomes a little messy.

I don't expect you to do anything about this right now, I may have a play myself later

Julusian (changes requested):

thedist:
Yeah make any edits you need, my reasoning for this PR is that I'm quite extensively rewriting the vMix module in Typescript, and it was giving warnings because some of my feedback callbacks utilize the bank data which is in the 2nd param, but the interface as it currently stands only defines the 1st param.

Julusian:
I've made some changes and merged this in 529d079c8272b21c09f12ba4bdd9e2cd66da449d.

Thanks for doing the bulk of it :)

PR #1565, opened 2021-04-15
fix: broken links in "Getting Started" guide

McHauge:
Just noticed two broken links that I forgot to update when I moved some stuff around. I have checked all links in the document, they all work from within companion, but will not work on GitHub. do to the way they are defined. not sure why they need 
"-header-"

in front of where it should link, for GitHub use this should not be there and they will work without but not in companion so...


Julusian:
Companion explicitly adds the prefix https://github.com/bitfocus/companion/blob/master/webui/src/GettingStarted.jsx#L41
Maybe this should be removed? I don't know why it is there

McHauge:
Well if it could work on both github and Companion that would be great, all just go though and update all the links, if that becomes possible

McHauge:
@Julusian was this something you would look at or?

McHauge:
Yarn format has been run on this, and it seems to work for me, but @Julusian please take a look at the other file. if the change will break anything else.

McHauge:
Test Comment

PR #1570, opened 2021-04-20
feat: Pass button being updated to feedback call

istnv:
Adds the current button (page, bank) being processed to the module feedback function call.
Allows the module to see which button is being updated.
For @MeestorX

### DIFF
--- a/lib/feedback.js
+++ b/lib/feedback.js
@@ -203,10 +203,10 @@ function feedback(system) {
 										definition.callback !== undefined &&
 										typeof definition.callback == 'function'
 									) {
-										var result = definition.callback(feedback, bank_obj)
+										var result = definition.callback(feedback, bank_obj, { page: page, bank: bank })
 										self.setStyle(page, bank, i, result)
 									} else if (typeof instance.feedback == 'function') {
-										var result = instance.feedback(feedback, bank_obj)
+										var result = instance.feedback(feedback, bank_obj, { page: page, bank: bank })
 										self.setStyle(page, bank, i, result)
 									} else {
 										debug('ERROR: instance ' + instance.label + ' does not have a feedback() function')


MeestorX:
That's some speedy service right there, @istnv ! 
Thank you. 
Q: Is the bank/page is already part of the actions parameters? In the past I had the action.id but not it's page/bank.

istnv:
@MeestorX 
The 'button' is passed as an extra parameter to both action and feedback:
`action(action, button)` and `feedback(feedback, bank, button)`
Use `button.page` and `button.bank`. 
The action call also includes `button.surface` (which surface was pressed)

MeestorX:
Very cool. When was that added, BTW?
I have to see if I can use that to replace this mess:
```
getActionPgBk(id) {
        for (let pg in bank_actions)
            for (let bk in bank_actions[pg])
                for (let act in bank_actions[pg][bk])
                    if (bank_actions[pg][bk][act]['id'] == id) {
                        return {page: pg, bank: bk}
                    }
        for (let pg in bank_release_actions)
            for (let bk in bank_release_actions[pg])
                for (let act in bank_release_actions[pg][bk])
                    if (bank_release_actions[pg][bk][act]['id'] == id) {
                        return {page: pg, bank: bk}
                    }            
    }
```

istnv:
The action call was updated as part of the 'This' button and page feature. Would not work without it :)

MeestorX:
Nice. See, I need to stay on top of these things! :-)

Thank you.

MeestorX:
Ok, weird. I'm not seeing it.
I'm getting these parameters for an action:
```
{
  id: 'Q72eq7Yln',
  label: 'XkS8EoGfN:macroRec',
  instance: 'XkS8EoGfN',
  action: 'Rec',
  options: {},
  delay: 0
}
```
No button parameter...

istnv:
It is an additional parameter, not part of the original `action`:
`instance.prototype.action = function (action, button) {`

MeestorX:
> It is an additional parameter, not part of the original `action`:
> `instance.prototype.action = function (action, button) {`

There you go. Got it. Nice!!

Julusian:
@MeestorX I'm curious why you need to know where the button is located. Can you elaborate?

MeestorX:
Macros, baby! :-)

krocheck:
@Julusian I think the better move here would be for `get_page` to return a `cloneDeep` of the bank configuration and add `page`, `bank`, and `deviceid` to the array.  We can switch `extras` over to this without breaking it and therefore have parity between `action(action, extras)` and `feedback(feedback, bank)` in the end.  I don't think returning a clone of the bank config is a bad thing and by the looks of `bank_changefield` the UI doesn't make changes directly to the object anyway.

krocheck:
My bad ... `deviceid` is action specific, so that's fine to have it added for that type.

Issue #1205, opened 2020-08-24
Show alternate text/visual when button pressed


istnv:
**Describe the feature**
Add a secondary button visual that would display when a button is down/pressed.
Include text, alignment, colors and picture.

**Is this platform dependent (windows, mac, ..)?**
No.

**Usecases**
Add a (different) dynamic variable the the 'down text' :
- for data that only needs to be checked periodically.
- for more specific details relevant to a button.
- easily change 'this' button's color when pressed

Currently, a button needs to know where it is (page/bank) to use 'Internal' change bank action. This makes buttons hard to copy.

dnmeid:
Since this is beeing revisited atm, I have this ideea to make button states. Right now a button has only one "state", i.e. a set of colors and text and image. For backward compatibility I would make this state 0 in the future which is the default state. The you could add more states, maybe state 1 with a different text and different colors or a different picture. Now I think of a change state internal action, so you can change to state 1 if the button is pressed and state 0 if the button is released. You could even have more than two states and feedbacks could be used to change a state. Overall a quite elegant and flexible way to enhance buttons. And the coding effort should be not too much.
So what do you think?

Julusian:
This is very similar to what I have been thinking about as a feedback and variables rewrite.

I think this would be good to have as a default state, then have additional override layers which can be selected like you propose.
This way, you can still define things once on the default and selectively change values with switching state. (eg, define the text once, but set a different png in each override state)

Long term, I would like to see this evolve into a more layer based system, with feedbacks set to enable a layer (with logical operators) instead of explicitly emitting a background and foreground colour. But that will become a bit of a mess to toggle like that via actions.

Also can we call this displayState (or displayLayers?) to clarify the purpose of this state

Julusian:
With #1571, this could be achieved with a new internal feedback to detect the buttons pressed state

Julusian:
There is an internal feedback 'When button is pushed' which does this now

https://github.com/bitfocus/companion-module-bitfocus-companion/commit/fa926cc5f12d5704b5347511664601b5cd1f1c3f

istnv:
@Julusian I was going to offer to add that after you put in the feedback toggle :)
Well done!

PR #1571, opened 2021-04-20
feat: boolean feedbacks

### LINKED ISSUE #1543: with instant feedback, only change background or textcolor.
**Describe the feature**
If i program an 'instant feedback' i can for some instances select a color for text and background. 
if i want to keep the textcolor for the button and only change the background, this is not possible now.
same with keep backgroundcolor, and only change textcolor.
the feature i would like is to only select backgroundcolor or textcolor to change with instant feedback

**Usecases**
for a ptz-cam i program 8 buttons on a row for presets. when a preset is active i change the color of the text to green, and the rest to white
with tally-arbiter i change the background of this row of buttons to red when this cam is on program, but the buttontext also changes to the preset in this feedback. I would like to keep the color of the active preset-button on green.

Julusian:
Fixes #1543

Feedbacks are currently responsible for returning an object of the style properties that they should change. This means it is not easily possible to choose what properties a feedback will change.

With this change, feedbacks can state that they will instead return a boolean value (the existing flow is called advanced, as it is still useful for advanced feedbacks such as showing a generated png).
These 'boolean' feedbacks then have their styling effect managed by core. The ui can dynamically change the properties being set based on what the user selects.

image
(ui might want some attention to make it pretty, I have gone for functional and easy to implement.

### Before
image

### After
image

This is a start towards some of the ideas in https://github.com/bitfocus/companion/issues/1205

### DIFF
--- a/lib/feedback.js
+++ b/lib/feedback.js
@@ -197,27 +197,17 @@ function feedback(system) {
 								}
 
 								try {
-									function handleResult(result) {
-										if ((definition && definition.type === 'boolean') || typeof result === 'boolean') {
-											if (result) {
-												self.setStyle(page, bank, i, { ...feedback.style })
-											} else {
-												self.setStyle(page, bank, i, {})
-											}
-										} else {
-											self.setStyle(page, bank, i, result)
-										}
-									}
-
 									// Ask instance to check bank for custom styling
 									if (
 										definition !== undefined &&
 										definition.callback !== undefined &&
 										typeof definition.callback == 'function'
 									) {
-										handleResult(definition.callback(feedback, bank_obj, { page: page, bank: bank }))
+										const result = definition.callback(feedback, bank_obj, { page: page, bank: bank })
+										self.setStyle(page, bank, i, definition, feedback, result)
 									} else if (typeof instance.feedback == 'function') {
-										handleResult(instance.feedback(feedback, bank_obj, { page: page, bank: bank }))
+										const result = instance.feedback(feedback, bank_obj, { page: page, bank: bank })
+										self.setStyle(page, bank, i, definition, feedback, result)
 									} else {
 										debug('ERROR: instance ' + instance.label + ' does not have a feedback() function')
 									}
@@ -534,8 +524,16 @@ feedback.prototype.unsubscribeFeedback = function (feedback) {
 	}
 }
 
-feedback.prototype.setStyle = function (page, bank, index, style) {
+feedback.prototype.setStyle = function (page, bank, index, definition, feedback, style) {
 	var self = this
+
+	if ((definition && definition.type === 'boolean') || typeof style === 'boolean') {
+		if (style) {
+			style = { ...feedback.style }
+		} else {
+			style = {}
+		}
+	}
 
 	if (self.feedback_styles[page] === undefined) {
 		self.feedback_styles[page] = {}
--- a/lib/feedback.js
+++ b/lib/feedback.js
@@ -197,17 +197,27 @@ function feedback(system) {
 								}
 
 								try {
+									function handleResult(result) {
+										if ((definition && definition.type === 'boolean') || typeof result === 'boolean') {
+											if (result) {
+												self.setStyle(page, bank, i, { ...feedback.style })
+											} else {
+												self.setStyle(page, bank, i, {})
+											}
+										} else {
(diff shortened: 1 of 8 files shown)


stevespaw:
I love this concept - what would it take to add - Change the "Title" text as feedback, using the same variable tools as the regular title? This would be huge...

Thanks!

Julusian:
@stevespaw that should already work as part of this. Any of these properties can be changed:
image
And the ui is reusing the same input logic as above, so it should give the same variable suggestions as the normal button text field (I havent tested to confirm)

krocheck:
@stevespaw you can already change the text as part of feedback

stevespaw:
> @stevespaw you can already change the text as part of feedback

@krocheck   Hmmm, I don't know if it is already there in the latest build. I would be looking for - add feedback - On a feedback with Color choices?  Example may be when it is in record - text change to "recording".  Maybe some method I am not aware of ??

Julusian:
@stevespaw This isnt in any builds yet. It is waiting on a review before it gets merged.
It also requires modules to support this new way of doing feedbacks, so wont be immediately available for most modules

I did forget to rename the feedback I used to demo this, that should really be called `atem: Change from one ME program source`

stevespaw:
@Julusian That what I was thinking, I just wanted to bring up the text idea.  So you are thinking that all of the "main" button pieces, could be in a "feedback"  - in an module where they may be using just color choices for feedback now ???   That is AWESOME.....

krocheck:
> 
> @krocheck Hmmm, I don't know if it is already there in the latest build. I would be looking for - add feedback - On a feedback with Color choices? Example may be when it is in record - text change to "recording". Maybe some method I am not aware of ??

@stevespaw if you return `{ text: "new text" }` instead of something like `{ color: self.rgb(255,255,255), bgcolor: self.rgb(0,0,0) }` that will change the text of the button.  That functionality has been around for a very long time.  I use it in `shure-wireless` for the `sample` feedback because it's all dynamic text controlled by parameters in the feedback.

https://github.com/bitfocus/companion-module-shure-wireless/blob/d9e2b05a5be9c3753f81932a266418d6b1090e7c/feedback.js#L110

Julusian:
> @Julusian That what I was thinking, I just wanted to bring up the text idea. So you are thinking that all of the "main" button pieces, could be in a "feedback" - in an module where they may be using just color choices for feedback now ??? That is AWESOME.....

Yep, exactly that. In the screenshots you can see I randomly chose to change the PNG and the background, but it works for any of the fields.

@krocheck What you are showing there is a very nice way of a module taking over the drawing to do something complex, which is a very different use case. It is true that modules have been able to set in feedback for ages, very few do because 90% of the time a colour change is what is wanted

lib/feedback.js
```diff
@@ -197,17 +197,27 @@ function feedback(system) {
 								}
 
 								try {
+									function handleResult(result) {
```
krocheck: can you make this a class function?

krocheck:
How would we permit an 'upgrade' for developers wanting to switch from `advanced` to `boolean` feedback types?

Would this open the door to having feedback used as triggers for this type by not providing the style options but instead letting banks be triggered?

Julusian:
> How would we permit an 'upgrade' for developers wanting to switch from `advanced` to `boolean` feedback types?

I think we could write a help method on instance_skel for creating upgrade scripts, and documentation/examples on using that.
I don't think we can try and do anything too automatic, due to the chance of non-standard naming and presets needing to be updated to match too.

I am a little concerned about the amount of work it will take to move modules to use this. It would be good to convert everything possible to give a consistent feel across all modules, but it can be done gradually 

> Would this open the door to having feedback used as triggers for this type by not providing the style options but instead letting banks be triggered?

Yes, this is partly motivated so that feedbacks can be used in the triggers panel, but not solely for that.

I see this as the start of being able to use feedbacks to be logical operations. This could be for limiting when actions will run, doing an 'and' on multiple feedbacks to apply one style change, or something else.

lib/feedback.js
krocheck: @Julusian if this gets addressed, I'm otherwise good with this PR

Julusian: done. Ive pushed the logic to inside of setStyle
I need to retest it, and look into an upgradeScript helper, then I shall merge it

PR #1577, opened 2021-04-22
Triggers: Fixes variable conditions

onfire4g05:
This fixes a bug where variables with conditionals would never fire since it was looking for the currently changed variable only vs all the variables to make sure they matched.

### DIFF
--- a/lib/schedule/variable.js
+++ b/lib/schedule/variable.js
@@ -43,15 +43,44 @@ class variable extends plugin_base {
 	matches(label, key, value) {
 		const var_check = `${label}:${key}`
 
-		this.watch
-			.filter((x) => {
-				if (Array.isArray(x.config)) {
-					return x.config.every((x) => this._check_variable(var_check, value, x))
-				} else {
-					return this._check_variable(var_check, value, x.config)
+		this.watch.forEach((x) => {
+			if (Array.isArray(x.config)) {
+				// Since we're checking multiple vars, we need to make sure one of the ones we're looking for HAS been updated and is valid
+				// If it hasn't, we won't run any further logic
+				let valid_var = x.config.find((x) => x.key === var_check)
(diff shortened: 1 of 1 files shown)


krocheck:
@onfire4g05 can you do a `yarn format` and commit Prettier's changes

onfire4g05:
@krocheck done

cxselph:
@onfire4g05 I tried triggers with your fix that was merged and verified that I could use multiple variables to trigger a button. I was hoping that was going to fix the Time Of Day trigger. This seems to have been broken in an update somewhere. Should your fix be related or would that be a new issue? If its new, I will add it as a new issue. Any help or direction would be helpful. Thanks for your contributions.

onfire4g05:
Hmm, I wasn't aware TOD triggers were broken. 😅 I just checked one of our computers and the TOD ones seem to be working correctly (version is a couple of weeks old tho, before the swap to the new UI).

cxselph:
Yea. I have a computer running 2.2.0 build 3117 that works fine. It is the older interface. That makes sense something possibly happened with that update.

sydneyjd:
Ah, So it is not just me loosing my marbles over this. Apparently all of the timers that were there when I updated Companion to the latest beta two weeks ago are still working, but none of the new TOD timers are ever being triggered.

onfire4g05:
I think it's due to an equality check. I noticed an issue the other night while editing old events, and it changed to ints vs strings. I'll check into it.

cxselph:
Thanks!

sydneyjd:
Has there been any headway on a fix for the TOD triggers?

Julusian:
@sydneyjd yes, that has beeen fixed since build 3326

sydneyjd:
Awesome, thank you @Julusian.

PR #1598, opened 2021-05-14
backport: lib/db load, recovery, and save interval changes

krocheck:
This backports from [feat/registry](https://github.com/bitfocus/companion/tree/feat/registry) various updates to `lib/db`, including:

- ES6 refactor
- Disabling `db_save` in favor of a save interval and `dirty` flag
- Add an auto-recovery of `db.bak` if `db` is corrupt or missing
- Attempts to save a `db.corrupt` copy if corruption is detected
- Adds `fs` checks for both `db` and `db.bak` to support those operations

This class was fairly rigorously tested a ways back.  I ran a small collection of tests and the functionality seems to work as tested earlier.  Changes made in [22c1e4](https://github.com/bitfocus/companion/commit/22c14e9f37a9667296c795039be55c025fbca069#diff-69f9a48e7478eb5df50c35b61f5638a078f274a604fff85d69732c2651ef8b47) have been incorporated

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -89,7 +89,7 @@ class Database {
 			//this.save()
 		})
 
-		this.load()
+		this.loadSync()
 
 		// db defaults
 		if (this.store.userconfig === undefined) {
@@ -128,12 +128,22 @@ class Database {
 		this.dirty = false
 		this.lastsave = Date.now()
 
-		if (withBackup === true && fs.existsSync(this.cfgFile) && fs.readFileSync(this.cfgFile, 'utf8').trim().length > 0) {
+		if (withBackup === true) {
 			try {
-				await fs.copy(this.cfgFile, this.cfgBakFile)
-				this.debug(`${this.name}_save`, `backup written`)
+				const file = await fs.readFile(this.cfgFile, 'utf8')
+
+				if (file.trim().length > 0) {
+					JSON.parse(file) // just want to see if a parse error is thrown so we don't back up a corrupted db
+
+					try {
+						await fs.copy(this.cfgFile, this.cfgBakFile)
+						this.debug(`${this.name}_save`, `backup written`)
+					} catch (err) {
+						this.debug(`${this.name}_save`, `Error making backup copy: ${err}`)
+					}
+				}
 			} catch (err) {
-				this.debug(`${this.name}_save`, `Error making backup copy: ${err}`)
+				this.debug(`${this.name}_save`, `Error checking db file for backup: ${err}`)
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


lib/db.js
```diff
@@ -15,181 +15,374 @@
...
+		this.system.on('db_get', (key, cb) => {
+			if (typeof cb == 'function') {
+				cb(this.getKey(key))
+			}
+		})
 
-		self.db = JSON.parse(data)
-		debug('db loaded')
+		this.system.on('db_save', () => {
+			// Not wired ... interval has control
```
Julusian, line 88: should this behave the same as 'db_dirty'? so that it does definitely save eventually?

lib/db.js
```diff
@@ -15,181 +15,374 @@
...
+	async doSave(withBackup) {
+		const jsonSave = JSON.stringify(this.store)
+		this.dirty = false
+		this.lastsave = Date.now()
 
-		try {
-			await fs.copy(pathDb, pathDbBak)
-		} catch (err) {
-			debug('db_save', 'Error making backup of config: ' + err)
+		if (withBackup === true && fs.existsSync(this.cfgFile) && fs.readFileSync(this.cfgFile, 'utf8').trim().length > 0) {
```
Julusian: these `fs.*Sync` methods should be swapped with their async counterparts. this is to avoid doSave blocking the event loop, and blocking actions from running.

lib/db.js
```diff
@@ -15,181 +15,374 @@
...
+		}
+
+		return out
+	}
+
+	/**
+	 * Attempt to load the database from disk
+	 * @access protected
+	 */
+	load() {
```
Julusian: it would be good to call this `loadSync` to make it clear it is an intentionally synchronous operation

Julusian (commented):
I think this looks mostly good. A couple of tiny bits

lib/db.js
krocheck: Maybe, but in my research a `db_save` always followed a `db_set`, which itself now does a `setDirty()` so it would be redundant.

lib/db.js
krocheck: k

lib/db.js
Julusian: good point. lets leave it like this then

lib/db.js
krocheck: done

lib/db.js
krocheck: this has been updated, and since I read the current DB in to check for an empty file I decided to attempt a parse so we don't backup a corrupted file.  thoughts?

Julusian: I like it :)

Julusian (approved):

PR #1604, opened 2021-05-24
feat: boolean feedbacks as trigger sources

### LINKED ISSUE #651: Feedback triggers other button actions
It would be great if the instance feedback of the ATEM could be used to trigger buttons in Companion.

Example: Video Op is switching on the ATEM as normal. Companion takes the signal and triggers buttons (on page 99 ;-) )which could then trigger all the Tally Apps, TSL, or even Remotecam presets, etc.

### LINKED ISSUE #1562: Feature Request - Trigger off feedback
**Describe the feature**
Be able to create a trigger off not just from a variable, but from feedback from a module

**Usecases**
This would allow the ability to press a button based on the feedback from another module - e.g.
- Mute a source when a media clip has finished playing

Julusian:
Closes #651
Closes #1562

It has ended up with a couple of feedback improvements rolled into one PR.

The primary aim is to use boolean feedbacks as trigger event sources.
image

As part of this, the existing variables implementation has been replaced, as by creating a new feedback type in the internal module, existing triggers can be converted over. This lets us provide a more consistent ui, and allows for using variables and other feedbacks as part of the same trigger source.

Other bits:
1) feedback invalidation has been restructured to reduce the scope of the invalidation.
  When a feedback was changed on a bank in the ui, that would invalidate every feedback belonging to that instance. Now the update is scoped to just the bank that changed, and at times down to just the single feedback which changed
1) additionally, when multiple feedbacks on a bank are being checked, a single `graphics_invalidate` call is now made, instead of for each feedback which changed. While the rendering is cheap, this will help mostly on reducing the data being sent to the ui/remote clients.
1) modules call check multiple feedback types at once through `self.checkFeedbacks('bank_pressed', 'bank_text')`
1) modules can ask companion to check certain feedbacks by id instead of all of a type. `self.checkFeedbacksById('vvbta3jtaD', 'Ba_1C3NF3q')`. This will help reduce the performance impact for feedbacks which used to call `self.checkFeedbacks('bank_pressed')` a lot (eg the incoming 'variable_value' feedback for the internal module which had to invalidate whenever any variable changed). This should be used with the subscription callbacks as the data source.
1) 

I have a concern about the change to feedback handlers.
They now sometimes get null for some parameters, as they don't correspond to a bank. Some feedbacks may not like this, and could throw an error when checked. This should be safely handled so isn't a problem, but is something module authors will need to be aware of.
However, this is a change in the instance_skel typescript typings which has upset most/all of the existing typescript modules. They will all need to be updated to be ok with it, at the same time as merging this


Julusian:
@krocheck I dont know if you want to look over this or how much.

But this will be a little painful to merge due to the typescript interface changes causing the `yarn install` to fail on every typescript module. I am happy to go through them all and update them if we are happy with this feature and the backwards compatible api changes

krocheck:
Yeah, I'll take a look

krocheck:
I have concerns about converting over the variable trigger type to feedback using the internal module.  I think that's a less obvious way to access that functionality.  I've not looked at the underlying performance gains for the change and since the system isn't in a production release its probably forgivable (and you wrote a conversion for existing triggers).  @onfire4g05 may have an opinion.

instance_skel.d.ts
```diff
@@ -60,7 +61,11 @@ declare abstract class InstanceSkel<TConfig> {
 	 * Processes a feedback state.
 	 * @since 1.0.0
 	 */
-	feedback?(feedback: CompanionFeedbackEvent, info: CompanionFeedbackEventInfo): CompanionFeedbackResult
+	feedback?(
```
krocheck, line 64: Is there a reason this isn't in the types files and I see a duplicate in there?

lib/feedback.js
```diff
@@ -21,7 +21,7 @@ var debug = require('debug')('lib/feedback')
 var _ = require('lodash')
 var shortid = require('shortid')
 
-function feedback(system) {
+function feedbacks(system) {
```
krocheck: Don't know that I agree with changing the class name while the file name stays the same ... not a huge deal.  Was there a scope or variable naming reason for this change?

onfire4g05:
I won't be able to look over this until next week, but from my brief overview I think this makes sense and would work, assuming multi conditionals still work. I'll definitely check it out more in depth early next week and can also do some testing with some configs I have setup.

lib/feedback.js
Julusian: This was a mistake. I changed it, and then attempted to undo that but made a typo

instance_skel.d.ts
Julusian: I dont understand what you mean?

Julusian:
> I have concerns about converting over the variable trigger type to feedback using the internal module. I think that's a less obvious way to access that functionality. I've not looked at the underlying performance gains for the change and since the system isn't in a production release its probably forgivable (and you wrote a conversion for existing triggers). @onfire4g05 may have an opinion.

I almost didn't add the `variable_value` feedback and do the conversion, but realised that it will be the first follow up request to be able to do an AND operation across the two types.
Once that feedback existed, removing the variable trigger was a usability thing, as there were two ways of doing the same thing, one which is more flexible and one which is easier to find (and the way all your old triggers will be defined), with no way to convert between the two.

Perhaps to make it more visible we should change it to be the default option, with a condition added? (eg internal:hms = '')

If anything, I expect this feedback way to perform slightly worse, as it has added a bunch more steps to the same check, but not enough to matter. I added the `checkFeedbacksById` to keep the impact as low as possible

@onfire4g05 I think I tested it with multiple conditions, but I shall verify that. I have made sure to check that the conversion from the variable style works fine

instance_skel_types.d.ts
```diff
@@ -176,7 +176,11 @@ export interface CompanionFeedbackBase<TRes> {
 	label: string
 	description: string
 	options: SomeCompanionInputField[]
-	callback?: (feedback: CompanionFeedbackEvent, bank: CompanionBankPNG, info: CompanionFeedbackEventInfo) => TRes
+	callback?: (
```
krocheck, line 179: I was referring to this.  Seems like a duplicate of that other one I referenced

Julusian: oh right.
it kinda is a duplicate, but there are 2 ways to define the handler for a feedback. The type could be shared a bit more, but that makes it harder to read so I dont think its beneficial with how little they should be changing

lib/feedback.js
```diff
@@ -169,60 +150,125 @@ function feedback(system) {
...
+		} else {
+			cb(undefined)
+		}
+	})
 
-						if (type !== undefined && feedback.type != type) {
-							continue
-						}
+	/** Check if a feedback is supposed to be checked, for some constraints */
+	function feedback_check_constraints(constraint, feedback) {
```
krocheck: i'd prefer to just make this a class function now, since that's what I'm going to do when merging into feat/registry

krocheck:
I have a good sense of these changes and I think once I merge into feat/registry there some abstraction that can be done to simplify the scheduler plugin's code to lean more directly on feedbacks.

With your constraints system have you validated that stacked feedbacks still take the stack order even if one in the middle is being targeted for a check?

lib/feedback.js
Julusian: I dont like how this separates the definition from the usage, but ive done it

lib/feedback.js
Julusian: fixed

Julusian:
>I have a good sense of these changes and I think once I merge into feat/registry there some abstraction that can be done to simplify the scheduler plugin's code to lean more directly on feedbacks.

@krocheck I have to avoid any large architectural changes here to minimise your pain. With this I have been finding myself fighting the plugin system a little, as the ui is so specific it made more sense to hardcode it than to try and make it generic. but it may make sense for the rest still.
The time of day and maybe interval could probably be changed into feedbacks too if wanted, but Im not going to do that myself (for now)

> With your constraints system have you validated that stacked feedbacks still take the stack order even if one in the middle is being targeted for a check?

Yes, when on a bank its a direct swap for existing filtering logic. And for the triggers, its also doing some invalidation. order doesnt matter here, because its just combining booleans

krocheck:
> 
> @krocheck I have to avoid any large architectural changes here to minimise your pain. With this I have been finding myself fighting the plugin system a little, as the ui is so specific it made more sense to hardcode it than to try and make it generic. but it may make sense for the rest still.
> The time of day and maybe interval could probably be changed into feedbacks too if wanted, but Im not going to do that myself (for now)
Haha ... don't hold back progress on my account.  I've had some complicated merges already, so do what makes the most sense.

tim-barr:
I look forward to testing this.  Thanks everyone. 
Although an individual test may not care about the sequence of the testing, the sequence of multiple triggers matters and more than one triger may be true and they may need to be prioritized.  Are they executed in the order listed?  Is there a way to alter the sequence?

Julusian:
I believe the order the triggers are executed in is not consistent. 
I am not sure about trying to make them predictable, but it will rely on timing factors out of the control of the triggers system. So it would work in some cases, but not others.

If you need this, could you open a new issue, and it can be discussed/evaluated there

PR #1610, opened 2021-05-27
feat: upgrade scripts via static methods

Julusian:
Tested:
* Import page with missing _configIdx (both internal, and other when disabled and enabled)
* Import page with low _configIfx (both internal and other when disabled and enabled)
* Import page with missing internal/other instance info
* Upgrade instance with static scripts
* Upgrade instance with legacy scripts
* Upgrade disabled instance at startup with static scripts
*


lib/instance.js
```diff
@@ -207,57 +207,119 @@ function instance(system) {
...
 
-					if (typeof self.active[id]._init == 'function') {
-						debug('Running _init of ' + id)
+					let changed = false
+					for (var i = idx + 1; i < targetIdx; ++i) {
+						debug('UpgradeConfig: Upgrading to version ' + (i + 1))
+
 						try {
-							self.active[id]._init()
+							const result = upgrade_scripts[i](upgrade_config, actions, release_actions, feedbacks)
```
krocheck: still has `release_actions` in the parameters?

lib/instance.js
```diff
@@ -207,57 +207,119 @@ function instance(system) {
...
+
+					if (idx + 1 < targetIdx) {
+						// Save config, but do not automatically call this module's updateConfig again
+						// Either only _configIdx changed, or the module is not active yet
+						self.system.emit('instance_config_put', instance_id, instance_config, true)
 					}
+
+					// If anything was changed, update system and db
+					if (changed) {
+						system.emit('config_save')
```
krocheck: I know it was there before, but this shouldn't be needed ... its the sysconfig (ip, port, etc)

lib/instance.js
Julusian: I realised that yesterday. The way the scripts are called in this needs some work as I forgot to account for the difference in method signatures.
Ill fix it when I pick this up again

lib/instance.js
```diff
@@ -435,6 +501,22 @@ instance.prototype.activate_module = function (id, modin) {
...
+			self.active[id].disableInstanceUpgradeScripts()
+
+			const scripts = mod.GetUpgradeScripts()
+			if (scripts !== undefined && scripts.length > 0) {
+				// New instances do not need to be upgraded
+				self.store.db[id]._configIdx = scripts.length - 1
+			}
+		}
+
 		if (self.active[id]._versionscripts !== undefined && self.active[id]._versionscripts.length > 0) {
```
krocheck: We may have an existing bug here.  This `if`'s internal says `// New instances do not need to be upgraded` but `activate_module` is also called by `instance_enable` and if an instance was disabled on startup and later enabled it would present the same and have its `_configIdx` overwritten by this block?

Julusian: yep, looks like you are right there. I have a fix for that now

Julusian:
@krocheck I am happy with this now
I have updated the description with the flows I have tested, can you think of any cases I have missed?

I have kept the new implementation very separate from the old upgrade flow, as while bits may be reusable we want to remove the old flow asap so this avoid the task of having to untie the code later

I tested with a couple of modules, whose changes are https://github.com/bitfocus/companion-module-bmd-atem/tree/feat/static-upgrade-scripts and https://github.com/bitfocus/companion-module-bitfocus-companion/tree/feat/static-upgrade-scripts

krocheck (approved):

Issue #313, opened 2018-10-29
Headless Raspberry Pi support


jpkelly:
Starting Companion on boot of a Raspberry PI without the need for a display connected or logging in to a VNC or SSH session.

Preferably this would be enabled via a systemd service.

Some basic config options exposed to the command line such as setting the network interface

willosof:
This is right around the corner, but it will for now require that you clone the repository and follow the developer guide, and not start it via the normal prebuilt packages.

willosof:
Did you make this @haakonnessjoen ?

haakonnessjoen:
There is a alpha version in the source repo: headless.js. But currently it isn't that straightforward to use for the end user, since you would need to install node, and run the update scripts before you can use it. So I think we need to do something clever before we close this, and also more testing.

depili:
What would be the easiest way to import the configuration for the headless mode instance?

haakonnessjoen:
On raspberry pi? Try cp ~/.config/companion/db ~/companion/ if I don't remember wrong.

The most user friendly way would be to start it normally, export the full config as file, and then run headless and import it again, using the gui.

depili:
Getting the headless mode to run requires:
1. installing the dev environment, which has it's own issues on a fresh rasbian: #478 and #479 
2. running ./tools/build_writefile.sh
3. copying the configuration to ~/companion/
4. running ./headless.js wlan0 or ./headless.js eth0

depili:
After all that I got the headless.js running. With osc data updates coming in every 30ms from clock-8001 the update rate on buttons with variables lags 1-4 seconds behind and after about 1,5 minutes something just gives up and the stream deck goes all black. 

In the all black state pressing a button on the stream deck causes it to return to the logo display after a small delay.

The companion code continues to run and still prints osc updates and button gfx generation messages into the console but everything is just dead on the stream deck side.

depili:
Deleting all but the depili-clock-8001 module seems to improve the stability considerably. Now at over 6 minutes and counting where as before it never stayed stable for 2 minutes. The updates still lag behind and arrive in clumps with the display often skipping some second display updates.

haakonnessjoen:
People have reported that installing the raspian lite image makes usb stable with stream deck using headless.js. Not sure about latency though.

depili:
[...]

PR #1613, opened 2021-05-29
feat: Implement fillImage retries for elgato-mini and elgator-xl

bradleywehmeier:
The standard lib/usb/elgato.js has a mechanism for retrying the `fillImage` call a couple of times before giving up on the USB device. This Pull Request applies that retry to the elgato-mini and elgato-xl.

The original retry logic was added as a result of a discussion on Issue https://github.com/bitfocus/companion/issues/313#issuecomment-524150012
In our use case, we had our Mini disconnect from its RPi4 when several image updates were made in quick succession. After running for a month now, we've seen debug logs where it would have disconnected before, but with the patched code it continues working just fine.
Note: I've tested with a Mini but not the XL since I don't have an XL but the code is otherwise identical.

### DIFF
--- a/lib/usb/elgato-mini.js
+++ b/lib/usb/elgato-mini.js
@@ -20,13 +20,18 @@ var debug = require('debug')('lib/usb/elgato_mini')
 var elgato_base = require('./elgato-base')
 var sharp = require('sharp')
 var image_write_queue = require('./image-write-queue')
+var setTimeoutPromise = util.promisify(setTimeout);
 
 function elgato_mini(system, devicepath) {
 	var self = this
 
 	elgato_base.apply(this, [system, devicepath, 'Elgato Streamdeck Mini'])
 
-	self.write_queue = new image_write_queue(function (key, buffer) {
+	var image_filler = function(key, buffer, attempts) {
+		// null/undefined => 0
+		attempts = ~~attempts
+		attempts++
+
 		return sharp(buffer, { raw: { width: 72, height: 72, channels: 3 } })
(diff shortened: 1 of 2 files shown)


Julusian (commented):
It would be better to do the retry from the point where we have `newbuffer`.
This change isnt guarding against sharp throwing an error, which I think is correct, but each time we retry it is sending the buffer back through sharp to do the same scaling.

Have you used async/await in js before? This function could really do with being rewritten to use that syntax and doing so would make the function possible to do as a while loop instead of chaining function calls

bradleywehmeier:
Good suggestion on the aysnc/await. I was trying to keep the change minimal, but I agree the async/await is more maintainable than chaining function calls.

lib/usb/elgato-mini.js
```diff
@@ -20,30 +20,42 @@ var debug = require('debug')('lib/usb/elgato_mini')
...
+				.raw()
+				.toBuffer()
+		} catch (e) {
+			self.system.emit('log', 'device(' + self.serialnumber + ')', 'debug', `scale image failed: ${e}`)
+			self.system.emit('elgatodm_remove_device', self.devicepath)
+			return
+		}
+
+		let attempts = 0
+		while (true) {
```
Julusian: Please make this be a for loop.
`for (let attempts=0; attempts < 3; attempts++) {`

`while(true)` is a fairly risky line of code, and much more likely to accidentally turn into an infinite loop when someone comes back here to make changes in the future.

The rest of the async changes look good :)

PR #1622, opened 2021-06-05
Update package.json

astellingwerf:
While `npm install`ing on a freshly installed Mac OS X 10.15.7, I ran into the issue that the tag name for barco-pulse has a `v` in front of it, that is not mentioned in the `package.json` file.


Julusian:
While you are correct in that this is wrong, it looks to only be a problem when using `npm install` and not `yarn install`. As we only support yarn (and using npm will introduce some subtle dependency issues), I don't think we need to change this.
Instead I have pushed a change that will make `npm install` explicitly fail immediately instead of maybe mostly working

Issue #1449, opened 2021-02-24
Allow a per button option to remove the buton number bar


JoshWobbles:
Ive seen this shot down elsewhere with the response that some actions 'need' the bar, or its helpful etc... that's fine, but there are plenty of cases where it isn't needed so why not give us the option to turn it off on a button by button basis?

josephdadams:
related to #981

PR #1628, opened 2021-06-11
feat: add userconfig interfaces to instance_skel

krocheck:
The "Remove Topbar" feature introduced some issues for modules dynamically drawing the canvas in feedbacks and injecting a raw pixel buffer for rasterization (e.g. `shure-wireless` and `studiocoast-vmix`).  The modules needed the ability to get and subscribe to the `remove_topbar` setting to change raster dimensions and placements.  This also required some changes/fixes to the remove topbar implementation to accommodate.

New `instance_skel` functions:

- `getUserSetting(key, cb)`  - get the value of a userconfig setting, returned via callback
- `subscribeUserSetting(key, cb)` - same as get but will provide return via callback whenever the setting changes
- `clearUserSettingSubscriptions()` - manually clear/stop all user setting subscriptions
- (there is additional 'behind the scenes' implementation in `instance_skel`, including a `destroy()` wrapper that executes the clear)

Other changes:

- Topbar rasterizing was moved to a function so it can be called at the end of drawing like normal or during a png64 or img64 exception so that it is always present
- Added img64 rendering try-catch
- Added `get_userconfig_key` system call to do the single setting fetch
- Added `feedback_check_all` system call
- Added run of `feedback_check_all` in `lib/graphics` on remove topbar change prior to re-rendering all banks; moving both to a 1s delay so modules can make adjustments

Sample use-case is in `shure-wireless` [here](https://github.com/bitfocus/companion-module-shure-wireless/commit/0d7e0ca8b5bff6b98328d16c51e33fd46df7348a)

### DIFF
--- a/instance_skel.d.ts
+++ b/instance_skel.d.ts
@@ -108,6 +108,8 @@ declare abstract class InstanceSkel<TConfig> {
 	setVariables(variables: { [variableId: string]: string | undefined }): void
 	/** Get the value of a variable from this instance */
 	getVariable(variableId: string, cb: (value: string) => void): void
+	/** Get the value of a userconfig setting */
+	getUserSetting(settingId: string, cb: (value: any) => void): void
 	/** Recheck all feedbacks of the given types. eg `self.checkFeedbacks('bank_style', 'bank_text')` or `self.checkFeedbacks` */
 	checkFeedbacks(...feedbackTypes: string[]): void
 	/** Recheck all feedbacks of the given ids. eg `self.checkFeedbacksById('vvbta3jtaD', 'Ba_1C3NF3q')` */
--- a/instance_skel.d.ts
+++ b/instance_skel.d.ts
@@ -116,7 +116,7 @@ declare abstract class InstanceSkel<TConfig> {
 	/** Get the value of a userconfig setting */
 	getUserSetting(settingId: string, cb: (value: any) => void): void
 	/** Subscribe to changes for a userconfig setting */
-	subscribeUserSetting(key: string, cb: (value: any) => void): void
+	subscribeUserSetting(settingId: string, cb: (value: any) => void): void
 	/** Manually clear/stop all user setting subscriptions */
 	clearUserSettingSubscriptions(): void
 
--- a/instance_skel.d.ts
+++ b/instance_skel.d.ts
@@ -108,12 +108,17 @@ declare abstract class InstanceSkel<TConfig> {
 	setVariables(variables: { [variableId: string]: string | undefined }): void
 	/** Get the value of a variable from this instance */
 	getVariable(variableId: string, cb: (value: string) => void): void
-	/** Get the value of a userconfig setting */
-	getUserSetting(settingId: string, cb: (value: any) => void): void
 	/** Recheck all feedbacks of the given types. eg `self.checkFeedbacks('bank_style', 'bank_text')` or `self.checkFeedbacks` */
 	checkFeedbacks(...feedbackTypes: string[]): void
 	/** Recheck all feedbacks of the given ids. eg `self.checkFeedbacksById('vvbta3jtaD', 'Ba_1C3NF3q')` */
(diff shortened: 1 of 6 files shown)


krocheck:
squash and merge

Julusian:
I have mixed feeling about the api here, what follows is my thought process on it without any real conclusion.

If we are to do https://github.com/bitfocus/companion/issues/1449 or want to natively support different sized buttons (the xl wants 96px vs 15-keys 72px) then we will need to be able to provide the information on a per-feedback basis. Likely as one of the parameters in the callback.
However that does make it harder to do the image generation in advance of the feedback needing it. While not a problem right now, it could be if a module wishes to use sharp to do some stuff as that works async.

If it was part of the callback data, then maybe it would work though, as it could also be fed into the call to subscribe. Of course this will be more work for the module developer, but will be more future proof for when/if we want to go down this route (I am tempted to see how bad adding mixed resolution drawing would be..)

But, maybe this is a worthwhile thing to start with, as this is a flow that a few modules are already trying to handle via internal apis and is solving an immediate problem.

krocheck:
I feel like allowing multi-raster size support and whether the topbar is removed or not are independent issues (unless we intend to make that a per-device setting).  Reason being it seems like a waste to be processing a bunch of images that will only get used if that switch flips.  But if its better for the modules to just start returning 72x58 and 72x72 rasters and then we can add additional rasters later for native support, such as 96x96 for the XL, I suppose that's fine.  Of course, when I math'd it, the same topbar aspect makes the main raster for that size 77.33x96, so maybe native drawing would be better.  That's going to be a mess because we'll have to redo fonts.  It would be a ton more work for me in the module to also support 96x96 because I'd need to make a new icon set.  Unless a hi-res device comes along that doubles the resolution, I have little interest in doing that.  And we'd have to deal with people being able to upload PNGs for different raster sizes as well?

I think this interface is import to have (module read access to userconfig) even if we switch to multiple raster support later that would change the current use-case to something else.

Julusian:
> I feel like allowing multi-raster size support and whether the topbar is removed or not are independent issues (unless we intend to make that a per-device setting). Reason being it seems like a waste to be processing a bunch of images that will only get used if that switch flips. But if its better for the modules to just start returning 72x58 and 72x72 rasters and then we can add additional rasters later for native support, such as 96x96 for the XL, I suppose that's fine.

I wouldn't want to generate excessive rasters. I am thinking that if we need multiple sizes of something, then we do multiple calls to the feedback callback. With the 3rd parameter becoming:
```
{
  deviceId: 'serial',
  page: 1,
  bank: 10,
  width: 72, // add this
  height: 58 // add this
}
```
Modules will need to be able to handle any resolution acceptably well (eventually). To begin with they can assume it to be just the two.  

> That's going to be a mess because we'll have to redo fonts.

Yeah, but bitfocus did pay me a while ago to get https://github.com/ericfreese/node-freetype2 working better for another project (I assume for buttons), so it would be worth trying to switch to that.

> It would be a ton more work for me in the module to also support 96x96 because I'd need to make a new icon set. Unless a hi-res device comes along that doubles the resolution, I have little interest in doing that.

I can understand that. But you could handle that by generating in 72x72, and scaling the result. It will be more code in the module to do that, but means that core is free to do what it wants with resolution

> And we'd have to deal with people being able to upload PNGs for different raster sizes as well?

Eventually yes, but that will probably just be letting them upload any resolution and scaling it down. I think we already handle this, and the ui is an artificial limit to try and keep db/export sizes down

> I think this interface is import to have (module read access to userconfig) even if we switch to multiple raster support later that would change the current use-case to something else.

I don't see anything else in the userconfig that makes sense for modules to be aware of. And giving blanket access like that makes me concerned about breaking something when a property gets removed


I think this comes down to whether we do something which is enough for now, or if we want to be futureproof. The cost being either very little work for modules right now, and they will be very broken when the long plan is done, or more work right now, and less later.

krocheck:
I'll work this alternate idea

PR #1629, opened 2021-06-11
feat: switch remote control info to tabs

krocheck:
Changes the "Remote Control" information area on the right side of the "Settings" tab to a tabbed interface.


Julusian (approved):
Haven't run it, but the code looks sane. (I'm so glad github had hide whitespace changes)

Issue #1626, opened 2021-06-09
Enable the Internal Button Press button action to select PRIOR or NEXT 


ltjost:
**Describe the feature**
Add NEXT REGULAR and PRIOR REGULAR Button options to the selections allowed for the "internal:Button Press" action or as new internal actions [edit: see comment EXECUTE NEXT CUE and EXECUTE PRIOR CUE commands info]

**Usecases**  
 [edit: see comment EXECUTE NEXT CUE and EXECUTE PRIOR CUE commands info]
I have tested Companion with SCS and with an x32 and with the OSC module.  I use x32Theatre Control and SCS regularly...but it is fairly easy to just manage theater cues directly in Companion and skip x32tc or SCS by use of the OSC commands in Companion.  In audio mixing for theater, it is somewhat typical to create cues that assign channels to DCAs and also name the DCA at each Cue.  It is also fairly normal to mute channels not in DCAs or manage mutes at the DCA level or however the user desires.  I have written a script interpreter that allows me to enter a large variety of OSC commands for the x32 through an easy to use input via excel spreadsheets (a JavaScript gets the tab delimited excel file and creates SCS OSC commands from it).  Running SCS or x32tc requires focus, unless Companion is used, and adds add another layer of software that is not really needed.  Companion can do more and can do it faster, I just need to create a script to take excel and turn it into a Companion config file, which looks pretty straightforward having already done this for SCS.  However, in running theater cues we don't always have time to navigate to the next cue, rather depend on a NEXT and/or PRIOR capability that executes the next or prior cue.  With Companion, unique buttons can be generated for each of the hundred+ cues in a show...but there is no NEXT or PRIOR internal option for button presses...but there is a command to press a specific button.  Allowing PRESS NEXT REGULAR button would be great, as would PRESS PRIOR REGULAR BUTTON, either as selections for "internal:Button Press" or as unique "internal" actions.  With this, a sound engineer could just keep pressing a NEXT button at the top of the Companion Emulator or Streamdesk device to run show cues for the x32 or other sound board without needing SCS or x32tc.  It would enhance to use-cases for the StreamDeck or emulator to include enabling use of this tool for direct theater cue management. See [SCS.WorkWebs.com](SCS.WorkWebs.com) for the tool I mentioned for creating SCS files from excel documents)

ltjost:
[...]

Issue #1589, opened 2021-05-02
[BUG] Latch/Toggle Doesn't work with Stream Deck Plugin


jkhouw:
**Describe the bug**
"Latch" does not work with elgato streamdeck plugin.  The button executes the key release as soon as the key is released. 

**To Reproduce**

1. Define a button in companion with "latch/Toggle" ticked
2. assign an action to key down/on like change button background color to red
3.  assign an action to key up/off like change button background color to black
4. run Stream Deck software with plugin installed.
5. Press "Test" actions in Companion Admin console - the button will latch and function properly.  
           You can see the button on device, in Companion Admin, and StreamDeck software will all have RED background.
    Press "Test" again to simulate unlatch and 
            You can see the button on device, in Companion Admin, and StreamDeck software will all have BLACK background.
            
6. **Press the physical button on the device** - the button will not remain RED.  If you hold it, it will stay red until you release it.

**Expected behavior**
Expected Latch/Toggle functionality to work with Stream Deck Plugin.  

**Screenshots**
Sample of button config I was testing with.
![image](https://user-images.githubusercontent.com/83513885/116800275-ac42ab00-aacd-11eb-918d-302ca103349c.png)

**Desktop (please complete the following information):**

- OS: Windows 10 Pro/Build 19041.928
- Browser: Edge/Chrome/Firefox
- Companion Version 2.1.3 (6b6820cd-2696)

Julusian:
I cant reproduce this issue in either 2.1 or 2.2. Assuming it has been fixed since this was reported

Issue #1132, opened 2020-06-12
Copy individual or group configuration actions and options


Brennan-Dan:
**Describe the feature**
To be able to copy and paste individual or group of actions and their options from a button configuration page. (not a copy button that is already in the companion)

So if you wanted to copy a certain action or list of actions on a button to add to a list of actions on another button. (saves lots of time having to reinput the came action over and over again if that action is used on lots of buttons. 

**Is this platform dependent (windows, mac, ..)?**
no

**If documentation is required to implement, do you know where to find it?**

**Usecases**
Behringer X32 Talkback system where 1 button can turn on or off multiple channels/buses, change the background color on “host” button and somebody else’s Stream Deck/Page button too.

Anything else using multiple identical commands.

josephdadams:
#712

Mozoloa:
unfortunately this is still not possible and this issue has been closed because it was originally only asking for duplication within a button

> #712

dnmeid:
This feature request is still on the list. It will have to be split in two individual tasks.
1. make copy/paste of actions or feedbacks (or steps or any other part of a button configuration)
2. make this possible for selections of actions or feedbacks
Basically there is a good chance of copy/paste functionality for all items which are dragable and dropable, but this is the groundwork which has to be done first.

Mozoloa:
That would be a huge time saver for copying an action on existing buttons, especially if you can paste several times

MartinM-dev:
it would be great to be able to copy on several buttons at once

PR #1630, opened 2021-06-14
feat: step button type to replace latching

Julusian:
This is a rethink of the latch button behaviour. The code to handle latching has been full of precarious logic, that has been very buggy in the past (eg #995) #1589. It looked and felt like an after thought.

I have ripped out the existing logic and reimplemented it with some additional functionality.

First off, there is a slight behaviour change in that previously the release actions would be executed on button up, they are now executed on the down press. This was the source of a large amount of the complication.

Instead of latching, there is now a step button type. This is no longer simply a checkbox in the button styling, but a new option in the button type menu:
image

Once you have selected this 'step/latch' option, the actions tables are a little different to before. To begin with you have one called 'Step 1 actions', and there is an option below to add another. With this, you are not limited to having a latch, but can have a 3 or 4 way cycle. In fact, there is no limit. You could build a cue list as steps on a single button, and keep pressing the same button again and again #1626.

For example, this will cycle through 3 inputs on an atem:
localhost_3000_(1080p)

Also note, that on the right of the table heading is a group of buttons which allow for re-ordering and removing steps, as well as setting which one will be executed next (indicated in green)

Internal actions and feedbacks have been added ([separate PR](https://github.com/bitfocus/companion-module-bitfocus-companion/compare/feat/action-steps)):
image
image

Presets expect a slightly different structure, to match the changes. The 'old' format is handled and converted across, following the same rules as existing buttons being upgraded
Regular button:
image
Stepped button:
image

It isnt currently possible to convert between a regular and stepped button, due to the differing number of action sets they can have.

On the technical side, this changes the db format to have a single structure of 'action_sets' instead of 'actions' and 'release_actions'. Old databases and exports will be upgraded appropriately.
Presets in the old format are also 'upgraded' on the fly, so that they will work correctly.


ltjost:
Thank you for cross referencing to my prior/next cue concepts.  I look forward to exploring this.  One item I am having a difficult time understanding through just reading this information is how the user will be able to jump to specific step live in case the planned sequential cue execution sequence is not sufficient to address the actual experience presented.  For example...in running a theatrical event, the cast skips ahead and/or then jumps back and the sequence needs to be adjusted on the fly...or at a rehearsal the sequence is totally not part of the plan.  Adding and removing cues seems to have been handled as has moving them around, which is excellent.  Again, looking forward to this!  Thanks.

josephdadams:
@Julusian add "copy action" while you are at it :) #712 and #1132

Julusian:
@josephdadams haha, soon. Once I start on that, it will end up with a bunch of changes in the area of button action management

@krocheck I am happy with this now. I have tested everything I can think of. Including importing some full and page exports I found on old issues.
The only thing I have left is to figure out a plan for upgrading existing 'is pressed' feedbacks if they were targeting a latch. I should do something, which will involve an upgradeScript and some extra data for the internal module to figure thing out.

Julusian:
@MeestorX 

> Hoping that regular actions can be moved in and out of step actions,

I dont follow what you mean here. A button is either regular or stepped, and the actions are intentionally not aware of which it is. (I have plans for another new type of button)

> (along with copy/cut/paste of regular and step actions!).

That is another task on my growing todo list. Fairly near the top too

> I'll be interested to see how to create presets with step actions as well.

I have added some screenshots to the description showing the new format for writing presets

MeestorX:
> @MeestorX
> 
> > Hoping that regular actions can be moved in and out of step actions,
> 
> I dont follow what you mean here. A button is either regular or stepped, and the actions are intentionally not aware of which it is. (I have plans for another new type of button)

Ah. I thought you have an action that happens on every push along with a step action. Didn't realize it was one or the other, thought you could combine them on one button.

However, you will need to be able to create step actions groups on one button, though e.g.
Group 1:
Step 1: Mute Ch 1
Step 2: UnMute Ch1

Group 2:
Step 1: Mute Ch 2
Step 2: UnMute Ch2

So that the 1st push Mutes Ch1 & Ch2, 2nd push Unmutes Ch1 & Ch2. Or maybe that's done a different way?

jrviz:
Is there a derivation of this to allow a short, long and in-between timed button press structure rather than a sequence? e.g. use a timer value on button release to identify the Step set to trigger? 

I know that this is a different feature and has been requested but it may be that this new feature would be a good starting point.

Would need care on how delays are handled (perhaps they are always relative to guarantee a time sequence) , but I guess there is a similar issue for this button type if a step set has a long delay in it and a users keeps pressing the same key to stack up delayed steps as it sequences around.

Julusian:
> Is there a derivation of this to allow a short, long and in-between timed button press structure rather than a sequence? e.g. use a timer value on button release to identify the Step set to trigger?
> 
> I know that this is a different feature and has been requested but it may be that this new feature would be a good starting point.

There is not. The structures introduced here will allow for that to be done, but I have not started on it yet. It is on my list and I have been thinking about how it will work, but I am holding off until this has been merged

krocheck:
@Julusian for the feedback issue, add a checkbox on the left side under delay for "Indicate Bank as Pressed?" or something like that.

krocheck:
@Julusian I was able to work through a merge.  I'm working on some post-merge errors and cleanup.

Julusian:
@krocheck I fixed the feedback backwards compatability with a new option on the feedback (set in existing feedbacks via an upgrade script)
image
It treats the button as pressed if it is the step type and not on the first step.

A very quick test of this pr post your merge and it appears to be working

krocheck:
This feature will be in a future release of Companion, but is currently only in the development/experimental track.  There are no accessible downloads of that track at this time.  Eventually it will be available in the public betas and thereafter a stable release.

PR #1634, opened 2021-06-17
feat: custom variables

Julusian:
Inspired by @McHauge's generic-variable module

A new 'Custom variables' button has been added to the variables tab
image

Inside, you can create and edit custom variables
image

These all get exposed as `internal:custom_something` (with `something` being the name you provided)
At startup the variable will be set to the startup value.

There is an action in the [internal module](https://github.com/bitfocus/companion-module-bitfocus-companion/tree/feat/custom-variables) to set the value of a custom variable
image

I haven't tried to do anything too fancy yet, (such as setting over http) I am hoping for that to trickle in from people who know what is wanted more than I do. This PR is mostly about creating the basic structure and flow which can be expanded upon.


krocheck:
Custom variables works better in my mind as its own class, but that's not worth addressing here.

Do you want to include these in exports and imports?  Export page might want to detect and include, similarly to how instances get included.  You'll need to do a reset in the `reset everything` call.

Julusian:
> Custom variables works better in my mind as its own class, but that's not worth addressing here.

Maybe. To me it feels intertwined enough and not enough code to justify separating it out.

> Do you want to include these in exports and imports? Export page might want to detect and include, similarly to how instances get included. You'll need to do a reset in the reset everything call.

Yeah they probably should be included in exports. Its going to be tedious to look through everything, and is going to have to inspect 'internal' actions, but very doable. As long as we don't want to show anything in the ui for it, as that will become a pain to figure out how to present it.

Good point, I forgot about reset all. Even though I added some other things to that yesterday...

krocheck:
Can you work on the import/export for at least full config and reset.

Julusian:
> @krocheck Can you work on the import/export for at least full config and reset.

Done for full import/export.

I think that per-page will need some thought on how to implement it. I am tempted to expose the ability for any module to set these variable values, which will massively complicate export. I dont know if that is useful or a good idea though

krocheck:
> I am tempted to expose the ability for any module to set these variable values, which will massively complicate export. I dont know if that is useful or a good idea though

I don't think that's a good idea at the moment.  Can put my finger on why except that it seems very 'can-of-worms'.

Cool beans on the full import/export/reset!

krocheck (approved):

Julusian:
> I don't think that's a good idea at the moment. Can put my finger on why except that it seems very 'can-of-worms'.

Yeah it does. Something I shall think on

McHauge:
Hi this is great, just did some testing on it yesterday, and the only thing currently that I feel it's missing is these two actions/functions in the internal module (picture below), other than that it's a great step, and enables stuff like radial buttons and more :)

image

McHauge:
@Julusian and @krocheck I have added those two things to the Custom_Variables branch of the internal module, so I'm happy with it now, and I can't really see anything missing, for a first release for public testing.

PR #1643, opened 2021-06-28
Use button offset for Streamdeck Mini

istnv:
When a Streamdeck Mini (6 button) controller is attached, this adjusts the buttons to start at column 2 so they will match the 'mask' displayed in the GUI.

### DIFF
--- a/lib/usb/common.js
+++ b/lib/usb/common.js
@@ -30,6 +30,9 @@ common.prototype.toDeviceKey = function (key) {
 		return key
 	}
 
+	// adjust column for mini
+	key -= self.streamDeck.MODEL == 'mini' ? 1 : 0
+
 	if (key % global.MAX_BUTTONS_PER_ROW > self.keysPerRow) {
 		return -1
 	}
@@ -51,6 +54,9 @@ common.prototype.toGlobalKey = function (key) {
 
 	var rows = Math.floor(key / self.keysPerRow)
 	var col = key % self.keysPerRow
+
+	// adjust column for mini
+	col += self.streamDeck.MODEL == 'mini' ? 1 : 0
 
 	return rows * global.MAX_BUTTONS_PER_ROW + col
 }


Julusian:
While this is technically a bug, it has been there since the mini has been supported (I think?), so changing it now is technically a breaking change. Anyone already using a mini will have to adjust their config to compensate.
As such, I'm not sure if we should change this, or if we can do something to minimise disruption.

Julusian (changes requested):
I think should be done in usb/elgato-mini.js by overriding the methods and applying the offset either before or after calling into the method on the parent class.

istnv:
It did work in 1.4, and got lost in the re-write to 2.0.0. I've added this to every version since.
I did my survey on Facebook and there are very few Minis being used, but it would need an announcement.
7 out of 330 responses use the Mini.
It has been broken for so long now it should be a configuration option to 'Apply offset to mini?'

Julusian:
> It has been broken for so long now it should be a configuration option to 'Apply offset to mini?'

I am wondering that. We could make upgraded configs not apply the offset, but new installs could by default. 
Let me think on it. I am wondering if it would be worth giving more options for placement at the same time

istnv:
>I am wondering if it would be worth giving more options for placement at the same time
That could be very useful. 
If the Mini was in the Upper Right or Bottom Right corner and you could have a 15 key and 6 key on the same page without any overlapping buttons.

josephdadams:
I use mini's, and it would not be a big deal at all to reprogram. I'm one of the 7. :) I second the upper/lower right, it would be nice to use them in that way.

krocheck:
The reason for starting on column two is because the default page buttons in column 1 are non functional given the lack of a third row.  As a note for anyone making a PR to allow this to be adjusted per-device, the pin code system is setup for numbers 1-6 (i believe) to be displayed in the 6 buttons (obviously limiting the usefulness but still functional), so the adjustment will need to apply to that rendering code as well.

istnv:
@krocheck This PR does covers both which button is pushed and which image to render on the button.

krocheck:
Oh ... was on my phone and thought I click into an issue, not a PR from Slack.  Cool beans!

krocheck:
Yeah, this is a breaking change, so you'd need to do a user configurable or device specific option to change the behavior.  From my chair the current behavior is working as intended, as it was never a bug when we converted from 12 > 15/32 because it maintained the functionality from the 12 (fix page buttons) / < v2.0 architecture.

istnv:
> ... as it was never a bug when we converted from 12 > 15/32 because it maintained the functionality from the 12 (fix page buttons) / < v2.0 architecture.

It did not maintain function and hasn't worked 'as described' since the XL support was added, but there are so few Minis being used it was not a big issue.

I have a diff-apply script I use to add this patch every time I re-sync to companion base.
When this 'bug' report was posted, I offered that patch as a solution.

When I have time I will move this to the mini-specific driver as suggested. This can wait to be merged until a way to add user selection is implemented (I like the configurable offsets).

krocheck:
> It did not maintain function and hasn't worked 'as described' since the XL support was added, but there are so few Minis being used it was not a big issue.

Oh wait, I just audited the code and I had a misunderstanding of what's going on here ... blame in part reviewing bugs/prs on my phone.  Yeah, so the Mini isn't following the 2,3,4 and 10,11,12 mapping like the mask suggests my read shows?

Great, add that to list of oops from v1.3 to v2.  So its a bug that's been in the wild for about a year now.  Obviously making this configurable is ideal, but I don't think it's is a bug at this point even though its definitely not working as originally intended.

istnv:
> ... even though its definitely not working as originally intended.

I never submitted the patch because nobody else seemed to notice :)
Now I'm trying to move the offset into the mini specific driver without completely duplicating the common deviceToKey, keyToDevice functions.

Julusian:
Also worth noting, this change wont change the behaviour of satellite streamdeck minis. Im not sure how to get them offset in a simple/generic fashion

istnv:
Looking at this again.
If there is a future global configuration 'offset' per surface (for non-elgato devices) it does not make sense to bury this in the mini-only driver portion of the code. This patch location is probably a better place to support other devices.

Julusian:
@istnv yeah thats true

What I would like to see is for every surface which is smaller than 8x4 (or maybe even those for simplicity) to have an xoffset and yoffset config fields. And for those properties to be owned and used solely by `lib/device.js`. (like what I did with the page number in #1651).

Side thought, why are the devices the ones to transform the index from local to global? That feels like something that should be done by lib/device.js as it is the same forumla for everything.

By doing it there, it will work the same for every surface and be very generic. The only other thing to handle is setting the default offset for minis, if we even want to (it could get fiddly, especially to handle satellite devices) 

Some thought is needed for the , as ideally it should be made to work on any (within reason) size grid. Perhaps we could even do something better for the mini and have a shift button?

istnv:
At this point, an existing mini should probably not be automatically updated. A new one would probably benefit from the x+1 offset. 
I have built a 'shift' button with page change on press, page restore on release. A lot of my code for this-page, new-page, old-page logic was added to ensure the correct page was always displayed on the buttons.
It would be nice to have that at a more global level.

Julusian:
The offset change might have been intentional, so maybe the default position shouldnt be changed? (I recognise the name from a list of core developers)
image

I still think that having the position be configurable would be good to do

istnv:
Too bad nothing in the documentation was updated either. I have all 3 sizes of streamdeck and it is *much* more convenient to leave the page buttons in place when changing pages on the other decks rather than remember where they got moved on the 'mini' page.

Julusian:
Looks like it was changed here https://github.com/bitfocus/companion/commit/bf498bc3dcbd9b5b9b687e8110a9fbbfc96bef3f#diff-d94817dbeefe2b8d1966cc6d36cfba8c1f458f1e5e593d5f63e1834eceae0074
From the rest of the changes I can't tell if it was intentional.

willosof:
So, what did we want to do with this?

istnv:
> So, what did we want to do with this?

I'd like to see a GUI option to select the offset for any device, which gets used to adjust the buttons. 
I am not that good with the front-end code, though.

fbosman:
I think we must raise companion to a next level:
Decouple the buttons from the Streamdeck layout: define independent buttons with an ID and create mappings from those buttons on specific tablets/streamdecks/web-front-ends.
This will greatly enhance the
Usability and maintenance.

Julusian:
> I'd like to see a GUI option to select the offset for any device, which gets used to adjust the buttons.
> I am not that good with the front-end code, though.

I agree that this is the best solution. Ill get it done when I have a chance.

@fbosman 
> I think we must raise companion to a next level:
> Decouple the buttons from the Streamdeck layout: define independent buttons with an ID and create mappings from those buttons on specific tablets/streamdecks/web-front-ends.

Long term yes, but that is a large and complex change that is too late to achieve in 2.2.

Julusian:
Replaced by #1781

PR #1644, opened 2021-06-28
feat: make udp/tcp/osc server ports configurable

Julusian:
Pretty self explanatory from the screenshot

image

Currently, the defaults match the old behaviour, but perhaps we should disable them by default, to reduce chance of 'security issues'?

I think we should change the tcp and udp port numbers, to be outside of the [Ephermal port range](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Dynamic,_private_or_ephemeral_ports) as that should reduce the issues users have with the ports sometimes being in use (which is resolved by a reboot)


krocheck:
yes, i would recommend they be disabled by default.  I’m ok with moving them out of ephemeral if we ensure an upgrade method that leaves all existing installs as-is.

McHauge:
I like the look of this and would agree that maybe on new installs these should be turned off by default, just because, but old setups should keep them on as they are not.  at least that's my take on it

Julusian:
Thanks both for the input.

So on new installs they are now disabled by default, with udp and tcp suggesting using port 16759. This was randomly chosen, I have no attachment or reason for that particular number. TCP and UDP may as well be the same port, so I have done that.

Due to how early the userconfig is loaded in the process of things, I had to devise a new way to check what values should be set, simply using the knowledge of if a db was loaded or the defaults were used.

krocheck (approved):

PR #1651, opened 2021-07-03
feat: rewrite satellite api

Julusian:
This is a rewrite of the satellite api used by https://github.com/julusian/companion-remote
The existing api has been there since 1.4.0, but until 2.1.1 it had a few breaking bugs:
* limited to 15 keys
* devices could be leaked when clients disconnect
* companion could crash when devices were removed

Additionally, the api is quite hard to use. It is binary encoded data, which makes adding new functionality without breaking clients very difficult. And when implementing a client you have to make sure to encode with the correct endianness and byte packing.

To improve on this, I propose replacing it with a text based protocol. While this does make the protocol less efficient on bandwidth, I dont think it is enough to be a problem and it makes implementation and maintenance of the protocol much simpler.
For 2.2 I propose we run with both apis. This will allow for minimal breakage for users using the existing api. In 2.3 we should remove the old satellite api

## API spec
The server by default runs on port 16622, but this may become configurable in the future. You should make sure to support alternate ports to allow for future compatibility as well as firewalls or router port forwarding.  
Each message is presented as a line, with a `\n` or `\r\n` terminator.  
Messages follow the general format of `COMMAND-NAME ARG1=VAL1 ARG2 ARG3="VAL3 with spaces"\n`. Note the `ARG2` which is shorthand for `ARG2=true`  
Key numbers are in the range of 0-31.  

Upon connection you will receive `BEGIN Companion Version=2.2.0-d9008309-3449` stating the build of companion you are connected to. This should not be relied on to be meaningful to your application, but can be presented as information to the user, or to aid debugging.

### Messages to send
Upon receiving an unknown command, the server will respond with the format `ERROR MESSAGE="Unknown command: SOMETHING"`  
Known commands will get either a success or error response like the following:
* `COMMAND-NAME ERROR MESSAGE="Some text here"\n`
* `COMMAND-NAME OK\n`
* `COMMAND-NAME OK ARG1=arg1\n`

#### Close connection
`QUIT`
Close the connection, removing all registered devices

#### Ping/pong
`PING payload`
Check the server is alive, with an arbitrary payload
Responds with `PONG payload`

#### Adding a satellite device
`ADD-DEVICE DEVICEID=00000 PRODUCT_NAME="Satellite Streamdeck"`
* DEVICEID should be a unique identifier for the hardware device. such as a serialnumber, or mac address.
* PRODUCT_NAME is the name of the product to show in the Surfaces table in the UI

Optional parameters:
* KEYS_TOTAL - number of keys the device has. Must be in the range 1-32 (default 32)
* KEYS_PER_ROW - number of keys per row. Must be in the range 1-8 (default 8)
* BITMAPS - true/false whether you want to be streamed bitmaps for display on the buttons (default true)
* COLORS - true/false whether you want to be streamed colors for display on the buttons (default false)

#### Removing a satellite device
`REMOVE-DEVICE DEVICEID=00000`
* DEVICEID the unique identifier used to add the device

#### Pressing a key
`KEY-PRESS DEVICEID=00000 KEY=0 PRESSED=true`
* DEVICEID the unique identifier used to add the device
* KEY number of the key which is pressed/released
* PRESSED true/false whether the key is pressed

### Messages to receive
No responses are expected to these, and to do so will result in an error.

#### Ping/pong
`PING payload`
The server is checking you are still alive, with an arbitrary payload
You must respond with `PONG payload`

#### Pixel buffer for key
`KEY-DRAW DEVICEID=00000 KEY=0 DATA=abcabcabc`
* DEVICEID the unique identifier of the device
* KEY number of the key which the pixel buffer is for
* DATA base64 encoded pixel data. Currently 72x72 pixels of 8bit RGB (this may be configurable in the future)
Note: this is only sent for devices which were added where `BITMAPS` was true

#### Color change for key
`KEY-COLOR DEVICEID=00000 KEY=0 COLOR=#00ff00`
* DEVICEID the unique identifier of the device
* KEY number of the key which the pixel buffer is for
* COLOR hex encoded 8bit RGB color for the key
Note: this is only sent for devices which were added where `COLORS` was true

#### Reset all keys to black
`KEYS-CLEAR DEVICEID=00000`
* DEVICEID the unique identifier of the device

#### Change brightness
`BRIGHTNESS DEVICEID=00000 VALUE=100`
* DEVICEID the unique identifier of the device
* VALUE brightness number in range 0-100

### DIFF
--- a/app.js
+++ b/app.js
@@ -184,6 +184,7 @@ system.ready = function (logToFile) {
 	var loadsave = require('./lib/loadsave')(system)
 	var preset = require('./lib/preset')(system)
 	var satellite = require('./lib/satellite_server')(system)
+	var satellitev2 = require('./lib/satellite/satellite_server_v2')(system)
 	var ws_api = require('./lib/ws_api')(system)
 	var help = require('./lib/help')(system)
 
--- a/lib/device.js
+++ b/lib/device.js
@@ -323,7 +323,7 @@ device.prototype.drawPage = function () {
 		self.data = graphics.getImagesForPage(self.page)
 
 		for (var i in self.data) {
-			self.panel.draw(i, self.data[i].buffer)
+			self.panel.draw(Number(i), self.data[i].buffer)
 		}
 	} else {
 		/* TODO: We should move this to the device module */
--- a/lib/elgato_dm.js
+++ b/lib/elgato_dm.js
@@ -23,7 +23,7 @@ var usb = require('./usb')
 var elgatoEmulator = require('./elgato_emulator')
 var elgatoPluginDevice = require('./elgato_plugin')
 var satelliteDevice = require('./satellite_device')
-var preview = require('./preview')
+var satelliteDeviceV2 = require('./satellite/satellite_device_v2')
 var system
 
 HID.setDriverType('libusb')
@@ -284,6 +284,11 @@ elgatoDM.prototype.addDevice = function (device, type) {
 		self.updateDevicesList()
 	} else if (type === 'satellite_device') {
 		instances[device.path] = new satelliteDevice(system, device.path, device)
+		instances[device.path].deviceHandler = new deviceHandler(system, instances[device.path])
+
+		self.updateDevicesList()
+	} else if (type === 'satellite_device2') {
+		instances[device.path] = new satelliteDeviceV2(system, device)
 		instances[device.path].deviceHandler = new deviceHandler(system, instances[device.path])
 
 		self.updateDevicesList()
--- a/lib/elgato_dm.js
+++ b/lib/elgato_dm.js
@@ -172,44 +172,42 @@ elgatoDM.prototype.refreshDevices = function (cb) {
 	function scanUSB() {
 		debug('USB: checking devices (blocking call)')
 		var devices = HID.devices()
-		for (var i = 0; i < devices.length; ++i) {
-			;(function (device) {
-				if (
-					!ignoreStreamDeck &&
-					device.vendorId === 0x0fd9 &&
-					device.productId === 0x0060 &&
-					instances[device.path] === undefined
-				) {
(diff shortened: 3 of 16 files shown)


Julusian:
I have added some additional device logic fixes here, as while debugging an issue with colours being resent on changing the brightness, I discovered that the config page number handling logic was a bit broken and confusing.

Back in 1.2.0 it was made so that companion would remember the last page a streamdeck was on, and return it to that upon initialising the streamdeck. In ~1.3.0 it was added to set the page number of a surface in the config. These two features have never interacted well, in that if you set the page in the config, then that will override the last page behaviour with no way to change it back. As a side note, the page option was added to the regular streamdeck in 2.1.0, before then it would only be confusing for the mini.
The best fix I see here is to clarify this is the expected behaviour and make it possible to choose how it will behave. So the ui has an adjustment to the surface edit modal to allow for this.
image

Additionally, while digging around here I was getting really lost trying to follow the deviceconfig data flow, so I have straightened it out to be simpler. Some changes had to be made to every surface device type here, not all have been tested yet, but I believe it should be safe for them all.

Julusian:
I've made a change to combine the `KEY-DRAW` and `KEY-COLOR` messages into a single `KEY-STATE` message.
This is so that there is a single command for all state changes for a button, which may be useful in future extensions of the protocol

McHauge:
@Julusian Hi was thinking, would it be possible to also have a "text" only part for the key state, just like with the colour. as well as a true/false for if you want it as part of the updates

Julusian:
@McHauge yeah that would be really easy to add (just need to be careful to encode it safely). At the point where the KEY-STATE is being sent we have access to the full style object for the key. I have been expecting that to be wanted by someone :)

McHauge:
sounds, great!

also how long away would you think it is from joining the main branch?

PR #1675, opened 2021-07-25
feat: provide raster with feedback info

krocheck:
Adds `width` and `height` to the `info` parameter of feedbacks.  Also fires a re-check of all feedbacks after a change of `remove_topbar`.


lib/feedback.js
```diff
@@ -108,6 +108,18 @@ function feedback(system) {
 		self.system.emit('feedback_save')
 	})
 
+	system.on('feedback_check_all', function () {
```
Julusian: This isnt rechecking any feedbacks used in triggers, but the name suggests it should.
Perhaps call it `feedback_check_all_banks`?

Julusian (changes requested):
I think it would be better to have the width and height as part of the 3rd parameter to the callbacks (with the page and bank number)
https://github.com/bitfocus/companion/blob/feat/feedback-raster-size/lib/feedback.js#L225
https://github.com/bitfocus/companion/blob/master/instance_skel_types.d.ts#L77

lib/feedback.js
krocheck: There was actually a conflict with `feedback_check_all` because it already existed.  Fixed

lib/feedback.js
```diff
@@ -246,6 +270,16 @@ function feedback(system) {
 		}
 	})
 
+	system.on('feedback_check_all_banks', function () {
```
Julusian: I think you can use `feedback_check_all` instead of this. No second parameter necessary

Julusian (approved):

lib/feedback.js
krocheck: Thought of it, but there's no need for this case to cycle through any triggers.  Your call.

Julusian: Oh good point, I didnt realise that was subscribed to over there also..
Perhaps just pull out the function here to a shared implementation then

krocheck: Yeah, I can do that.  `scheduler` listens to `feedback_check_all`, which I'm not keen about but it works.  Yeah I can abstract the interior calls for this and `feedback_check_all` because of that though.

PR #1716, opened 2021-09-09
Fix: update feedback button text when variable value changes

### LINKED ISSUE #1715: [BUG] Feedback button text that contain variables do not get updated when variable value changes
**Describe the bug**
When adding a variable to the Button Text of a Feedback, the initial value is displayed but it's not updated over time if the value of the variable changes.

**To Reproduce**
While I'm sure this applies to many use cases, what I was trying to do was create a button to trigger the start of a cue on QLab. The initial text of the button was "play cue 1". While the cue was playing, (using a feedback) I wanted to change the color of the button and display the time remaining on the cue using one of the existing variables - `$(qlabfb:r_ss)` .

Steps to Reproduce:
1. Added/configured an instance of the **Figure 53 QLab with feedback** module (labeled the instance `qlabfb`)
2. Created a **Regular Button**
3. Set the **Button Text** to "Play Cue 1"
3. Added a **Start (cue)** action (configured with an existing cue, in my case cue "1", which has a duration of 15 seconds)
4. Added a **Indicate Cue is running** feedback
  a. Set the **Cue Number** to "1"
  b. Added **Text** to the **Change style properties**
  c. Set the **Button Text** property to `$(qlabfb:r_ss)` (variable the holds the remaining seconds)
  d. Set the desired **Color** and **Background** (irrelevant to this bug)
5. I pressed the button on the Stream Deck
6. Cue 1 started playing on QLab
7. The button color and background changed as expected
8. The button text changed to "15", but remained at "15" for the duration of the cue
9. Once the cue stopped playing the button style reverted back to the default (including the Button Text back to "Play Cue 1")

**Expected behavior**
I expected the button text to update along with the value of the variable.

**Screenshots**
image

- OS: Windows 10 21H1
- Browser: Chrome Version 92.0.4515.159 (Official Build) (64-bit)
- Companion Version: - 2.2.0-419c63bf-3541

**Additional context**
I have a proposed fix and will be submitting a pull request shortly.

estilles:
Resolves #1715

### DIFF
--- a/lib/feedback.js
+++ b/lib/feedback.js
@@ -595,6 +595,11 @@ feedback.prototype.setCachedStyle = function (page, bank, index, definition, fee
 
 	if ((definition && definition.type === 'boolean') || typeof style === 'boolean') {
 		if (style) {
+			if (feedback.style.text) {
+				self.system.emit('variable_parse', feedback.style.text, function (str) {
+					feedback.style.parsedText = str
+				})
+			}
 			style = { ...feedback.style }
 		} else {
 			style = {}


estilles:
Rebased to most recent commit and pushed. There were no updates to the code.

Julusian:
I don't entirely understand how this works, and it has a side effect of causing the ui to think there is a `parsedText` property defined on the feedback 
image

Julusian:
I have written a fix that will trigger updates in the correct flow. Please test and let me know of any issues https://github.com/bitfocus/companion/commit/130f2f495edce5881dfa0dfdbb59aefc9deeb1d8

estilles:
Thanks for that fix @Julusian. I really appreciate it.

PR #1736, opened 2021-09-27
Add url path to tabs to allow bookmarking

bryce-seifert:


Julusian:
It should be possible to do this directly with react-router instead of manually looking at `window.location`, which would be really good to do so that things like the back button work.

Julusian:
I've reworked this a bit and merged it in https://github.com/bitfocus/companion/commit/455de83f40a98fe8cd03fe801fe744bf43493224

PR #1746, opened 2021-10-04
update old livingasone module to resi

EmilyStacy:


Julusian:
Unfortunately it is unlikely that there will be another 2.1 release, but I have made this change in the 2.2 betas for you

EmilyStacy:
> Unfortunately it is unlikely that there will be another 2.1 release, but I have made this change in the 2.2 betas for you

Thanks a lot! @Julusian so we don't need to delete the legacy submodule right?

PR #1766, opened 2021-10-24
feat: ember+ service control

krocheck:
As a follow-up to #1644, the Ember+ service was not setup to be enabled via the settings page.  This adds that functionality.  Since the in place upgrade was designed to run once, this will disable the module for beta users since the merge on June 29, but will react appropriately for upgrades after a public release.


Julusian (approved):
LGTM

It looks to me like it will upgrade as you expect, but I could be wrong, Im not reading in detail

PR #1781, opened 2021-11-10
feat: streamdeck offset

Julusian:
Replaces #1643

New config fields have been added to the surfaces, to allow for them to be positioned in the grid.
image

The implementation is done generically. I have tested it with a 15 key, and it should work for a mini or a satellite device

### DIFF
--- a/lib/device.js
+++ b/lib/device.js
@@ -19,8 +19,6 @@ var system
 var debug = require('debug')('lib/device')
 var graphics
 
-var debug = require('debug')('lib/device')
-
 function device(_system, panel) {
 	var self = this
 
@@ -36,10 +34,10 @@ function device(_system, panel) {
 	self.page = 1 // The current page of the device
 	self.lastpress = ''
 	self.lastpage = 0
-	self.config = {}
 	self.userconfig = {}
 	self.deviceconfig = {}
 	self.panelconfig = {}
+	self.panelinfo = {}
 
 	system.emit('db_get', 'deviceconfig', function (res) {
 		if (res !== undefined && res !== null) {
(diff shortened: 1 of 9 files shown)


Julusian:
@istnv what do you think of this solution for #1643 ?

istnv:
> @istnv what do you think of this solution for #1643 ?

This looks great!

krocheck:
@Julusian does this detect the maximum number of offsets available, meaning an XL shows no available offsets while a Mini shows 2 vertical and 5(?) horizontal?  I couldn't clearly tell from the read or your description.

Without looking thoroughly at the implementation, I'm good with the approach.

Julusian:
> @Julusian does this detect the maximum number of offsets available, meaning an XL shows no available offsets while a Mini shows 2 vertical and 5(?) horizontal? I couldn't clearly tell from the read or your description.
> 
> Without looking thoroughly at the implementation, I'm good with the approach.

Yes, it uses the dimensions reported by the device to work out the range of values to allow in the ui. And if the allowed range is 0, then the option is not shown.

krocheck:
So I'm wondering if this would allow us to increase the total button x and y to the maximum of the XKeys to allow that implementation to address a single page instead of the way it spans multiple pages now?

Julusian:
@krocheck yeah it would help there, but we need to be careful as a bunch of the internal actions reference bank numbers explicitly. wouldnt surprise me if some modules are doing something weird with that information too
I am open to giving it a try early in 2.3, but the big question will be what should the new size be?

krocheck:
Yeah.  I looked at the XKeys Rack, which has 20 columns (the 128 has 16) and the max rows across the lineup is 8.

Although this is defined with globals, we technically could make it configurable with min/maxes and even put detection in so we still default to 32 but if something like a 128 gets plugged in to changes to 16x8 to accommodate.

krocheck:
Just thinking out loud, being able to bump the size like this would make it easier for people with multiple stream decks to be on the same page.  Like 2 XL's side by side meant to operate as one you can bump the canvas from 8x4 to 16x4 and offset the 2nd panel.

krocheck (approved):

Julusian:
I think that to be able to safely handle a flexible grid size we will need to change the button ids away from `1.12` to something that will be static no matter the size. Perhaps `1.2.3` or `1.2x3`. And for the db/internal data structures to reflect this change (probably just change an array to an object with keys of `2.3` or `2x3`.
Otherwise changing the size will require us to run an upgrade script on everything to make sure the values are correct, and to hope we dont miss anything etc. If we have a proper xy coordinate, then it removes that work and risk.
So I don't think it a quick or small thing to do, especially as I found some bugs in the 15 to 32 translation a few months back.

Also this will affect anyone doing anything over tcp/udp/osc as the button numbers to trigger anything there will change.
I do think that being able to change the grid size is a feature we need to do eventually, but I see it affecting a lot of places and affecting almost every user in some way so is not something to be rushed

krocheck:
Yeah, I agree on all points and many of those I was already starting to try and think through.  I would switch rows to being letters and the columns to numbers ... `1.1` = `1.A1`, `1.17` = `1.C1`.  I'd think we can leave the 1-32 addressing in external protocols for inbound commands or have an option for each service to enable 'legacy' addressing.  The Elgato plugin is my biggest concern.

Julusian:
yeah I like A1 as a naming scheme, should be nice and familiar for people too.
true, it should be possible to detect for most things based on whether they sent `13` or `B3`. or we could make it an option
I'm not worried about the plugin. Its implemented as a surface, so we should be able to easily limit it to the 32 without any changes to the plugin itself. All the surfaces will need to go through a generic translation layer to map the new ids to the existing sequential indices, and allow for them to be offset

krocheck:
LOL, yeah.  For the surfaces and offsets we can basically make some alphabet math utils and convert the default ids to a base XY and math it from there.

PR #1790, opened 2021-11-18
allow for other content type headers

josephdadams:
Doesn't assume the content type is going to be application/json unless a content-type header is not specified


josephdadams:
Actually after testing this is not needed. We can inject/override the content type using the existing structure

PR #1796, opened 2021-11-23
feat: add conditional config field display

estilles:
Conditionally display configuration fields based on the value(s) of other configuration fields.

A optional option `isVisible` can be assigned a function which receives the current `instanceConfig` to determine visibilty returning a Boolean. For backwards compatibility, if `isVisible` is not present the config field will always be displayed.

In the example below:
- the `version` field is always displayed
- the `community` field is only displayed when `version` is `v1` or `v2c` 
- the `engineID` field is only displayed when `version` is `v3`

```JavaScript
			{
				type: 'dropdown',
				id: 'version',
				label: 'SNMP Version',
				width: 6,
				choices: [
					{ id: 'v1', label: 'SNMP v1' },
					{ id: 'v2c', label: 'SNMP v2c' },
					{ id: 'v3', label: 'SNMP v3' },
				],
				default: 'v1',
				require: true,
			},
			{
				type: 'textinput',
				id: 'community',
				width: 6,
				label: 'Community',
				default: 'companion',
				isVisible: ({ version }) => version === 'v1' || version === 'v2c',
			},
			{
				type: 'textinput',
				id: 'engineID',
				width: 6,
				label: 'Engine ID',
				default: '',
				isVisible: ({ version }) => version === 'v3',
			},

```

Signed-off-by: Johnny Estilles <johnny@estilles.com>

### DIFF
--- a/instance_skel_types.d.ts
+++ b/instance_skel_types.d.ts
@@ -111,6 +111,7 @@ export interface CompanionInputField {
 	type: 'text' | 'textinput' | 'textwithvariables' | 'dropdown' | 'colorpicker' | 'number' | 'checkbox'
 	label: string
 	tooltip?: string
+	isVisible?: () => boolean
 }
 export interface CompanionInputFieldText extends CompanionInputField {
 	type: 'text'
--- a/instance_skel_types.d.ts
+++ b/instance_skel_types.d.ts
@@ -111,7 +111,7 @@ export interface CompanionInputField {
 	type: 'text' | 'textinput' | 'textwithvariables' | 'dropdown' | 'colorpicker' | 'number' | 'checkbox'
 	label: string
 	tooltip?: string
-	isVisible?: () => boolean
+	isVisible?: (config: { [id: string]: InputValue }) => boolean
 }
 export interface CompanionInputFieldText extends CompanionInputField {
 	type: 'text'
--- a/webui/src/util.jsx
+++ b/webui/src/util.jsx
@@ -29,6 +29,32 @@ export function socketEmit(socket, name, args, timeout, timeoutMessage) {
 	return pTimeout(p, timeout, timeoutMessage ?? `Timed out after ${timeout / 1000}s`)
 }
 
+export function sandbox(serializedFn) {
+	// proxy handlers for config proxy
+	const proxyHandlers = {
+		has: () => true,
+		get: (obj, prop) => (prop === Symbol.unscopables ? undefined : obj[prop]),
+	}
+
+	// wrap function in a with to limit the scope
+	const src = `
+		with (context) {
+			const fn = ${serializedFn}
+			return fn(config)
+		}
+	`
+
+	// eslint-disable-next-line no-new-func
+	const scopedFn = new Function('context', src)
+
+	return function (config) {
+		// create a sandboxed/proxy version of the context passed to the function
+		const configProxy = new Proxy(config, proxyHandlers)
+		// call scoped function with context that only includes config
+		return scopedFn({ config: configProxy })
+	}
+}
+
 function ErrorFallback({ error, resetErrorBoundary }) {
 	return (
 		<CAlert color="danger">
--- a/webui/src/util.jsx
+++ b/webui/src/util.jsx
@@ -49,9 +49,9 @@ export function sandbox(serializedFn) {
 
 	return function (config) {
 		// create a sandboxed/proxy version of the context passed to the function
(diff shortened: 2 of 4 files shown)


krocheck:
I definitely see the use-case here, but this is going to flag for the fact that you're putting UI logic into the core where-as its best to get a structure that can be handed off to the UI and let it do the logic.  The long-term goal of threading out instances could cause us to break this in the future as written with no easy/painless alternative methodology.

The best I can think of, which would be limiting, is an array and/or multi-dimensional array structure for the `isVisible` field that the UI can parse and loop through to do compares.  It not elegant and limits the feature from doing any complex logic, but I don't know how often/deep that would be needed anyway.  It could be possible for the `isVisible` field to either be a function to generate the array or a fixed array.  The system could check before the definition data gets passed to the UI and either execute the function and transmit the array that's returned or just send the fixed array.

I know developers have asked for this in the past and we've never thought through too much how to achieve it, so I'm really happy to see this functioning.  Other guys may have fancier ways in mind to keep the callback-style logic and send it to the UI for direct processing, but that's not my forte, so I'll leave that up to them.

estilles:
@krocheck ... all your points are well taken. Let me refactor this so I can address your comment regarding "putting UI logic into the core".

instance_skel_types.d.ts
```diff
@@ -111,6 +111,7 @@ export interface CompanionInputField {
 	type: 'text' | 'textinput' | 'textwithvariables' | 'dropdown' | 'colorpicker' | 'number' | 'checkbox'
 	label: string
 	tooltip?: string
+	isVisible?: () => boolean
```
Julusian: This should be `isVisible?: (config: { [id: string]: InputValue }) => boolean`

webui/src/util.jsx
```diff
@@ -29,6 +29,32 @@ export function socketEmit(socket, name, args, timeout, timeoutMessage) {
 	return pTimeout(p, timeout, timeoutMessage ?? `Timed out after ${timeout / 1000}s`)
 }
 
+export function sandbox(serializedFn) {
```
Julusian, line 56: I was curious about this, as it looks great but feels too simple.

Came across [this article](https://portswigger.net/research/attacking-and-defending-javascript-sandboxes) which has some criticisms of the original implemenation and suggests some improvements. It would be good to pull those in

This feels like something that there should be an npm packge for, but a quick search didnt come up with one

Julusian (commented):
Overall I am happy with the approach now, and it is a lot simpler than I was first expecting this would end up being.

webui/src/util.jsx
krocheck: @Julusian if one doesn't exist should the project deploy one as such?

krocheck:
> Overall I am happy with the approach now, and it is a lot simpler than I was first expecting this would end up being.

I agree, pending @Julusian notes

instance_skel_types.d.ts
estilles: You're right. I added it as a placeholder and forgot to go back and update. Will fix.

webui/src/util.jsx
estilles: > Came across this article which has some criticisms of the original implemenation and suggests some improvements. It would be good to pull those in

I'll check the article to see if there's anything I can improve.

> @Julusian if one doesn't exist should the project deploy one as such?

Yes I intended to eventually bundle it up into it's own npm package.

estilles: That article is GREAT! I need to incorporate some of those features into our sandbox function.

Julusian: > > @Julusian if one doesn't exist should the project deploy one as such?
> 
> Yes I intended to eventually bundle it up into it's own npm package.

I'm not sure if it is worth it, as it means that making any changes will be harder. Unless @estilles wants to reuse it in other things, then it would be worth it

estilles: I will probably create a more general use version of this in the figure and turn it into an npm module, but this one is a very customized for our use case.

estilles: @Julusian ... based on the article you pointed out (thanks by the way) I've added the following:

- allow specific global objects
- freeze allowed global objects to prevent `indexOf` attack
- use a double proxy approach
- add `"use strict"` to prevent `window` object leaking
- protect against "aysnc functions/generators" bypassing sandbox

Julusian (approved):

krocheck (approved):

Issue #1775, opened 2021-11-02
[BUG] The letters ÆØ are not displayed when the font size is 7pt


kjetilpp:
**Describe the bug**
The letters ÆØ are not displayed on the webpage or the stream deck when the font size is 7pt.

**To Reproduce**

Add a button with text including Ø or Æ, set size to 7pt. 

**Expected behavior**
Expected the letters to show.

**Screenshots**
![Capture](https://user-images.githubusercontent.com/5620578/139848250-e5b2e39e-16d1-4b7f-af16-827331d3f0b8.PNG)
**Desktop (plese complete the following information):**

- OS: [RaspberryPI]
- Browser [chrome]
- Companion Version 2.2.0 (2.2.0-1e65f702-3485)

**Additional context**
Have changed locale of the Pi to nb_NO.UTF-8

Julusian:
Related: #1316

KzBoy:
I believe this is finished if someone can merge my PR. #1799 

If there is an issue just let me know.

Issue #1316, opened 2020-12-14
[BUG] Swedish text characters not showing up with 7pt font size


Gartom:
**Describe the bug**
Swedish characters will not show up on buttons when using 7pt font size. All other font sizes work.

**To Reproduce**
1. Launch Companion GUI
2. Select Buttons tab
3. Create a regular button
4. Write Swedish characters "åäö ÅÄÖ" in the "Text" field 
5. Note that text is showing with 24pt sized font
6. Change font size to 7pt
7. Note that the Swedish characters disappeared from button (this also happens with Swedish characters in dynamic variables from module instances)

**Expected behavior**
Correct text printout on button.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Firefox 83.0
 - Companion Version 2.1.2

**Additional context**

dnmeid:
It's not a bug it's a feature. The 7pt font is hand-designed and very tiny. It is quite challenging to squeeze the ASCII letters into that tiny area. So all the non ASCII chars are working only in the larger fonts.

Julusian:
@dnmeid should this really be 'wontfix'?
I would expect the characters to show up in their ascii form instead. so that "åäö ÅÄÖ" will get shown as "aa0 AAO". Dropping characters completely is a bit weird.

Gartom:
I looked into the font directory and saw that 7px fonts seemed to be hand made for this purpose. I created similar fonts for å, ä, ö, Å, Ä, and Ö, using their corresponding ASCII numbers as file names. Could these be integrated?

![132](https://user-images.githubusercontent.com/4034632/103247268-9c238f00-4966-11eb-9d2a-566818bb0b9d.png)
![134](https://user-images.githubusercontent.com/4034632/103247270-9cbc2580-4966-11eb-9d3f-6a74fa5170bb.png)
![142](https://user-images.githubusercontent.com/4034632/103247271-9cbc2580-4966-11eb-827f-dbee29d28c9c.png)
![143](https://user-images.githubusercontent.com/4034632/103247272-9d54bc00-4966-11eb-92d7-0eea475c6f14.png)
![148](https://user-images.githubusercontent.com/4034632/103247275-9d54bc00-4966-11eb-823b-82e7b0acaf61.png)
![153](https://user-images.githubusercontent.com/4034632/103247276-9d54bc00-4966-11eb-8729-e8e324aeddec.png)





, see attached files.

Gartom:
[...]

PR #1799, opened 2021-11-25
feat: Add 7px character fonts 198(Æ) & 216(Ø)

KzBoy:
Added 7px character fonts 198(Æ) & 216(Ø)

Fixes issue #1775  (assuming I did this right, first time commiting something to Github)


krocheck:
@Julusian any issue you can see?  I'm not sure why the 7pt font excludes a number of characters.

KzBoy:
According to [this](https://github.com/bitfocus/companion/issues/1316#issuecomment-751836957) it's a feature. *shrugs*

Julusian:
@krocheck as far as Im aware its that they have to be designed by hand and it would take a lot of time to get a full extended alphabet

PR #1802, opened 2021-11-27
Add `isVisibile` to action/feedback options

estilles:
This is the next phase of my "conditional visibility" project. This PR includes:

- A slight refactoring of the `isVisible` implementation for config fields.
  In the original implementation the `isVisible` property was overwritten with a serialized copy of the function. I wasn't happy with that, so instead I now add and `isVisibleFn` property which is the serialized copy of `isVisible`.
- Add `isVisible` to action options
- Add `isVisible` to feedback options

Because instance configs are slightly different from actions/feedbacks under the hood, this implementation is slightly different than the visibility implementation for config fields, but it is functionally equivalent for all intents and purposes.

#### Sample Usage
The `isVisible` function receives a copy of the action state. The function then returns a boolean indicating the desired visibility.

```JavaScript
action.testAction = {
  label: 'Some Action',
  options: [
    {
      type: 'dropdown',
      label: 'Option 1',
      id: 'opt1',
      choices: [
        { id: 'yes', label: 'Yes' },
        { id: 'no', label: 'No' },
      ],
      default: 'yes',
    },
    {
      type: 'textinput',
      label: 'Option 2',
      id: 'opt2',
      default: '',
      isVisible: (action) => action.options.opt1 === 'yes',
    },
  ],
}

### DIFF
--- a/instance_skel.js
+++ b/instance_skel.js
@@ -177,6 +177,26 @@ instance.prototype.addUpgradeScript = function () {
 	)
 }
 
+instance.prototype.serializeIsVisibleFn = function (options = []) {
+	return options.map((option) => {
+		if ('isVisible' in option) {
+			if (typeof option.isVisible === 'function') {
+				return {
+					...option,
+					isVisibleFn: option.isVisible.toString(),
+					isVisible: undefined,
+				}
+			}
+		}
+
+		// ignore any existing `isVisibleFn` to avoid code injection
+		return {
+			...option,
+			isVisibleFn: undefined,
+		}
+	})
+}
+
 instance.prototype.setActions = function (actions) {
 	var self = this
 
@@ -185,19 +205,7 @@ instance.prototype.setActions = function (actions) {
 	} else {
 		actions = Object.fromEntries(
 			Object.entries(actions).map(([id, action]) => {
-				action.options = action.options?.map((option) => {
-					if ('isVisible' in option) {
-						if (typeof option.isVisible === 'function') {
-							return {
-								...option,
-								isVisibleFn: option.isVisible.toString(),
-							}
-						}
-					}
-					// ignore any existing `isVisibleFn` to avoid code injection
-					delete option.isVisibleFn
-					return option
-				})
+				action.options = self.serializeIsVisibleFn(action.options)
 				return [id, action]
 			})
 		)
@@ -248,19 +256,7 @@ instance.prototype.setFeedbackDefinitions = function (feedbacks) {
 	} else {
 		feedbacks = Object.fromEntries(
 			Object.entries(feedbacks).map(([id, feedback]) => {
-				feedback.options = feedback.options?.map((option) => {
-					if ('isVisible' in option) {
-						if (typeof option.isVisible === 'function') {
-							return {
-								...option,
-								isVisibleFn: option.isVisible.toString(),
-							}
-						}
-					}
-					// ignore any existing `isVisibleFn` to avoid code injection
-					delete option.isVisibleFn
-					return option
-				})
+				feedback.options = self.serializeIsVisibleFn(feedback.options)
 				return [id, feedback]
 			})
 		)
--- a/instance_skel.js
+++ b/instance_skel.js
@@ -180,28 +180,28 @@ instance.prototype.addUpgradeScript = function () {
 instance.prototype.setActions = function (actions) {
 	var self = this
 
-	actions = Object.fromEntries(
-		Object.entries(actions).map(([id, action]) => {
-			action.options = action.options?.map((option) => {
-				if ('isVisible' in option) {
-					if (typeof option.isVisible === 'function') {
-						return {
-							...option,
-							isVisibleFn: option.isVisible.toString(),
-						}
-					}
-				}
-				// ignore any existing `isVisibleFn` to avoid code injection
-				delete option.isVisibleFn
-				return option
-			})
-			return [id, action]
-		})
-	)
-
 	if (actions === undefined) {
 		self._actionDefinitions = {}
 	} else {
+		actions = Object.fromEntries(
+			Object.entries(actions).map(([id, action]) => {
+				action.options = action.options?.map((option) => {
+					if ('isVisible' in option) {
+						if (typeof option.isVisible === 'function') {
+							return {
+								...option,
+								isVisibleFn: option.isVisible.toString(),
+							}
+						}
+					}
+					// ignore any existing `isVisibleFn` to avoid code injection
+					delete option.isVisibleFn
+					return option
+				})
+				return [id, action]
+			})
+		)
+
 		self._actionDefinitions = actions
 	}
 
--- a/instance_skel.js
+++ b/instance_skel.js
@@ -19,6 +19,7 @@ var util = require('util')
 var debug = require('debug')('lib/instance_skel')
 var image = require('./lib/image')
 var icons = require('./lib/resources/icons')
+var { serializeIsVisibleFn } = require('./lib/resources/util')
 
 function instance(system, id, config) {
 	var self = this
@@ -177,27 +178,6 @@ instance.prototype.addUpgradeScript = function () {
 	)
 }
 
-// NOTE: *** This is a internal method. DO NOT call or override. ***
-instance.prototype.serializeIsVisibleFn = function (options = []) {
-	return options.map((option) => {
-		if ('isVisible' in option) {
-			if (typeof option.isVisible === 'function') {
-				return {
-					...option,
-					isVisibleFn: option.isVisible.toString(),
-					isVisible: undefined,
-				}
-			}
-		}
-
-		// ignore any existing `isVisibleFn` to avoid code injection
-		return {
-			...option,
-			isVisibleFn: undefined,
-		}
-	})
-}
-
 instance.prototype.setActions = function (actions) {
 	var self = this
 
@@ -206,7 +186,7 @@ instance.prototype.setActions = function (actions) {
 	} else {
 		actions = Object.fromEntries(
 			Object.entries(actions).map(([id, action]) => {
-				action.options = self.serializeIsVisibleFn(action.options)
+				action.options = serializeIsVisibleFn(action.options)
 				return [id, action]
 			})
 		)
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by krocheck** [comment]
I'd like to see you go ahead and get this everywhere relevant in one go rather then playing whack-a-mole.

**Comment by estilles** [comment]
> I'd like to see you go ahead and get this everywhere relevant in one go rather then playing whack-a-mole.

Got it. Will add to feedbacks as well.

**Comment by krocheck** [comment]
Don't forget about the scheduler interfaces

**Comment by estilles** [comment]
@krocheck 
> Don't forget about the scheduler interfaces

Since scheduler `EditModal` uses the `FeebackEditor` and `ActionsPanelInner` (which uses `ActionTableRow`) components to handle the display/editing of feedbacks and actions respectively, it automatically inherits the "option visibility" feature added to those components.

File: instance_skel.js [hunk 1]
```diff
@@ -227,6 +246,25 @@ instance.prototype.setFeedbackDefinitions = function (feedbacks) {
...
 		self._feedbackDefinitions = {}
 	} else {
+		feedbacks = Object.fromEntries(
+			Object.entries(feedbacks).map(([id, feedback]) => {
+				feedback.options = feedback.options?.map((option) => {
+					if ('isVisible' in option) {
+						if (typeof option.isVisible === 'function') {
+							return {
+								...option,
+								isVisibleFn: option.isVisible.toString(),
```
Julusian: At the various points we do this, we should set `isVisible: undefined` to make sure socket.io doesnt also try to serialize function

instance_skel_types.d.ts
```diff
@@ -197,6 +197,7 @@ export interface CompanionFeedbackBase<TRes> {
 	) => TRes
 	subscribe?: (feedback: CompanionFeedbackEvent) => void
 	unsubscribe?: (feedback: CompanionFeedbackEvent) => void
+	isVisible?: (config: { [id: string]: InputValue }) => boolean
```
Julusian: This signature looks wrong, and is missing one being added to the action type too

webui/src/Buttons/EditButton/ActionsPanel.jsx
```diff
@@ -291,6 +293,39 @@ function ActionTableRow({ action, index, dragId, setValue, doDelete, doDelay, mo
 	})
 	preview(drop(ref))
 
+	useEffect(() => {
+		const actionSpec = actionsContext[action.label]
+		const options = actionSpec?.options ?? []
+
+		for (const option of options) {
+			if (typeof option.isVisibleFn === 'string') {
```
Julusian: Im not sure if this is the right place to do this, as we will be doing it many times over the lifetime of a browser tab. 
If we simply add a ` && typeof option.isVisible !== 'function'`, then that should avoid duplicating work unnecessarily.

webui/src/Buttons/EditButton/ActionsPanel.jsx
```diff
@@ -291,6 +293,39 @@ function ActionTableRow({ action, index, dragId, setValue, doDelete, doDelay, mo
 	})
 	preview(drop(ref))
 
+	useEffect(() => {
+		const actionSpec = actionsContext[action.label]
```
Julusian: I think it would be better to move `actionSpec` outside of this useEffect. As it stands, any time any module updates its list of actions, or the current action gets edited the sandbox will be reapplied for every action component that is mounted. By simply moving that variable outside, it will only invalidate the ones where actionSpec was changed.

webui/src/Buttons/EditButton/ActionsPanel.jsx
```diff
@@ -291,6 +293,39 @@ function ActionTableRow({ action, index, dragId, setValue, doDelete, doDelay, mo
...
+		for (const option of options) {
+			if (typeof option.isVisibleFn === 'string') {
+				option.isVisible = sandbox(option.isVisibleFn)
+			}
+		}
+	}, [actionsContext, action])
+
+	useEffect(() => {
+		const visibility = {}
+		const actionSpec = actionsContext[action.label]
```
Julusian: again, I think actionSpec should be defined outside the useEffect.

webui/src/Buttons/EditButton/FeedbackPanel.jsx
```diff
@@ -255,6 +255,41 @@ export function FeedbackEditor({ feedback, setValue, innerDelete, setSelectedSty
...
+		const options = feedbackSpec?.options ?? []
+
+		console.log({ options })
+
+		for (const option of options) {
+			if (typeof option.isVisibleFn === 'string') {
+				option.isVisible = sandbox(option.isVisibleFn)
+			}
+		}
+	}, [feedbackSpec, feedback])
```
Julusian: I don't think this `feedback` reference is needed. It should be removed to avoid invalidations of the effect

webui/src/Buttons/EditButton/FeedbackPanel.jsx
```diff
@@ -255,6 +255,41 @@ export function FeedbackEditor({ feedback, setValue, innerDelete, setSelectedSty
 	const feedbackSpec = (feedbacksContext[feedback.instance_id] || {})[feedback.type]
 	const options = feedbackSpec?.options ?? []
 
+	const [optionVisibility, setOptionVisibility] = useState({})
+
+	useEffect(() => {
+		const options = feedbackSpec?.options ?? []
+
+		console.log({ options })
```
Julusian: Remove this log please :)

Julusian (changes requested):
Im wondering if we can deduplicate some of this code. The options object of actions and feedbacks aren't using a common type in the typescript, but they are the same, so perhaps there could be a `serializeIsVisibleFn(options)` function? Maybe the same can be used for the config version too?

I have some comments on the actions, which may also apply to the feedbacks/other versions

estilles:
> Im wondering if we can deduplicate some of this code. The options object of actions and feedbacks aren't using a common type in the typescript, but they are the same, so perhaps there could be a `serializeIsVisibleFn(options)` function? Maybe the same can be used for the config version too?
> 
> I have some comments on the actions, which may also apply to the feedbacks/other versions

Thanks for the comments @Julusian. I'll work on it this weekend. :-)

instance_skel.js
estilles: I didn't think it would be necessary because
- the reason we're "manually" serializing the function is because JSON doesn't support them and Socket.io doesn't do it on our behalf
- on the WebUI we only use `isVisibleFn` and ignore `isVisible`

That said, sure. I'll add. Better to be safe.

instance_skel_types.d.ts
estilles: Re: Signature - fixing ...
Re: Missing in action? Actually this like should be here at all (I forgot to remove it). `isVisible` is an option property, which is why I added the property to `CompanionInputField`. This should cover options in actions, feedbacks and configs.

webui/src/Buttons/EditButton/ActionsPanel.jsx
estilles: Got it. Fixing ...

instance_skel.js
```diff
@@ -177,12 +177,39 @@ instance.prototype.addUpgradeScript = function () {
 	)
 }
 
+instance.prototype.serializeIsVisibleFn = function (options = []) {
```
Julusian: I would prefer this method to not be on the class, especially as it doesnt need anything from the class.
As revealed in a slack thread, some module devs look at this file to see what methods are available, and even though this is nothing they will ever need to call, someone will ask about it.

Also could this be dangerous if a module dev decides to override it? Probably not, no worse than the lack of data validation elsewhere

If you want to keep it here (I know @krocheck prefers having methods on classes rather than loose functions), then it should at least have a comment calling it an internal method

lib/instance.js
```diff
@@ -703,11 +703,12 @@ instance.prototype.connect = function (client) {
 				if (typeof field.isVisible === 'function') {
```
Julusian: could this use the implementation of `instance.prototype.serializeIsVisibleFn`? That looks accessible from here

webui/src/Buttons/EditButton/FeedbackPanel.jsx
```diff
@@ -283,6 +316,7 @@ export function FeedbackEditor({ feedback, setValue, innerDelete, setSelectedSty
 								actionId={feedback.id}
 								value={(feedback.options || {})[opt.id]}
 								setValue={setValue}
+								visibility={optionVisibility[opt.id] === false ? 'none' : null}
```
Julusian: I think it would be better for the visibility prop to be a boolean rather than a css value.
If you want to leave it like this that is fine too, Im thinking for future proofing for when we want to change the rendering and how we make them hidden changes

Julusian (approved):
I have added some thoughts, but nothing major

webui/src/Buttons/EditButton/FeedbackPanel.jsx
estilles: That makes sense.

lib/instance.js
estilles: Oh yeah! Forgot to use it here. That's the reason I added it to the class as opposed to making it a loose function.

instance_skel.js
estilles: Yeah ... I could've gone either way (on the class/loose function). I just wanted to be able to use it from `lib/instance'.

estilles: I just realized I forgot to use it there. (rolling eyes to myself)

estilles: I'll add a comment on the method

krocheck: I'd be OK with this being a static function on instance: `instance.SerializeIsVisibleFn = function (options = []) {`.  Could also put it as a static on `lib/instance.js` or make a `lib/resources/util.js` and have it wild there.  But yeah, I agree that this doesn't make sense as a class function, so any of those alternates are cool with me.

estilles: Static it is! :-)

estilles: Actually ... I think it would be better in `lib/resources/util.js`

estilles:
> I have added some thoughts, but nothing major

Thanks. Took your thoughts into consideration and submitted some last minute changes.

estilles:
@Julusian / @krocheck 
Can I get one final sanity check? I'm finally happy with this implementation. Let me know what you all think.

Julusian (approved):

Issue #1795, opened 2021-11-21
[BUG] Companion Buttons appearing and activating outside of their set folder, despite another folder or page being in focus.


CravenInsomniac:
**Describe the bug**
Companion has stopped respecting Stream Deck buttons when companion buttons are placed in folders. To further clarify, Companion actions are being triggered irregardless of the buttons being pressed.

**To Reproduce**
1. Make a folder on Stream Deck software.
2. Place a companion button.
3. Exit folder,
4. Press the button that matches the placement of the companion button. It can be an empty button, a folder, or another type of button. The button's LCD will either briefly flash with the companion buttons graphic if the buttons is occupied, or it will be fully replaced by companion graphic if empty; and the action will be activated.

**Expected behaviour**
When companion buttons are placed in folders, you should need to navigate to said folder in order to not activate companion action by accident.

**Screenshots**
[Link](https://imgur.com/a/kdTkWIT) to Imgur post and accompanying screenshots.

**Desktop (please complete the following information):**

- OS: Windows 10 Pro 21H1 Build:19043.1288
- Companion Version: 2.1.3 (2.1.3-6b6820cd-2696)
- Elgato Stream Deck Version: 5.1.3.14750
- Stream Deck Firmware Version: 1.00.012

Julusian:
This sounds like companion and the streamdeck application are both assuming they have exclusive control of the device.
If you simply unplug and replug the streamdeck then companion should lose it and then the streamdeck software will have sole control.

Essentially you need to be careful of order things start up in. Companion will claim every streamdeck it can find when it starts or when you click scan in the surfaces tab. The elgato software will claim every streamdeck at startup and watches for new ones.

CravenInsomniac:
So, currently I have the Elgato Stream Deck and Companion software start via Window's Task Scheduler. Am I understanding you correctly when I say I should delay Elgato's software so it starts after Companion? Or should I do it in the reverse?

CravenInsomniac:
After tinkering it seems that having Companion start before Elgato's Stream Deck software is indeed the way to go, leaving this open as the bug still occurs, I've just got a workaround. If anyone else happens upon this issue, I do have a recommendation. Make a button on the stream deck to restart the software itself.

To do this, make an "🚀 Open" button (found in System drop down). Have that button open a batch script that contains
[...]

Issue #1771, opened 2021-10-29
Disable access to stream deck


Grappa2020:
I like to use stream deck connected directly to application and run companion for other function without affecting stream deck.
so it will be nice to have the ability to disconnect from physical stream deck same as it implemented now for xkeys on version 2.

Julusian:
I have been wondering if there should be an option to choose between companion attempting to use streamdecks, and enabling support for the elgato plugin

Julusian:
This is implemented in the latest betas. There is now a setting to switch between having the elgato plugin api enabled, or to scan for local streamdecks

PR #1808, opened 2021-12-02
feat: config field to enable the elgato plugin instead of scanning for local devices

Julusian:
This should resolve some common issues from users who try the elgato plugin instead of native streamdeck support. eg #1795

The plugin api is now disabled by default. It has a userconfig field to enable it, and doing so will stop companion attempting to connect to streamdecks by itself.
So users will have to explicitly choose which way they want it to work.

There are still some edge cases where the two can fight, but this should solve the bulk of the cases and will resolve some of the issues observed at startup once it has already been setup.

I don't know how we can do this without blanket disabling the plugin for everyone, hopefully we are ok with that being done.

This can also be used to disable streamdeck loading #1771

### DIFF
--- a/app.js
+++ b/app.js
@@ -185,7 +185,13 @@ system.ready = function (logToFile) {
 	var preset = require('./lib/preset')(system)
 	var satelliteLegacy = require('./lib/satellite/satellite_server_legacy')(system)
 	var satellite = require('./lib/satellite/satellite_server')(system)
-	var ws_api = require('./lib/ws_api')(system)
+	var elgato_plugin_server
+	system.emit('get_userconfig_key', 'elgato_plugin_enable', function(enabled) {
+		if (enabled) {
+			// Only enable if box was ticked at startup
+			elgato_plugin_server = require('./lib/elgato_plugin_server')(system)
+		}
+	})
 	var help = require('./lib/help')(system)
 	var metrics = require('./lib/metrics')(system)
 
--- a/app.js
+++ b/app.js
@@ -185,13 +185,7 @@ system.ready = function (logToFile) {
 	var preset = require('./lib/preset')(system)
 	var satelliteLegacy = require('./lib/satellite/satellite_server_legacy')(system)
 	var satellite = require('./lib/satellite/satellite_server')(system)
-	var elgato_plugin_server
-	system.emit('get_userconfig_key', 'elgato_plugin_enable', function(enabled) {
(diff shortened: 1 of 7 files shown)


krocheck:
Yeah, something like this has been needed.  I'm cautious to 'break' existing setups by requiring the WS api (I'm fine with the refactor) to be enabled.  I'm wondering if a dropdown of `Auto Detect`, `Enabled`, and `Disabled` would be better where new installs can be `Disabled` but upgrade would take `Auto Detect`.

Or as-is we attempt a one-time detection on upgrade to make a best effort to set this correctly for the user?

Julusian:
I can understand your concern, I did think that at first.

I'm not convinced that auto-detect will be possible to make in a reliable fashion. Seeing as both bits of software fighting over streamdecks is a fairly regular issue that users post about, this suggests to me that we wont be able to rely on checking for the elgato software at companion startup, as we are already doing that. So unless we want to check every few minutes and disconnect from them all once we detect the software as running, I don't see auto-detect as being very reliable. And if we try to watch for the elgato software appearing, I expect it will lead to odd behaviour when we decide to disconnect from a streamdeck (eg Do we reset it to logo, or assume the elgato software was the last to draw?)
While this is an annoyance for all the plugin users, its a simple fix that just needs to be communicated to them well.

As for a one-time upgrade, we could look at the db and see if we have an entry for the plugin (indicating it has been used at some point before), but again that will often be wrong and makes the manual upgrade steps be more complex. This way it is a very simple note that every user needs to review of "If using the elgato plugin you need to enable it in the settings" rather than "If the elgato plugin stops working, make sure it is enabled in the settings. If direct streamdeck control stops working, make sure the plugin is disabled in the settings"

krocheck:
@Julusian I pushed a commit that dynamically starts and stops the web server as the checkbox is ticked.  I didn't remove the "(Requires Companion restart)" since that's sort of true although they could just cycle the stream deck(s) on enable or once disabled do a rescan USB.

krocheck (approved):

krocheck:
@Julusian your call when to hit 'Squash and merge` on this one.  I'm good, don't want to jump since I pushed some of my own code in.

Julusian:
@isaiahfranco yeah, it got broken with this. Fixed in the latest build

isaiahfranco:
Great! I'll have to give it a shot this week.

PR #1814, opened 2021-12-06
Adding code signing

JeffreyDavidsz:


Julusian:
@JeffreyDavidsz cool, I was thinking of looking into this later today
Is there a reason for using the separate action and not letting electron-builder do the signing while building?

JeffreyDavidsz:
Nope I wanted to see how this works (did not use GitHub actions before) that's all. But unfortunately this option is a no go (works on docker only) so I revert to electron signing or an other action ;-)

PR #1829, opened 2021-12-22
Added info about key commands to buttons page

JTF4:
I've been asked so many times about a faster way to copy, paste, and move buttons that I wanted to add a little bit to the UI to tell people about it.

I just added `You can also use common key commands such as copy, paste, and cut to move buttons around. You can even press the delete or backspace key with any button highlighted to delete it.` to the current `<CAlert>` on the buttons page.


josephdadams:
Good job

PR #1833, opened 2021-12-27
add makepro-x glue module v1.0.0

hrueger:
Hi maintainers,
this is the first version of the MakeProX Glue module.
If there's anything else that needs to be done before this module can be added?


estilles:
@hrueger You should request adding your module in the Slack module-development channel

PR #1842, opened 2022-01-06
feat: https

krocheck:
Adds the ability to run an HTTPS server for the Companion interface.

**Architecture**

In order to maintain a robust platform, the HTTPS server can be enabled (similarly to TCP, UDP remote control) but never replaces or disables the standard HTTP interface.  To complete this, `Express` and `io` needed to be abstracted in order for the additional service to hook them just the same.  All of the service configuration is done in `UserConfig`.  This includes the self signed certificate generator because the service ultimately just reads the information from `userconfig` and simplified the implementation.  In the end it still feels like a kludge so aspects of this may want to become their own thing and store as such in the `db`.

**Self Signed Certificate**

This solution includes a self-signed certificate generator.  The user is able to define the `Common Name` and the certificate's expiration period.  Once generated that certificate can be renewed against the original parameters or deleted, allowing a new certificate to be generated.

**External Certificate**

Also available is an option to use certificate files from the file system.  The required files are the `Private Key` and `Certificate`.  A `Chain` file can also be included.

**Error Handling**

Robust error handling was put in, and so there shouldn't be a case of a crash from this service and should provide very good error reporting to the user to indicate any failures of the service.

**Class Design**

Most of this uses ES6 classes and was designed to slot into future versions with ease.

**TODO**

- [ ] Cleanup design/css (@Julusian can you give me a hand here please)
- [x] Add user documentation


haakonnessjoen:
I have not tested it, but it looks good to me while looking it over. But I would have loved variables to be declared as either const or let, instead of var's. :D

krocheck:
I just cleaned that up for the new classes ;)

Julusian:
@krocheck I have a bug
If you start companion with https disabled, then enable it fails bind to the right port: `lib/server_https Listening at https://127.0.0.1:undefined +0ms`

Julusian (approved):
I'm not too keen on this

krocheck:
@Julusian I can't reproduce that port bind issue on Win.  The best I can think to do is put a `setTimeout` on the call to `this.init()` for a few seconds.

krocheck:
baah ... i'm an idiot,  I re-read, re-produced, and theoretically fixed.

Issue #481, opened 2019-02-05
Auto Rescan for Panels


hadphild:
**Describe the feature**
Auto Rescan for panels

**Is this platform dependent (windows, mac, ..)?**
NA

**If documentation is required to implement, do you know where to find it?**
NA

**Usecases**
If a USB Hub PSU looses power and you need to rescan when power comes back up. Can you Rescan every 1 min so these re-appear again

haakonnessjoen:
It should be configurable in that case, since people use the scan function of companion to control whether you connect it to companion or the stream deck app

hadphild:
We are using Companion to do a Basic install. I am running a 50m USB 2. If power is lost at the other end really want a it to boot back up and add its self back to the system.

depili:
Would also be really nice for headless where currently you have no indication for lost stream deck and no way to re-add it except restarting.

haakonnessjoen:
I will look at this soon. As a configurable option.

But @depili, even if it is headless, you can still control it via the web interface, so there is more ways to re-add it than restarting :)

PerRoine:
Is the scan still a blokking action Håkon ? I know it was but has that been resolved ?

depili:
Using the web interface would require hooking another computer to the same
private network in addition to the clocks and raspberry pi control panels,
which is way less than ideal.

alatteri:
I +1 this issue.  Needs to be either a continuous auto-scan if communication to a panel is lost, or some configurable interval.

DeeJayMX:
If you can expose a full simple url to trig a rescan, it could be scriptable by Windows itself with the device disconnect / reconnect event.

I saw that it's a web socket command that trig the rescan, I don't know how to connect and send webmessage by my automation system but I can trig simple URL.

My only way to do it right now is to close the app and restart it on disconnection / connection...

Not really bad but not really good...

llamafilm:
Another use case for this:  I have Companion running on multiple computers which are all on a networked KVM.  So each time I switch the KVM to a different host, I have to re-scan USB.  It would be great to recognize a new device automatically.

Julusian:
This will be in 3.0

john-bednar:
@hadphild @dszoka not sure what are the planned changes in 3.0 or if that will help, but we have created auto surface rescan for companion-pi - https://github.com/bitfocus/companion-pi/pull/4

PR #1844, opened 2022-01-06
feat: add rescan control to HTTP, TCP, UDP, OSC interfaces

llamafilm:
This should make it easier to integrate with other systems, related to #1274 and #481.

### DIFF
--- a/docs/5_remote_control/1_tcp_udp.md
+++ b/docs/5_remote_control/1_tcp_udp.md
@@ -20,6 +20,9 @@ Remote triggering can be done by sending TCP (port `51234`) or UDP (port `51235`
   _Change text color on a button (#000000)_
 - `STYLE BANK <page> <bank> BGCOLOR <color HEX>`  
   _Change background color on a button (#000000)_
+- `RESCAN`
+  _Make Companion rescan for newly attached USB surfaces_
+
 
 **Examples**  
 Set the emulator surface to page 23:  
--- a/docs/5_remote_control/2_http_remote_control.md
+++ b/docs/5_remote_control/2_http_remote_control.md
@@ -12,6 +12,9 @@ Remote triggering can be done by sending `HTTP` Requests to the same IP and port
   _Change text on a button_
 - `/style/bank/<page>/<bank>?size=<text size>`  
   _Change text size on a button (between the predefined values)_
+- `/rescan`  
+  _Make Companion rescan for newly attached USB surfaces_
+
 
 **Examples**  
 Press page 1 bank 2:  
--- a/docs/5_remote_control/3_osc_control.md
+++ b/docs/5_remote_control/3_osc_control.md
@@ -14,6 +14,8 @@ Remote triggering can be done by sending OSC commands to port `12321`.
   _Change color of text on button_
 - `/style/text/ <page> <bank> <text>`  
   _Change text on a button_
+- `/rescan 1`
+  _Make Companion rescan for newly attached USB surfaces_
 
 **Examples**
 
--- a/lib/server_api.js
+++ b/lib/server_api.js
@@ -130,7 +130,7 @@ function server_api(system) {
 			system.emit('log', 'TCP/UDP Server', 'debug', 'Rescanning USB')
 			system.emit('devices_reenumerate')
 			response_cb(null, '+OK')
-	 	} else {
+		} else {
 			response_cb(true, '-ERR Syntax error')
 		}
 	})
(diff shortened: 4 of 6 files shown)


llamafilm:
There's probably a more efficient way to do this natively within Companion, but I just wrote a quick and easy Python app that checks for new USB HID devices every second and then triggers a rescan.  Only tested on Windows.

```python
# Monitor for USB changes and trigger Companion rescan when Stream Deck is connected

import time
import hid
import requests

companion_url = 'http://127.0.0.1:8000'

def is_connected(already_connected):
    for d in hid.enumerate(vendor_id=4057):
        if 'Stream Deck' in d['product_string']:
            if already_connected:
                # it was already here
                return True
            else:
                # it was newly connected so trigger Companion
                print(f"Found {d['product_string']} - serial {d['serial_number']} - device ID {d['product_id']}")
                requests.get(companion_url + '/rescan')
                return True

    # it's not connected
    if already_connected:
        # it was just disconnected
        print('Stream Deck was disconnected.')
    
    # else it was never here
    return False


def main():
    connected = False
    
    while True:
        # look for Stream Deck every second
        connected = is_connected(connected)
        time.sleep(1)

main()
```

krocheck:
@Julusian this will cause a merge conflict for #1844, which is OK and I will deal.  I'd rather that then putting this PR into conflict.

krocheck:
@llamafilm [historically], the reason we don't do a periodic rescan is because it is a blocking action in the current architecture and therefore can interrupt precise timing/delays of a bank if it fires at the wrong time, a new device is detected, and syncs up.  Most computers that's not an issue, but Raspberry Pi's could have an issue.  This question has come up many times over the history of the app and I still stand by not doing that by default.  I'm now tempted to allow a user config option for users to turn on a periodic rescan and define the interval.  Thoughts @Julusian?

krocheck (approved):

krocheck (changes requested):
On 2nd thought, @llamafilm  ... it seems like we have good parity of control between the HTTP, TCP/UDP, and OSC remote control protocols.  I'd like to see this implemented and documented across all of them instead of just HTTP.

llamafilm:
Good point, here you go

Julusian:
> @llamafilm [historically], the reason we don't do a periodic rescan is because it is a blocking action in the current architecture and therefore can interrupt precise timing/delays of a bank if it fires at the wrong time, a new device is detected, and syncs up. Most computers that's not an issue, but Raspberry Pi's could have an issue. This question has come up many times over the history of the app and I still stand by not doing that by default. I'm now tempted to allow a user config option for users to turn on a periodic rescan and define the interval. Thoughts @Julusian?

Yeah it is a blocking operation to scan for hid devices that could interrupt other things. A quick test on a pi zero w 2 that I have in front of me says it takes ~10ms to perform, excluding any time we will take to process the results of that.
So I believe that is the reason we don't scan on an interval right now. I have been very tempted to use a library to watch for new devices, and use that to trigger a rescan (like [companion-satellite](https://github.com/bitfocus/companion-satellite/blob/master/src/devices.ts#L32-L34) does), but the library for that is currently a little crash prone on linux, and is often slow at providing builds for new versions of node/electron (it held back companion-satellite on an older electron than companion for a few months).

I have a plan which has been discussed with maintainers for improving detection library. A brief summary is that [node-usb](https://github.com/node-usb/node-usb) already supports proper detection on every platform other than windows. on windows they are for now doing a slow poll. I intend to submit a PR to implement something proper. 
So I think we should use the watching feature of node-usb for the first implementation of auto-watching. If we make it opt-in and label it as experimental/unstable or a bunch of warnings, then its there for those who need it, but with no guarantees. Then one day that will be true watching on every os, without any extra work from us.
If we are concerned about performance impact of the watching or scanning, we could do like what is being done for the usb devices, and run the scans in a child process or worker thread. that may not completely avoid any impact, but will minimise it


As for adding these endpoints to the apis, I am fine with doing that. Its simple enough to not be a burden going forward, and allows for some other workflows (I think people do it today by having the scan triggered from a button, and using http/whatever to press that button)

krocheck:
Thanks @llamafilm! Did you test the additional implementations?

@Julusian I'm good with your plan.

istnv:
A while back I posted a set of `udev` rules that automatically run a rescan when a streamdeck is plugged in. 
It runs a shell script that 'pushes' a button with the re-scan action. Works great on Linux (ubuntu) and Raspberry Pi. This will remove the need to dedicate a button.

krocheck:
> Thanks @llamafilm! Did you test the additional implementations?
> 
> @Julusian I'm good with your plan.

@llamafilm ... didn't answer my question above.  Please and thank you!

llamafilm:
I already answered - yes I tested the other protocols (osc, tcp, udp)

krocheck:
Yeah ... totally my bad.  Let's just go with a momentary lapse of reading abilities!

PR #1845, opened 2022-01-07
feat: periodic USB rescan setting

krocheck:
Adds `Periodic USB Rescan` and `Rescan Interval (in seconds)` user config settings.  They drive an interval in the device manager to execute the periodic rescan.


Julusian:
https://github.com/bitfocus/companion/pull/1844#issuecomment-1007960859

TLDR: We should use https://github.com/node-usb/node-usb#usbdetection instead. It currently uses a polling strategy on windows (Im working on making this better), but other oses use native watching methods. That should minimise the risk of a performance impact

PR #1869, opened 2022-01-19
feat: update streamdeck lib

Julusian:
Updates the streamdeck lib to the version at the new name.
Also consolidates the streamdeck code to use the same wrapper for all models, as they are basically all the same.

Only minimally tested so far. It works with a single MK2, not tested any further yet
Not urgent to merge, it doesnt add anything new for us


krocheck:
Yeah, I like where this is going to offload the individual panel support upstream.  I like the strategy to hold for the moment.

lib/usb/elgato-streamdeck.js
```diff
@@ -0,0 +1,216 @@
...
+function elgato_streamdeck(system, devicepath) {
+	var self = this
+
+	self.config = {
+		brightness: 100,
+		rotation: 0,
+		page: 1,
+	}
+
+	process.on('uncaughtException', function (err) {
```
krocheck: shouldn't we just be doing this upstream in the thread handler?

Julusian: yeah probably, but it was already here so I didnt question it

Issue #103, opened 2018-06-19
Request: Authentication for GUI


Deanbau:
**Describe the feature**
Simple password required for access to web GUI. Password is set in the application window.

**Is this platform dependent (windows, mac, ..)?**
No

**If documentation is required to implement, do you know where to find it?**
None required

**Usecases**
Running Companion on a open network, it would be advisable to have a password on the app controlling your show

willosof:
This is in theory something I wanted to fix for 1.2.0, but after thinking about it, I find a couple of issues with this whole thing.

Running companion on an open network will:
1) Expose OSC triggering
2) Expose ArtNet triggering 
3) Expose the AdminGUI
4) Expose the webserver HTTP /press/page/bank URL's

Should ONLY option 3 be password protected, but 1,2 and 4 be accessible from everyone? The closest thing to a solution here would be to make a IP filter, telling companion which IPs to allow remote communication from. Would that be an option? With that we could block 1,2,3,4 all together.

PerRoine:
I think the IP Blocking is a good way to increase security. and as you state that will increase the security for all exposed ports not only Gui and HTTP wich leaves some gaping holes or reduced functionality for companion of course I do want a enable disable the whole blocking function.

hadphild:
Companion for the install market would be great for just stopping a person changing buttons.

mschoeffmann:
> Should ONLY option 3 be password protected, but 1,2 and 4 be accessible from everyone? The closest thing to a solution here would be to make a IP filter, telling companion which IPs to allow remote communication from. Would that be an option? With that we could block 1,2,3,4 all together.

I think #573 is what it's about here.
I do not think that there is a problem with unencrypted/unprotected osc/artnet/http servers (for triggering) since control networks are normally closed networks.

But a really important thing would be a protection for the admin GUI to prevent unauthorized changes to configuration (for fixed installations).

danielatsavweb:
Has this feature been implemented yet? I don't see it. Thanks.

olliraa:
> Has this feature been implemented yet? I don't see it. Thanks.

Also interested. has this been implemented? Running 2.1.4.

PR #1880, opened 2022-01-25
feat: basic ui locking

Julusian:
A very very basic ui 'locking' mechanism
closes https://github.com/bitfocus/companion/issues/103

The locked screen:
Screenshot from 2022-01-25 00-37-54

Settings:
image

Warning when session is about to timeout:
image

100% insecure to anyone who knows how to javascript, but sufficient to keep out casual users who are likely to break everything.


krocheck:
The only thing that I noted was that I thought putting a password in with timeout set to `0` would lock out Companion on connection but then simply stay unlocked once logged in.  This is not bad and probably makes more sense the way you set it up.

krocheck (approved):

Julusian:
It is now possible to set it up to have a timeout of 0 meaning it will never time out. It will be locked when opening the ui, and can be manually locked in the top corner

PR #1898, opened 2022-02-07
feat: serve webui and docs out of zip files in electron

Julusian:
In electron builds we have all the webui assets loose in a folder on disk, because serving them from the asar often breaks after companion has been running for a few days.
Instead of having them loose, we could serve them from a zip file, to keep the number of files small.

I guess this will have a small performance impact when serving the files, but for the amount of traffic will be getting I dont see it being noticable

This also fixes an issue that will no doubt be reported for with the docs, as they are being served from the asar, and so will hit the same issue
See https://github.com/bitfocus/companion/issues/567 for more info


krocheck:
Overall I think this is fine.  I'll be interested to see if we're introducing new bugs while fixing others.

Only concern from my cursory look (without running it myself) is if the module help includes still work?  Can't tell if I saw code related to that removed and not necessarily replaced elsewhere.

Julusian:
This could move to develop if you prefer? With a safer fix for the docs issue?
I don't see this as having much risk, the zip lib being used is the basis of the newer versions of the yarn package manager, so unless Ive done something really wrong with how its invoked I believe will be stable. Other than that virtual filesystem, the rest of the serving code is indentical

Yes, module help is done here https://github.com/bitfocus/companion/blob/8f05f9f1dccb04a7fb314e90c4aeb41953007f12/lib/instance.js#L138, so is untouched by this.

krocheck:
In terms of where to merge, I think its flip a coin.  From a certain perspective its a late, decently major change, but under the `develop` cycle and waiting I see it getting about the same amount of broad testing in before going to a general release as it would just going in now.  Your flip, so you can't blame me for changing my mind later.

PR #1916, opened 2022-02-14
Add Gitpod config file

MatteoGheza:
I use Gitpod when I have to test some software before installing on my dev machine or when I have to do some basic fix (like fixing a typo).
I added a custom Gitpod config (.gitpod.yml and a custom Docker image at .gitpod.Dockerfile).
It's very fast at installing dev dependencies, and it should be faster [enabling Gitpod prebuilds](https://www.gitpod.io/docs/prebuilds).
BTW, Gitpod it's free if you use it for less than 100 hours monthly, and it's an open source piece of software.

Here I attach a demo video (please ignore the "Nx Console" popup, it's installed on my account).

https://user-images.githubusercontent.com/27729549/153942440-814cfb26-4ec5-495b-b316-90fd5e324ad7.mp4


josephdadams:
We use this with Tally Arbiter and it's pretty nice. @MatteoGheza set it up for us there.

.gitpod.yml
```diff
@@ -0,0 +1,15 @@
+image:
+  file: .gitpod.Dockerfile
+
+ports:
+  - port: 8000
+    onOpen: open-preview
+
+tasks:
+  - init: |
+      ./tools/update.sh
```
Julusian: It wold be better to make this be:
```
- init: |
    yarn update
  command: yarn dev-headless
```

Should `yarn dev-webui` also be run to allow for editing of the webui code? If so the open port above should be changed to `3000`

.gitpod.Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+    unzip \
+    cmake
+
+USER gitpod
+
+RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | PROFILE=/dev/null bash \
+    && bash -c ". .nvm/nvm.sh \
+        && nvm install v${NODE_VERSION} \
+        && nvm alias default v${NODE_VERSION} \
+        && npm install -g typescript yarn pnpm node-gyp" \
```
Julusian: Most of these npm dependencies are unnecessary. `yarn` is the only one that needs to be installed for us

.gitpod.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM gitpod/workspace-base
+
+ENV NODE_VERSION=14.19.0
```
Julusian: I've just added .node-version and .nvmrc files to the repository so that we can start to consolodate all the different node versions defined around the place.
Could you update this to read from there?

MatteoGheza: Done.

.gitpod.Dockerfile
MatteoGheza: Removed pnpm from installed modules.
I think that installing TypeScript and node-gyp can be useful for testing local dev modules and similar things.

.gitpod.Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN NODE_VERSION=$(cat /home/gitpod/.nvmrc) \
+    && echo Node version: $NODE_VERSION \
+    && curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | PROFILE=/dev/null bash \
+    && bash -c ". .nvm/nvm.sh \
+        && nvm install v$NODE_VERSION \
+        && nvm alias default v${NODE_VERSION} \
+        && npm install -g typescript yarn node-gyp \
+        && echo 'export PATH=/home/gitpod/.nvm/versions/node/v${NODE_VERSION}/bin:$PATH' >> /home/gitpod/.bashrc \
+    " \
+    && echo "Node and Yarn installed"
```
MatteoGheza: What do you think about doing something similar in the Companion Docker image (maybe without nvm, but using parts of https://github.com/nodejs/docker-node/blob/main/17/bullseye-slim/Dockerfile https://github.com/nodejs/docker-node/blob/main/17/bullseye/Dockerfile)?
Doing this we can do something fully-optimized for Companion, with the same supported Node version, and keeping the image size small.

Julusian: Will that save much/any space compared to the `node:14-slim` image we are already using?

.gitpod.Dockerfile
Julusian: I'm not convinced that they are useful.

I find that typescript have too many breaking changes for that to work. I think it was 4.5 that made it so that `} catch (e) {` had to be changed to `} catch (e: any) {` or `unknown` could be used instead of `any`, and there have been other things in previous releases.
Every module which needs typescript has it as a devDependency already, so this global version should never be needed.

If node-gyp is needed to build something, then it needs to be defined as a project dependency somewhere, or it will cause the general companion builds to fail. And it will fail on my machine, and on the machine of almost every other module developer.

.gitpod.Dockerfile
MatteoGheza: Nothing or almost nothing, but we can use a specific Node version and we can use less Docker layers, since we can run everything in only one run command. We can do this for both the builder and the production sections of Dockerfile.

Julusian: I don't think we need to use a specific version, we've been doing fine with very vague ranges until now. We only need ^14.18 but I don't thin that is valid node-version syntax.

And you will struggle to get less layers than we have without flattening base image layers, which will stop those layers being reused across images. And we need to consider if the fractionally smaller image is worth the maintainable cost

A few weeks ago I did an overhaul of the docker files, as I noticed the image was a massive 3gb. It's down to a sensible size now, going further will be hard.

.gitpod.yml
MatteoGheza: Currently it doesn't work, since every port on Gitpod is exposed on a different host (example: port 8000 `https://3000-matteogheza-companion-[...].gitpod.io/` and port 3000 `https://3000-matteogheza-companion-[...].gitpod.io/`).
I tried to run this (after exporting the `REACT_APP_SERVER_URL` env), but it doesn't work for CORS reasons.

chrome_oFbWTpUVpn
```
Access to XMLHttpRequest at 'https://8000-matteogheza-companion-[...].gitpod.io/socket.io/?EIO=4&transport=polling&t=O0ZE0Eq' from origin 'https://3000-matteogheza-companion-[...].gitpod.io' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

MatteoGheza:
Now it should work. Fixed a bug with NVM.
Now it installs the Node version defined in `[.nvmrc](https://github.com/bitfocus/companion/blob/master/.nvmrc)`

Julusian:
I haven't forgotten about this PR, but I am having trouble understanding the value of supporting gitpod with it unable to serve both the ui and backend. 

If it can be made to serve the ui and for that to talk to companion correctly, then I think this has some value and I am ok with merging it.

If we need to do a small check to disable the cors check under certain conditions then that is ok with me.
Perhaps when the `DEVELOPER` variable is set, as well as another variable (`DISABLE_CORS_CHECK`?)

Julusian:
This has been sat stuck for far too long now.

I am still struggling to see the value of this in its current form. Not being able to serve the ui in development mode is a deal breaker to me, as that would severely limit what could be worked on with this.  
Overall, I am not opposed to this, but it needs this ui serving for me to be willing to merge it.
Hopefully that is as simple as needing to mess with CORS settings as I suggested above.

If someone is willing to figure this out, please do submit a new PR

PR #1922, opened 2022-02-17
feat: add methods on instance_skel to set and get custom-variable values, and new input field type

Julusian:
This will allow for some system.emit calls to be removed in variable modules, and should avoid any module (except internal) to need to subscribe to the list of custom-variables. Instead of needing to subscribe to that data and update action/feedback definitions, the ui will handle populating the dropdown control

Partner change for the internal module: https://github.com/bitfocus/companion-module-bitfocus-companion/compare/feat/module-custom-variable-api


abebeos:
I like this one very(!) much. Do you expect to merge it within days, or will it take weeks?

Just as a follow-up: do you plan to introduce the same for `InternalVariable/InternalVar`? This would save again many effort. (and I understand at this point your comment to not invest to much refactoring effort for https://github.com/bitfocus/companion-module-bitfocus-companion/issues/53)

abebeos:
@julusian Am I too early in testing this locally?

I get an error when trying this within a local dev env:

```js
ContextData.jsx:61 Failed to load variable values list TimeoutError: Timed out after 5s
    at index.js:37:64
(anonymous) @ ContextData.jsx:61
```

My setup:

* purged `C:\Users\XXX\AppData\Roaming\companion` completely

```sh

PS P:\companion> git status
On branch feat/module-custom-variable-api
Your branch is up to date with 'upstream/feat/module-custom-variable-api'.

PS P:\companion\module-local-dev\companion-module-bitfocus-companion> git status
On branch feat/module-custom-variable-api
Your branch is up to date with 'upstream/feat/module-custom-variable-api'.
```

Julusian:
That error doesnt feel related, but I do not remember if I tested this much.

This is currently on hold, as we want to think about whether it makes sense for all modules to be able to set custom variable values. For most modules they should probably care about get the value of any variable, rather than specifically custom variables.
Maybe it should only be the internal modules, and the generic connection modules that should be able to set custom variables.

abebeos:
You could finish it as undocumented/alpha, with a remark that it will possibly be removed. (Personally, I like to have full freedom. Devs should know what they do, and keep away from stuff they do not master).

Julusian:
> You could finish it as undocumented/alpha, with a remark that it will possibly be removed. (Personally, I like to have full freedom. Devs should know what they do, and keep away from stuff they do not master).

That is the situation we are in today with the many system.emit calls to hook into things that modules shouldnt be. Which makes it hard for us to do internal changes without breaking a load of modules. A lot of those calls are guaranteed to break at least a little in 2.3/3.0 as they will become asynchronous

Additionally, we would like to move towards being able to have modules be 'plugins' which means we would greatly benefit from a stable api, that will remain compatible across multiple versions, so having experimental things that we expect to change go against that plan. Especially as this will allow modules not under our control to be used, so we can't go and fix them as we break things like we can today (if we notice we broke something today that is. when I made it possible to batch update variables, I broke ~10 modules which were listening for changes to variables)

abebeos:
> That is the situation we are in today with the many system.emit calls to hook into things that modules shouldnt be.

You can't block experienced developers to use whatever they wish to use. I can do for example with python-code nearly whatever I want, shouldn't be much different for node.js.

> so having experimental things that we expect to change go against that plan.

I meant that you can implement it for the alpha/beta/RC 2.3,  This way you could get feedback, do some trials etc. to see what is the right design decision (e.g. to make the functionality internal-only prior to release). In the end, you could keep it on the branch without merging until you decide.

(I could give you e.g. some feedback from the `customvar` trial-implementation, which is currently my main focus)

Julusian:
> You can't block experienced developers to use whatever they wish to use. I can do for example with python-code nearly whatever I want, shouldn't be much different for node.js.

Well no, but also there has been too much free reign. I am at the point in the 'feat/module-lib' branch where I now need to reimplement a lot of the `system.emit` calls to avoid breaking modules. Its going to be tedious because the new threading (and by proxy isolation) model means modules can only access things we explicitly expose to their threads.

> I meant that you can implement it for the alpha/beta/RC 2.3, This way you could get feedback, do some trials etc. to see what is the right design decision (e.g. to make the functionality internal-only prior to release). In the end, you could keep it on the branch without merging until you decide.
>
> (I could give you e.g. some feedback from the `customvar` trial-implementation, which is currently my main focus)

Ah right. Yeah that would be a reasonable thing to do. Which is kind of where we are at with it. There is this proposal, and we need to figure out if it is genuinely useful to modules. At some point we could ask module devs for input, but not until 2.2 is finalised, and possibly 2.3 is ready for general use by devs (every time I try it I find something else completely broken).
At some point I should also try and get feedback on the new module-base api and structure, but again its way too early as I havent written any of the tooling that would make it easy to use.

abebeos:
@Julusian , please ignore https://github.com/bitfocus/companion/pull/1922#issuecomment-1046286589, it works find on my side (forgot to `./tools/update.sh` and disable some of my dev modules)

abebeos:
@Julusian , here is my updated code, in the related branch:

https://github.com/abebeos/companion-module-bitfocus-customvar/blob/feat/module-custom-variable-api/index.js

To get rid of "system", I would need replacements for:

```sh
self.system.emit('custom_variable_set_expression', opt.name, opt.expression)
self.system.emit('variable_get', id[0], id[1], (v) => (value = v))
self.system.emit('instance_actions', self.id, actions)
```

and finally, to make it feature-complete, an `InternalVariable` field similar to the introduced `CustomVariable` UI field type (needed for `custom_variable_store_variable` action).

Once I can complete this, I can provide further feedback.

Julusian:
I dont think this is necessary.
As mentioned above, modules should get variable values by parsing a variable (or perhaps we should add a 'variable' type dropdown for them to use instead of this 'customvariable' one).
And modules should not be setting custom variables. If they 'need' to do this, they should be setting a normal variable, then the user can use an action to copy the value into a custom variable.

PR #1934, opened 2022-02-23
Add a check to the rest service for http or https

JTF4:
When I was working on a module, I accidentally forgot to add http:// to the IP before sending it to the rest service. I thought that it might be nice to have some checks in place to check if incoming url has `http://` or `https://` at the beginning, and if it doesn't, then append a default `http://` to it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "companion",
-	"version": "3.0.0",
+	"version": "3.1.0",
 	"description": "Companion",
 	"main": "main.js",
 	"type": "module",


Julusian:
Should this be targetting 2.2 or 2.3?
In 2.3 this class is being marked as deprecated with the intention of being removed or rewritten in 2.4

krocheck:
I don't mind bolstering this a little even if its deprecated but I'm not a fan of assuming `http://` when none is provided.  I'd rather invoke the `cb` and send an error back to the module.

JTF4:
I like the CB idea. I'll make it cb an error message and also emit one to debug.

JTF4:
Just changed it to respond with an error

lib/Service/Rest.js
```diff
@@ -9,6 +9,12 @@ class ServiceRest extends CoreBase {
 		this.system.on('rest_get', (url, cb, extra_headers, extra_args) => {
 			this.debug('making request:', url)
 
+
+			if(!url.startsWith('http://') || !url.startsWith('https://')){
+				cb(true, { error: 'Url does not start with http:// or https://' })
```
krocheck: @Julusian should this technically be: 
`cb(true, { error: new Error('Url does not start with http:// or https://') })`

JTF4:
I had a brain fart when I did the logic and used an `or` operator instead of an `and` operator. With the `or` operator it will always  report an error. I've changed it to an `and` operator to fix it.

lib/Service/Rest.js
Julusian: Yeah probably should be

JTF4: I'll update that then.

krocheck: @JTF4 are we still expecting this update?

JTF4: @krocheck yes. I got bogged down with some other projects. I'll probably update the commit tomorrow or the day after.

JTF4: @krocheck would it be worth promisifying the process? Or is that I thing for a future companion release?

PR #1945, opened 2022-02-27
major: Add first-run configuration wizard

krocheck:
This feature adds a one-time configuration wizard as a modal popup for new installs and upgrades.  If it detects the lack of a `userconfig` setting called `v22_wizard` it will run and if the wizard is successfully completed or canceled out that setting is set to `true`.  A button has been added to the sidebar to manually launch the wizard.  It is non-destructive, meaning there is a confirmation step before it applies changes and can be canceled up to that point.
image

#### Welcome
Simple welcome step indicates new and upgrade users should step through and review the settings.
image

#### USB Surface Detection Settings
Provides plain language on if the Elgato Plugin should be enabled or not, as well as the option for X-keys support.
image

#### Remote Control Services
Provides enable checkboxes for all of the listeners and displays dependent settings for the enabled settings.
image

#### Admin GUI Password
Provides the settings for configuring an admin password.
@Julusian I need a little help here.  I couldn't figure out a way to disable the `Next` button if this is enabled but no password is entered.
image

#### Review Changes
Provides a summary of how the environment will be configured.  The language does adapt a little bit based on this being a first run or a manual run.
image

#### Finish
After applying the changes it displays information on next steps including a notification to restart Companion should the changes made to USB detection require it.
image

I tested this fairly thoroughly.  @willosof you're welcome to fancy up the UI.  I did the best I could.

### DIFF
--- a/app.js
+++ b/app.js
@@ -196,6 +196,10 @@ system.ready = function (logToFile) {
 	system.on('exit', function () {
 		elgatoDM.quit()
 	})
+
+	setTimeout(function () {
+		system.emit('ip_rebind')
+	}, 2000)
 }
 
 exports = module.exports = system
--- a/lib/server_http.js
+++ b/lib/server_http.js
@@ -57,7 +57,6 @@ function server_http(system, express) {
 	system.emit('config_object', function (config) {
 		self.config = config
 		system.on('ip_rebind', self.listen_for_http)
-		self.listen_for_http()
 	})
 
 	return self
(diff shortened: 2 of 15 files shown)


Julusian:
I wonder if the userconfig key should be `"setup_wizard": "v2.2"`, so that we can reuse the field when we want to add another wizard for new things we add in a later version. That shouldnt change any real logic for now, as this being the first version makes sense to be doing a truthy check rather than checking the explicit value

krocheck:
That would work.  I put it with a version in the name so that in a v2.3 or v3 we can detect for an upgrade vs new install.  I'll probably do it as a number though ... either 22 or 220 so you can do an easy numeric compare.  And that way I can set initial default to 0.

PR #1948, opened 2022-03-02
unify custom_variable events

abebeos:
(early draft, re https://github.com/abebeos/companion/issues/1)


abebeos:
@Julusian , this is an example for a slow step-by-step refactoring. Each step is essentially mergeable, as it improves the general code-quality. The refactoring-step which moves the event-code to reusable methods needs ideally to be applied to all events within variable.js. Further refactoring goes then on, after thinking a little bit based on the new available constructs.

(it does not look like i can get sponsors for this work, but will do a little more anyways, out of my curiosity)

Julusian:
I appreciate the PR, but I have been sitting with this for a month now and I still dont think that this is a way we want to go for now.

With 2.2.0 stable released, the master/beta branches only want to see bug fixes or other essential changes. We are explicitly trying to avoid the master/beta branches having continual changes, because we want to use those to release updates to the modules more frequently. 

I am still happy with the strategy we are undertaking in develop, so I dont see anywhere for this to go. The only place I can think of would be to start another branch for this parallel refactoring track, but then why? Unless we abandon the work we started in develop and start it again, I dont see another refactoring branch being beneficial

PR #1956, opened 2022-03-11
feat: Add healthcheck and expose Satellite API port in Dockerfile

hammady:
This PR does the following:
1. It adds a healthcheck to the Dockerfile so that the container is marked as healthy/unhealthy. This is useful when running it as a swarm service so that it gets automatically restarted.
2. Expose the Satellite API port. As per Docker best practices, all open ports should be documented in the Dockerfile so that they are available for discovery to other tools and by other users
3. Minor cleanups to the `.dockerignore` file

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,7 @@
+.dockerignore
 node_modules
 */node_modules
-module-local-dev/*/node_modules
-module-local-dev/*/.git
+module-local-dev
 *.log
 electron-output
 webui/build
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,8 @@ RUN apt update && apt install -y curl && \
 ENV COMPANION_CONFIG_BASEDIR /companion
 RUN mkdir $COMPANION_CONFIG_BASEDIR && chown node:node $COMPANION_CONFIG_BASEDIR
 USER node
+# Export both web and Satellite API ports
+EXPOSE 8000 16622
 
 HEALTHCHECK --interval=30s --timeout=3s --start-period=5s --retries=3 CMD [ "curl", "-fSsq", "http://localhost:8000/" ]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,12 +38,17 @@ FROM node:14-bullseye-slim
 WORKDIR /app
 COPY --from=companion-builder /app/	/app/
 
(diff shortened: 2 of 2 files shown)


krocheck:
> 2\. Expose the Satellite API port. As per Docker best practices, all open ports should be documented in the Dockerfile so that they are available for discovery to other tools and by other users

@hammady @Julusian is there anything we can/should do for our other services that user configurable or is that not a major issue compared to this?

hammady:
I'm not aware of other services. In principal the Dockerfile ports are purely for documentation and discovery (-P in the docker run command will publish all documented ports to random ports on the host).

Julusian:
as they are user chosen ports I dont know if it would make sense to.
but the satellite port is currently hardcoded so makes sense

hammady:
Docker works differently when it comes to choosing ports. Even if the port was configurable at the app level, the Dockerfile would make the choice and configure the app accordingly (e.g. by setting an environment variable) making sure the same port is EXPOSEd. This is not a problem because the port number is only relevant to the container. When a user runs the container, they can choose to publish the container port to the same or to a different port on the host (-p 16622:16622). This allows multiple containers to expose the same port but a user is responsible for choosing different host ports when publishing. This is explained in the context of vanilla docker engines but the same concept applies to orchestrators like swarm and kubernetes but using different terminology.

Julusian:
But as we don't support setting those user chosen ports though environment variables (they are controlled through the ui) 
image
should they be exposed?

Looking at the list in the ui, there are some that probably should be as they are fixed numbers, even though they are disabled by default.
I suppose there are also some modules which start their own servers for devices to connect back to, but we definitely shouldnt do anything for them

hammady:
I see the issue here. Users of Companion expect to configure ports from the UI, but if running on docker, the Dockerfile should document the exposed ports. It is like when you allow users to configure the ports, but they should be aware that those should be open as well on the firewall.
Here is my suggestion:
- Allow those ports to be configurable using environment variables
- In the Dockerfile, set environment variables to configure all core services
- Expose those fixed ports in the Dockerfile so docker users can expect which ports they need to publish on the host
- In the web interface, disable the text input fields for ports that are already configured using environment variables
- Put a note in that web page saying that those disabled text fields correspond to ports that are configured via environment variables

PR #1976, opened 2022-03-29
feat: add support export and show config folder

krocheck:
Adds "Show config folder" to the tray menu and "Export support bundle" to the log tab.

Minor change to headless that `companion.log` writes to the config directory instead of executable path so it can be included in the export.

image
image


Julusian (approved):
I think it would be good to warn users about the potential for sensitive information in these exports. Perhaps the button should show a dialog to warn of that, with a button on the dialog to do the download?
Other than that I am happy :)

PR #2005, opened 2022-04-26
feat: report client wrappers which do not have an error handler

Julusian:
Various modules have forgotten to add error handlers to the udp/tcp wrappers we provide. This means that when they throw an error it will either crash companion or trigger the blocking error dialog.

To help mitigate this, we could add a check like this to log if a socket has been created and has no error handler setup.
I havent tested this yet.

Do all module devs run the 'dev' versions of commands? I know one who doesn't becuase the console is too noisy with internal logging. We might want to look at reducing the logging at the same time, or adding a 'module-dev' level of logging.

### DIFF
--- a/tcp.js
+++ b/tcp.js
@@ -126,14 +126,12 @@ function tcp(host, port, options) {
 	// Let caller install event handlers first
 	setImmediate(self.connect.bind(self))
 
-	if (process.env.DEVELOPER) {
-		setTimeout(() => {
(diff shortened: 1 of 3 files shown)


krocheck:
I don't think having this console only report if `process.env.DEVELOPER` is needed.  At least if a user sees it they'll report it then.

Julusian:
Sure, that resolves my concern about module-devs not seeing this too

PR #2009, opened 2022-04-28
feat: use jsonpatch for pushing changes to ui connections

### LINKED ISSUE #2006: [BUG] Companion Web Interface uses infinite bandwidth

- [X] I believe this to be a bug in companion

### Describe the bug

When loading the companion Web UI on a machine that is not the companion server I can the WebUI uses as much bandwidth as it can possibly capture. This can be well upwards of 10 megabytes per second. I cannot seem to see these requests in the network tab of the inspector in chrome. But beyond that not sure how to track this down. Below is a video so you can see the bandwidth usage in action. I have tested this on the past few betas as well as the latest. The screen recording is not from the latest, but actions are the same on the latest beta.

https://user-images.githubusercontent.com/9466882/165615950-4aacd406-fd60-483b-8061-3040e301a52f.mp4

### Steps To Reproduce

_No response_

### Expected Behavior

_No response_

```markdown
- OS: Mac OS 12.3 (Arm)
- Browser: Chrome
- Companion Version: 2.2.0 (2.2.0+4118-beta-6acb1059)
```

### Additional context

_No response_

Julusian:
Closes #2006

Reduce data sent to the client by sending json-patch instead of full objects.

Probably best for this to hit beta after 2.2.2 is released. It is very possible I made a mistake or a bad assumptions causing some data to get stuck or companion to crash


krocheck:
You want to wait until 2.2.2 on this one now?  Seems OK to start rolling this in.

Julusian:
That should be 2.2.3 now (assuming 2.2.2 happens soon as the tiny update for those few broken modules).
But yeah, there isnt much reason to hold off on this now. I only deployed it for testing myself yesterday, but Im unlikely to notice any issues there

PR #2019, opened 2022-05-17
Install dependencies of local dev modules in Docker

hammady:
Automatically install dependencies of local modules in docker. This also caches the dependencies so they will only be installed once. To enable this behaviour one should mount the local dev folder when running the container, example:

```bash
docker build . -t companion:dev # just once
docker run -it --rm -v $PWD/module-local-dev:/app/module-local-dev companion:dev
```

If no volumes are mounted as above, everything stays the same.

### DIFF
--- a/module-dev-docker-entrypoint.sh
+++ b/module-dev-docker-entrypoint.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+set -e
+
+# Install dependencies for all local dev modules
+
+if [ -d /app/module-local-dev ]; then
+  echo "Installing dependencies for all local dev modules..."
+  for module in $(ls /app/module-local-dev/); do
+    cd /app/module-local-dev/$module
+    if [ -d node_modules ]; then
+      echo "Skipping installation of module $module dependencies, to force, remove its node_modules directory"
+    else
+      echo "Installing dependencies for module $module"
+      npm install
+    fi
+  done
+fi
+
+exec "$@"
--- a/module-dev-docker-entrypoint.sh
+++ b/module-dev-docker-entrypoint.sh
@@ -11,7 +11,7 @@ if [ -d /app/module-local-dev ]; then
       echo "Skipping installation of module $module dependencies, to force, remove its node_modules directory"
     else
       echo "Installing dependencies for module $module"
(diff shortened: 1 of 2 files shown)


module-dev-docker-entrypoint.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if [ -d /app/module-local-dev ]; then
+  echo "Installing dependencies for all local dev modules..."
+  for module in $(ls /app/module-local-dev/); do
+    cd /app/module-local-dev/$module
+    if [ -d node_modules ]; then
+      echo "Skipping installation of module $module dependencies, to force, remove its node_modules directory"
+    else
+      echo "Installing dependencies for module $module"
+      npm install
```
Julusian: This should be a `yarn install --prod`.
The companion project requires all the modules use yarn for dependencies, so installing with npm will bypass the yarn lockfiles that are committed.
And `--prod` means we will only get production dependencies, which is all that should be needed here and matches what would happen if they were included as a builtin module

Julusian (commented):
I like it. As you can see, I never did get around to implementing then when redoing the docker image a while ago :)

module-dev-docker-entrypoint.sh
hammady: Done.

module-dev-docker-entrypoint.sh
```diff
@@ -0,0 +1,19 @@
...
+    if [ -d node_modules ]; then
+      echo "Skipping installation of module $module dependencies, to force, remove its node_modules directory"
+    else
+      echo "Installing dependencies for module $module"
+      yarn install --prod
+    fi
+  done
+fi
+
+exec "$@"
```
Julusian, line 19: Will this be run in the correct directory?
There are some cd being done above, so the cwd will be `/app/module-local-dev/something`.
So will the `exec ./headless_ip.js` work as expected?

hammady: Good observation. The cd works for multiple dependencies because they are absolute. The exec replaces the process including its working directory so yes the headless still works. 
I tried it on my machine with no issues and you can also double confirm.

PR #2039, opened 2022-06-04
Feat: Instance HTTP request handler

### LINKED ISSUE #2017: [Feat] Route HTTP requests from Companion to modules

### Describe the feature

This request follows the discussion thread on slack: [https://bitfocusio.slack.com/archives/CFG7HAN5N/p1652451464856329](https://bitfocusio.slack.com/archives/CFG7HAN5N/p1652451464856329)

As Companion already runs a HTTP server for the web interface it would be beneficial if in some limited capacity this could be exposed to modules that need external interactivity. 

The main difficulty of implementing such a feature would be to give modules the control and data they need to be able to respond to requests while limiting access to the underlying Express server so that no module can negatively impact the function of Companion itself.

My vision for this feature would be for Companion to have a URL path such as `/instances/` and for each running instance to have a path within that base on the instance name (this allows for multiple instances of the same module to have unique paths, and placing them inside `/instances/` will prevent a naming conflict should someone name their instance something that's already in use, such as `connections` or `buttons`).

Companions routing would simply need a wildcard handler for each instance, as any further routing would be handled within the module itself by parsing the path. Companion could then send an event to the module along the lines of:
```
{
  "method": "GET",
  "path": "/instances/SomeModule/data"
  "headers": {
    ...all headers from the request
  },
  "body": ...Body, if present
}
```

The module would then need to send a response back to Companion
```
{
  "status": 200
  "body": ...some JSON
}
```

 Some things that'll need discussion:
- Naming scheme for the each modules path. How to handle modules using names with characters not usable in a URL path?
- Does Companion need to expose request data beyond just request method, headers, and body?
- Should modules need to 'register' a handler with Companion, so that requests to a module without a HTTP handler can be immediately 404'd rather than timing out waiting for the module to respond.
- Should there be some sort of rate limit to prevent excessive requests to one module locking up Express?

### Usecases

Currently my Google Sheets module, periodically polls the spreadsheets API to get sheet values and metadata, but Googles rate limit is quite restrictive so the more apps that need to poll the data reduces the responsiveness. If I was to use a Google Sheet data source in vMix that'd be a 2nd app sharing the same rate limit so the polling for both vMix and Companion would have to be halved.

Alternatively, if I could poll for the data within the Companion module and then serve it over HTTP to others apps, such as a URL `http://localhost:8000/instances/gsheets/data` I can then point vMix or any number of other apps to that URL for data and there'd only be 1 app actually making requests to Google allowing for the rate limit to be used as efficiently and responsive as possible.

thedist:
Resolves #2017 

Implements a HTTP route in express `/instance/:label`, where the `label` param can either be an instance id, or the instance label.

The request/response objects are then emitted where `instance.js` handles checking if there's a matching instance with a http handler, if not it 404's.

If there is a matching instance with a http handler, a race is started between the `handleHttpRequest` function within an instance (which can either return a response, or return a promise that resolves a response) and a timeout promise so that should an instance not respond promptly (currently set to 30s) it'll return a 500 response.

The request object that's sent to the instance is a limited version of the Express request, such as:
```
{
  baseUrl: '/instance/vmix',
  body: 'someData',
  hostname: 'localhost',
  ip: '127.0.0.1',
  method: 'POST',
  originalUrl: '/instance/vmix/test/?foo=123&bar',
  path: '/test/',
  query: { foo: '123', bar: '' }
}
```

With the response object being any of:
```
{
  status?: number
  headers?: Record<string, any>
  body?: string
}
```

Also included in this PR is the inclusion of the `body-parser` module, so that requests to the handler containing a form, json, or text body can be parsed.


instance_skel.js
```diff
@@ -432,4 +432,8 @@ instance.extendedBy = function (module) {
 	util.inherits(module, instance)
 }
 
+instance.handleHttpRequest = function () {
```
Julusian: I would prefer this to not have a default implementation. You are already checking if this function exists before calling it, so dropping this should have no change.
And with a default implementation companion wont be able to tell what modules have implementations and what are using the base implementation

lib/instance.js
```diff
@@ -147,6 +147,65 @@ function instance(system) {
 		}
 	})
 
+	self.system.on('instance_http_req', function (req, res) {
+		let instance
+		res.set('Content-Type', 'text/plain')
```
Julusian, line 152: Are modules able to override this if it is done here?

lib/instance.js
```diff
@@ -147,6 +147,65 @@ function instance(system) {
...
+				originalUrl: req.originalUrl,
+				path: req.path,
+				query: req.query,
+			}
+
+			const httpTimeout = () => {
+				const delay = 30000
+				const defaultResponse = {
+					status: 500,
+					body: JSON.stringify({ status: 500, message: 'Internal server error' }),
```
Julusian: This might be better as a 502 Gateway timeout

Julusian (commented):
Looking good :)

lib/instance.js
thedist: Yes, modules can specify whatever headers they like, and if they specify a 'Content-Type' of something else, such as `application/json` then that'll overwrite the `text/plain`.

The default is just there so there's a consistent response if no content type is specified by the module, otherwise it would be Express trying to guess the type based on the body which is not always the best.

lib/instance.js
thedist: I went with a generic 500 as that'd cover whatever the error may be.

There's also the possibility for modules themselves to do their own errors and return more specific error codes/messages themselves as long as they do so before that default delay.

instance_skel.js
thedist: Sure, this can be changed. I only added it when the PR Checks errored as I originally had it as an abstract type and so didn't pass the checks when it encountered a module that didn't define it (ie, all of them). I've already removed the abstract keyword so I should be able to remove that default function and still pass the PR checks.

lib/instance.js
thedist: Changed to 504 Gateway Timeout.

I do think in the future we may want to consider more specific error messages, to help indicate if it's on the side of Companion, or the Module. Generic errors should be fine for the time being though.

Julusian:
@thedist is this ready to be merged?

thedist:
Functionally it works without issue.

One issue though is that when trying to build using `windist` I get an error `sharp: Installation error: Prebuilt libvips 8.12.2 binaries are not yet available for win32'-x64'` that causes it to fail. Not sure if it's related to this PR and the addition of a new module (body-parser) or if it's unrelated as I'm not particularly familiar with the build process as I always run a dev version. If this error is unrelated, or potentially on my end rather than this PR, then yeah this PR is good to be merged.

PR #2049, opened 2022-06-18
feat: learn button for actions and feedbacks

Julusian:
This lets modules define a `learn` callback for actions and feedbacks to help users fill in the properties.
image

The idea is that instead of having to define the values inside of companion, you can set it up on the device and press learn to populate the current values into the action or feedback.

The methods take the same data as supplied to subscribe and unsubscribe, and the method is expected to return the new options object. Undefined/null is also valid if the method wishes to 'reject' the call.

For example, I have tested this with the `internal: Check variable value` feedback, so that it will populate the `Value` with the current value of the variable.

The bmd-atem module will benefit from this, as some of the supersource layouts are hard to define inside companion without a lot of trial and error. This will let users build their layouts in other software, and 'capture' those onto buttons in companion.

### DIFF
--- a/instance_skel.js
+++ b/instance_skel.js
@@ -193,6 +193,7 @@ instance.prototype.setActions = function (actions) {
 				if (action && action.options) {
 					action.options = serializeIsVisibleFn(action.options)
 				}
+				if (action?.learn) action.hasLearn = true
 				return [id, action]
 			})
 		)
--- a/instance_skel.js
+++ b/instance_skel.js
@@ -247,6 +247,7 @@ instance.prototype.setFeedbackDefinitions = function (feedbacks) {
 				if (feedback && feedback.options) {
 					feedback.options = serializeIsVisibleFn(feedback.options)
 				}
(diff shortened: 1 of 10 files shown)


Julusian:
Example implementation in internal module: https://github.com/bitfocus/companion-module-bitfocus-companion/commit/0cf09c4461414fc4c3b85cfa7b72d76906995f66

MathiasGIZ:
Is it possible to implement this learning function as an internal action?
The usecase may be as mentioned above, but with the possibilty to save changes made on an Atem SSRC by hitting a button on your streamdeck, or by press and hold the button via duration group.
The action may be like this: "Learn values for all actions on button".

Julusian:
> The action may be like this: "Learn values for all actions on button".

That could work, but does feel a bit dangerous as modules could implement support for 'learn' in future updates. so what works today in 3.1 may do more in 3.2 (in a way you dont intend).
But I am not opposed to it being done like this

Could you open this as a new issue, it will get lost/forgotten in this long closed PR

PR #2053, opened 2022-06-20
Fix: http request types

thedist:
Minor typo fix to the type interface for the HTTP request update


thedist:
This has build conflicts with the typescript compilation of the vMix module as the one currently in beta has a type error with this PR. But v2.1.1 of the vMix module has the corrected types on its end so when that's added into the beta build this'll pass the PR tests.

PR #2056, opened 2022-06-26
Set custom variable value using remote control protocols

albertorighetto:
This PR will allow to set the value of a custom variable by using remote control protocols such as TCP/UDP/OSC/HTTP.
The variable must be already defined in Companion.

I choose che command syntax in a reasonable way but please let me know if it should be changed to something more meaningful or future-proof; below are my thoughts about the naming convention:

OSC and HTTP divides the "SET" part from "CUSTOM-VARIABLE" (i.e. /SET/CUSTOM-VARIABLE/) because if at any point in the future we need to add another "SET" command for something else we could maintain a reasonable syntax (i.e. /SET/OTHER-THING/. Maybe we should also change TCP/UDP to SET CUSTOM-VARIABLE instead of SET-CUSTOM-VARIABLE?

TCP/UDP syntax uses double quotes for name and value, allowing an easy parsing of them (both may include white spaces or special chars such as escaped double quotes).

HTTP syntax allows for easy GET requests, sending both the custom variable name and the value as parameters that could use URI encoding for special chars.

OSC syntax on the other hand includes the custom variable name as part of the address (/set/custom-variable/name_of_the_variable) and accepts any value (or the first element of multi-parameter message). I think this is easier to integrate in external software that may not have the ability to send multi-parameter OSC messages and also allows for arbitrary variable names, including the ones with forward slashes "/" or spaces.

### DIFF
--- a/docs/5_remote_control/http_remote_control.md
+++ b/docs/5_remote_control/http_remote_control.md
@@ -12,6 +12,8 @@ Remote triggering can be done by sending `HTTP` Requests to the same IP and port
   _Change text on a button_
 - `/style/bank/<page>/<bank>?size=<text size>`  
   _Change text size on a button (between the predefined values)_
+- `/set/custom-variable/?name=<name>&value=<value>`  
+  _Change custom variable value_
 - `/rescan`  
   _Make Companion rescan for newly attached USB surfaces_
 
@@ -25,3 +27,6 @@ Change the text of button 4 on page 2 to TEST:
 
 Change the text of button 4 on page 2 to TEST, background color to #ffffff, text color to #000000 and font size to 28px:  
 `/style/bank/2/4/?text=TEST&bgcolor=%23ffffff&color=%23000000&size=28px`
+
+Change custom variable "cue" to value "intro":  
+`/set/custom-variable/?name=cue&value=intro`
--- a/docs/5_remote_control/osc_control.md
+++ b/docs/5_remote_control/osc_control.md
@@ -14,6 +14,8 @@ Remote triggering can be done by sending OSC commands to port `12321`.
   _Change color of text on button_
 - `/style/text/ <page> <bank> <text>`  
   _Change text on a button_
+- `/set/custom-variable/<name> <value>`  
+  _Change custom variable value_
 - `/rescan 1`
   _Make Companion rescan for newly attached USB surfaces_
 
@@ -27,3 +29,6 @@ Change button background color of button 5 on page 1 to red
 
 Change the text of button 5 on page 1 to ONLINE  
 `/style/text/1/5 ONLINE`
+
+Change custom variable "cue" to value "intro":  
+`/set/custom-variable/cue intro`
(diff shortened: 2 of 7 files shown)


krocheck:
@Julusian I'm ok with this, just curious if we have any potential issues with quotes and escape characters.

istnv:
Adding SET to the address does diverge from HTTP and OSC standard usage: 
A 'PUT' operation to an HTTP address 'sets' a value of 'something' and 'GET' retrieves it 
OSC generally uses an empty parameter list as a request for the current value and uses the parameter list to update that value.

albertorighetto:
Thanx for the feedback.

What you're suggesting is to change
- OSC syntax to:
   /custom-variable/name_of_the_variable [<value>]
- HTTP syntax to:
  /put|get/custom-variable/?name=<name>[&value=<value>]
- maybe TCP syntax to:
  PUT|GET CUSTOM-VARIABLE "<name>" ["<value>"]

I'm fine with this suggestion as this opens the ability also to retrieve the values remotely.
On the other hand this will differentiate the syntax of the commands between the protocols.

My personal opinion about "PUT" is that even if it could be technically correct for HTTP it is less straightforward than "SET" when used as a command across multiple protocols.

Would love to hear other opinions on this before changing the code.

istnv:
TCP/UDP do not have a 'standard' for retreiving/updating information so `SET` would be more consistent with the other Companion actions

HTTP uses TCP and defines how clients and servers communicate. `GET` and `PUT` are types of page requests, not part of the url/address. You don't normally see that part.

Since Companion currently uses `GET` for doing things, adding `PUT` may be overkill and confusing to a lot of people currently using Companion. Very few are programmers :)

I would suggest something similar to OSC:

-  A bare address like `/custom-variable/myVar` returns the value
- An address with a value `/custom-variable/myVar?val=new%20value` sets the variable `custom:myVar` to `new value`


The OSC addressing is closer to how other systems respond.
OSC does have restrictions on what characters are allowed in the address:
#### Printable ASCII characters not allowed in names of OSC Methods or OSC Containers
character | name | ASCII code (decimal)
-- | -- | --
’ ’ | space |  
\# | number sign | 35
\* | asterisk | 42
\, | comma | 44
\/ | forward slash | 47
\? | question mark | 63
\[ | open bracket | 91
\ | close bracket | 93
\{ | open curly brace | 123
\} | close curly brace | 125

Julusian:
I think we should change the tcp/udp syntax to be `custom-variable (.+) set-value (.*)\n` to make it more consistent with setting a buttons text. It also gives a clearer scope of other `custom-variable` commands that can be implemented (get-value, set-default, get-default). This helps a little with some escaping issues (`custom-variable test set-value ""my value""` feels so wrong), but still isnt perfect.

@istnv I have taken some of your suggestions too. I havent taken everything for now as I think we would be better to do an overhaul of the various apis in v3.1 (or v2.4 if that will exist) and run both the new and old side by side for a couple of releases so doing it properly now will end up causing more pain if it clashes with the new implementation.  
I am writing a task to have a proper look at overhauling the protocols, as I think that is needed too

istnv:
@Julusian All good. I just wanted to bring some of the generally accepted usage for some of the protocols into the discussion. Something to keep in mind when development progresses.

PR #2127, opened 2022-09-19
button editor improvements

Julusian:
* actions and feedbacks can be collapsed/expanded to reduce screen usage
* actions and feedbacks can be duplicated
* actions can be dragged between sets (eg between down and up)

image

### DIFF
--- a/lib/Controls/Button/Base.js
+++ b/lib/Controls/Button/Base.js
@@ -124,6 +124,36 @@ export default class ButtonControlBase extends ControlBase {
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


Thymeauthe:
An additional Idea would be to duplicate all actions on a button to applicate on the same one with an other instance.

With Blackmagic Atem for example, we could program all actions we need for differents ATEM instances. We will just have to activate or not instance to do work on the switcher we use.

I know we could duplicate all the page to an other one for an other instance. But the idea would be insteresting to save pages.

PR #2129, opened 2022-09-20
module: Added neumannmueller-stageflow v1.0.2

jonkad:
Now with

- package.json
- yarn.lock

and request in correct branch ;-)


josephdadams:
Module added via proper channels

PR #2147, opened 2022-10-16
feat: identify legacy modules in ui

Julusian:
Instances list: 
image

Add instance list:
image

Confirmation when adding:
image


webui/src/Instances/AddInstance.jsx
```diff
@@ -89,8 +122,25 @@ const AddInstancesInner = memo(function AddInstancesInner({ showHelp, configureI
...
 
+	const confirmContent = (
+		<>
+			<p>
+				This module has not been verified to be compatible with this version of companion. It may be buggy or completely
+				broken.
+			</p>
+			<p>
+				If this module is broken, please let us know in{' '}
+				<a target="_blank" rel="noreferrer" href="https://github.com/bitfocus/companion/issues/9999">
```
Julusian: An issue needs to be created, with the correct id inserted here

Issue #2155, opened 2022-10-28
UI: Warning before exiting trigger modal if changes have been made


bryce-seifert:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

When you're working in a trigger modal, if you accidentailly click outside the modal before you've save you lose all of your work.

I suggest that if the user has made any changes, ie added actions or feedbacks, that there be a prompt before closing to help prevent losing work from errant clicks.

Julusian:
This is no longer an issue in v3.0. See #2165 
I shall leave this open for now, as this could be improved in 2.4 pretty easily

Julusian:
This is in the 3.0 betas

Issue #2040, opened 2022-06-05
Add Trigger Browser covers settings window


MeestorX:
### Is this a bug in companion itself or a module?

- [X] I believe this to be a bug in companion

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When adding a new Trigger (I tested using a Feedback trigger), and using the Browse function, after adding a new trigger, the Edit trigger window pops under the Browse window. The Edit should pop up to the right, or the Browse window should be to the left (or both) so one doesn't cover up the other.
![image](https://user-images.githubusercontent.com/29129658/172057207-be41cfb7-04bc-4ce8-bbf4-293b3a6b663f.png)

### Steps To Reproduce

Add a new trigger and use the Browse button to add it.
Trigger Browser window covers up the Edit window.
Clicking "Done" on the Browser window reveals the Edit Window.
See image, arrow shows where Edit Window has "popped under" the Browser window.

### Expected Behavior

Browser window not covering up the Edit window

### Environment (please complete the following information)

```markdown
- OS: MacOS
- Browser: Safari
- Companion Version: 2.2.0+4447
```

Julusian:
Im going to call this an enhancement rather than a bug, as this is working as intended even if it isnt the best (I didnt want to put effort into rethinking the layout when editing triggers).

I think this is really part of a larger task of reworking various bits of the ui to be more consistent

MeestorX:
Fair enough. If I had more experience with the new Browser, I would have realized that clicking "Done" is the solution. I was clicking around trying to move windows and things and getting nowhere, so I figured other might have the same issue until they realize that you need to click "Done" to move that window out of the way. Especially when you can't see that there's a hidden window behind the browser.

MeestorX:
Do LOVE that browser, though!! :-)

Julusian:
This is fixed in v3.0, the trigger editor is no longer a modal so this no longer happens.
See #2165

Issue #1908, opened 2022-02-13
Trigger on variable changed


BryanCrotaz:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

There is an existing trigger for when a variable becomes equal to a value.

There is a missing case, where you want to trigger on a variable change, regardless of the value.

New trigger, Variable Change, which just takes the name of a variable, or an expression made of lots of variables

### Usecases

Game scoring. Hold the score in a variable with buttons that increment/decrement it

Whenever the score changes, use an action via the Variable Changed trigger to send the new score to the display system (e.g. OBS) as a text update.

WIthout this feature you need to add the text update action everywhere that might change the score.

Julusian:
I think that this needs some thought on how best to implement.
I would prefer for it to be done as part of the 'feedbacks' triggers system, as I think it will be useful to be able to stack conditions (eg, you dont want it to run when you are playing out a clip), but I dont know how that will work. Perhaps it needs to be a 'pulse'?

This might be appropriate to move to https://github.com/bitfocus/companion-module-bitfocus-companion, depending on the concluded plan

BryanCrotaz:
Maybe all triggers should have a list of ALL and NONE conditions. Only if all the ALLs and none of the NONEs are true does the trigger happen. 

Should the trigger happen later?

Should triggers be on events of on state changes?

BryanCrotaz:
A 'trigger' is an event. A trigger pull.

I think there are three types of event:

1. A true event: button press, button release, change of state, startup etc
2. A state entry
3. A state exit

Each trigger can have a Condition which is 
<expression><operator><expression>
OR
List of conditions <ORed>
OR
List of conditions <ANDed>

This allows you to build any logic you need

A trigger in the UI could have multiple triggers with the same condition list

BryanCrotaz:
Example

I want to set up device 1 with respect to device 2

Trigger on:

- Device 1 enters online state
- Device 2 enters online state
- Device 2 feedback x changes

Conditions:
- AND
-- Device1 state = online
-- Device2 state = online

Action:
Device1 set my_state = device2 variable

gerbenvandekraats:
+1 for this trigger on variable changed 👍

eric:
[...]

Issue #1389, opened 2021-01-25
Combined Time Entries Per Button for Schedules/Triggers


sydneyjd:
**Describe the feature**
 Being able to, for example, have a certain button trigger at multiple different time intervals per given day, being able to change specific time entries depending on the day, and being able to configure that all as one trigger, vs having to create a trigger for each specific time instance and specific day needed where the trigger needs to be a different time per each day, etc.

**Usecases**
 We have some devices that get toggled on and off over the network, and the Scheduler does the job currently for automating based on time, but having to add a specific time for both an on and off state(or multiple on/offs) per specific day gets rather unwieldy. It would be handy to be able to specify a custom time schedule per wanted days under one specific button trigger instance. The Variable Value option seems to be partially there, but if I want to add more than one condition, there appears no way to make that condition an OR instead of an AND, or do different variations and multiples of that.

Julusian:
This will be possible in v3.0

PR #2165, opened 2022-11-05
feat: overhaul triggers system

Julusian:
During the initial overhaul of code for v3.0, the triggers system was neglected and was changed as little as possible. This meant that it works in its own way, with a conflicting style to the rest of the application.

With this change, it is more consistent in behaviour and feel with buttons, and has some improved functionality that became easier to add by reusing existing code or concepts.

To begin with, the editing experience now matches that of buttons. The triggers tab has been split to have a side panel for the editing to be done in, and changes are now updated live, rather than requiring 'save' to be pressed (#2040 & #2155)
image

As discussed in #1908, the use of feedbacks as the trigger 'event' has been split out to be a dedicated 'condition', that gets checked for each execution of the trigger. This allows it to be used in conjunction with other event sources, such as timeofday or interval. 
The 'events' section has been reworked to allow for multiple event sources for the trigger. (#1389) This allows for some more complex scenarios of when to run the trigger.

Additionally, with 'condition' and 'event' being separate now, it makes sense to have the 'on variable changed' event type. Which will cause the trigger to be run whenever the selected variable changes. (#1908)

------

Internally, triggers have been rewritten to be a type of 'control'. This makes using the same ui, and action/feedback logic much easier, as they share a lot of the same code with normal buttons.

Triggers running on an interval now share timers, which means that they will execute more in sync with each other. The same timer is used for timeofday triggers, which has been reworked to ensure that if time jumps (something caused the clock to skip a few seconds or longer), then the first tick after the target time will execute those triggers instead.  
This is only being ticked at the resolution of 1s, and this may want to be run at a higher resolution to minimise the risk of skipping whole seconds.


krocheck:
Those updates all helped my notes.

PR #2187, opened 2022-11-18
feat: combined stepped and normal button types

Julusian:
This needs some UI work, as the nesting of scrollbars causes some weirdness, but the nested tabs are essential to avoid the panel becoming a stupidly long list once there are multiple steps.

This started out as a combination of making rotary actions work on stepped buttons, and giving stepped buttons up actions. I think both of these are important things to support, and will allow for a single encoder to be used as a RGB colour wheel, by pressing to cycle which colour you will adjust, among many other uses.
The further I got into this, I realised that normal buttons behaved exactly like a stepped button with only one step, so why not expose them like that in the ui? This will give users the freedom to convert seamlessly between the types.

The downside is that the ui is feeling more cluttered, and some of it could be rather confusing. Hopefully we can tidy up the ui sufficiently, perhaps hiding some bits in certain scenarios

A 'normal' button now looks like this:
image

Key differences are that the actions and feedbacks are now in tabs, with a tab used to 'add step'.  
Buttons can then be made into stepped buttons as simply as adding a second step.

Each step gets its own down, up, and rotate action sets/groups, as well as long press groups.

The checkbox to 'auto progress step' now causes the step to progress on release. This can of course be disabled and the internal actions to control the current step can be used instead to control the current step.


Julusian:
I have a solution for this that I am pretty happy with now.  
Looking pretty similar to before, with no nested scrollbar.
image

Once you cross point of the new tabs, they stick to the top of the scroll region, and the rest keeps on scrolling
image

Clicking on one of the tabs, will scroll it so the tabs are at the top. And it makes sure that if there isnt enough content, then blank space is added so that the tabs can stick to the top:
image

PR #2204, opened 2022-11-26
feat: make surface ids unique

Julusian:
Connecting a streamdeck locally provides an id of `20DL0123456`, but if connected via satellite this becomes `satellite-20DL0123456`.

This results in some inconsistent behaviour, as plugging a streamdeck into different satellite machines will give it the same settings, but plugging it in directly will get different settings.

This changes the ids to always look like `streamdeck:20DL0123456` (or similar for other surface types).

This could break surface config for existing surfaces, but it is doing a crude attempt to fix up some ids. I expect there to be some friction for users here, partially as we need to either not change satellite or coordinate it.
It is attempting to gracefully handle old actions which use surface ids, trying to reliably fix them will prove complex

### DIFF
--- a/lib/Data/Metrics.js
+++ b/lib/Data/Metrics.js
@@ -37,7 +37,7 @@ class DataMetrics extends CoreBase {
 
 		try {
 			Object.values(devices).forEach((device) => {
-				if (device.id !== undefined && device.id !== 'emulator') {
+				if (device.id !== undefined && !device.id.startsWith('emulator:')) {
 					// remove leading "satellite-" from satellite device serial numbers.
 					const serialNumber = device.id.replace('satellite-', '')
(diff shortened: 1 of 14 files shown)


krocheck:
Could we attempt a best effort regex match for a surface id if no colon is in the action param to the table so as to not break the existing uses.  Maybe something similar if there’s a ‘satellite:xxxx’ to extract the serial and regex match to the online surfaces.

Julusian:
@krocheck oh yeah that would be pretty easy to implement actually.
something like:
1) try for exact id
3) try for `streamdeck:${id}`
3) try for `loupedeck:${id}`
2) try for `satellite-${id}`
I would say perhaps that could be `endsWith(id)`, but I dont think we need it that loose. We are going to have to keep this in place for a while so as to not break configs, 

The ui will make it look like these actions are broken, but thats probably a good thing as it should encourage users to set them to the new format

PR #2216, opened 2022-12-04
Update package.json

fbosman:
[companion-module-axis-ptz](https://github.com/bitfocus/companion-module-axis-ptz) version 2.0.0:
new autofocus commands
new feedbacks on limits and autofocus
new variables on limits and status


krocheck:
@fbosman please use the #module-development channel in slack for these requests.

Issue #2220, opened 2022-11-20
Internal: hex to dec conversion AND Internal: dec to hex conversion


joshnortheimer:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

An internal function that coverts a customer variable's value from hex/dec and back.

### Usecases

retreiving data via http GET and storing value, but it's hex and i need dec to interact in companion. want to then modify the value and http put it back.

Julusian:
This is possible in the 3.0 betas using https://github.com/bitfocus/companion/pull/2253.

PR #2253, opened 2022-12-23
feat: enhanced expression functionality

### LINKED ISSUE #2256: Ability to do math operation with time / Define a variable as a "TIME" value.

### Describe the feature

I would like some ability to define a variable as a time value and do math operations with it.  I am trying to set up a system to set trigger times via a UI on the buttons with actions.  This way I can adjust all of our automations via the Streamdeck itself without digging into the GUI.  I'm almost ther but its very convoluted dealing with MM >60 and HH etc.  This would be a huge help if I could define an event time then set a Time - Xmin for event Y etc with a trigger for something to happen with internal time = the trigger time.

### Usecases

_No response_

Julusian:
Resolves #2256 and #2220

This replaces the use of `@estilles/expression-parser` with `jsep`, which is able to parse much more complex expressions and supports plugins for additional syntaxes.

`jsep` doesn't include a way to execute/resolve the expressions, but it is simple to implement an executor from the AST it generates.

There are a couple of breaking changes in behaviour. The `^` operator has changed to match js behaviour. And some operator precedence have changed to match the js behaviour. 

In addition, it has gained a set of boolean logic operators, and the ability to parse functions and provide implementations.  
The functions are pretty barebones for now, more can be added as needed. There are a bunch of internal actions to do various operations and save into a custom-variable, which should be converted to this syntax.

For example, the following are now possible:
* `$(internal:time_s) + " seconds"`
* `round($(internal:time_s) * 3.101)`
* `substr($(internal:time), 4, 7)` will be possible once an implementation of `substr` is done.

In addition, I have made it be possible to use an expression for button text, which will allow for much more complex button text scenarios.

image
image

Using a feedback to have a slow flash (1s on, 1s off):
image


robertskiba:
Sorry for asking a few years later, but I still can't find a way to convert a decimal to a hex value in companion. Is this correct? or is there any way?

thedist:
> Sorry for asking a few years later, but I still can't find a way to convert a decimal to a hex value in companion. Is this correct? or is there any way?

You can use the `toRadix` Expression Function, which has been available for years. For example `toRadix(123, 16)` will give `7b`, which is the denary value 123 in base 16 (ie, hex).

All Expression Functions can be found on the Getting Started page in the Companion Web UI.

robertskiba:
I take my hat off to this ingenious solution, which I would never have thought of myself! You never stop learning.

PR #2258, opened 2023-01-01
feat: Improve module list search filtering

k-yle:
A lot of people give up when they search for a product but can't find it in the list. Searching for things like `black magic`, or `artnet` currently returns no results, which is unexpected. 

This PR improves the search filtering so that it matches items, even if spaces or punctuation isn't quite right. 

Also, should this PR target the `beta`, `develop`, or `master` branch? I couldn't find any guidelines

### DIFF
--- a/webui/package.json
+++ b/webui/package.json
@@ -17,6 +17,7 @@
     "classnames": "^2.3.2",
     "dayjs": "^1.11.6",
     "fast-json-patch": "^3.1.1",
+    "fuse.js": "^6.6.2",
     "html-entities": "^2.3.3",
     "http-proxy-middleware": "^2.0.6",
     "intersection-observer": "^0.12.2",
--- a/webui/package.json
+++ b/webui/package.json
@@ -17,7 +17,7 @@
     "classnames": "^2.3.2",
     "dayjs": "^1.11.6",
     "fast-json-patch": "^3.1.1",
-    "fuse.js": "^6.6.2",
+    "fuzzysort": "^2.0.4",
     "html-entities": "^2.3.3",
     "http-proxy-middleware": "^2.0.6",
     "intersection-observer": "^0.12.2",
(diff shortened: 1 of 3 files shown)


Julusian:
Beta is the correct target.

Thanks. I think this is a good idea to do. Perhaps we should go further and use a library for fuzzy searching (such as https://fusejs.io/demo.html) to make it even better at searching?

leeoniya:
some other fuzzy search options:

https://github.com/leeoniya/uFuzzy#a-biased-appraisal-of-similar-work

k-yle:
sure, I've pushed a commit which uses `fuse.js`. I also tried `match-sorter` but that finds nothing with one of the original examples (an erroneous space in the word: `black magic`), and `uFuzzy` finds nothing when a hyphen is missing (e.g. `artnet` instead of `Art-Net`)

leeoniya:
> and` uFuzzy` finds nothing when a hyphen is missing (e.g. `artnet` instead of `Art-Net`)

with default settings it won't. you can set adjust the settings to make it work:

```
intraChars: '.',   // allow any char to be inserted
intraIns: 1,       // at most once
```

Julusian:
looking at the demo: https://leeoniya.github.io/uFuzzy/demos/compare.html?libs=uFuzzy,fuzzysort&search=mario%20adventures&intraChars=.&intraMode=1&intraIns=1

I am liking the look of fuzzysort. perhaps with a threshold of -100000 to filter out some junk

k-yle:
I've switched it to `fuzzysort` which seems to work just as well

leeoniya:
> perhaps with a threshold of -100000 to filter out some junk

this is precisely the problem uFuzzy tries to address :). if you try that cutoff for "magic", you'll get an incredible amount of garbage again:

image

if you toggle `outOfOrder`, uFuzzy will get you much better results:

https://leeoniya.github.io/uFuzzy/demos/compare.html?libs=uFuzzy,fuzzysort&search=mario%20adventures&intraChars=.&intraMode=1&intraIns=1&outOfOrder

image

PR #2329, opened 2023-02-14
feat: replace sharp

Julusian:
The aim was to remove the large dependency, but it ended up also being a performance improvement. Not a big enough speed increase to be noticable, but its nice to know its there.

I started off intending to only replace the scaling from sharp. But after giving it a test, the `rotateBuffer` implementation is really slow compared to doing it in this library, so makes sense to replace too.

`@julusian/image-rs` is a minimal wrapper around https://github.com/image-rs/image. I chose this as the base as I wanted something compiled and with minimal complexity for producing builds. Currently it is fairly minimal in api, intended to provide only what we need. This could be extended to some image encoding/decoding, or more colour formats if wanted in the future.

```
companion rotate x 8,538 ops/sec ±4.65% (69 runs sampled)
@julusian/image-rs rotation x 34,966 ops/sec ±10.80% (71 runs sampled)

@julusian/image-rs scaling x 864 ops/sec ±4.55% (70 runs sampled)
@julusian/image-rs scaling async x 909 ops/sec ±2.09% (75 runs sampled)
sharp x 498 ops/sec ±1.79% (81 runs sampled)
```


Julusian:
I still think this is a good idea, so I'm going to merge this to develop where it can get a bit more testing, and be part of 3.1

PR #2404, opened 2023-03-31
feat: Set current step  if variable meets condition

riccefarias:
Button: Set current step  if variable meets condition


Julusian:
Thanks, Im also going to add a `bank_current_step_if_expression` version of this, to be able to use the expression syntax too

PR #2416, opened 2023-04-11
feat: Add fontsize to Satellite API

elianterelle:
This PR adds Fontsize as optional streamed data to the Sattelite API.
I added a new Flag `FONTSIZE` for the `ADD-DEVICE` command, similar to the already existing `TEXT`, `COLOR`, ...
if this flag is set to true, the fontsize is passed on every `KEY-STATE` message.

If this gets merged, the wiki would need to be updated at two places:
- Messages to send -> Adding a satellite device -> Optional parameters:
`FONTSIZE` - true/false whether you want to be streamed button font size for display on the buttons (default false) 

- Messages to receive -> State change for key:
`FONTSIZE` Font size as an integer for the key. This is only sent for devices which were added where `FONTSIZE` was true

### DIFF
--- a/lib/Service/Satellite.js
+++ b/lib/Service/Satellite.js
@@ -8,7 +8,7 @@ import net from 'net'
  * 1.1.0 - Add KEY-STATE TYPE and PRESSED properties
  * 1.2.0 - Add DEVICEID to any ERROR responses when known
  * 1.3.0 - Add KEY-ROTATE message
- * 1.4.0 - Add ADD-DEVICE FONTSIZE flag and KEY-STATE FONTSIZE property
+ * 1.4.0 - Add TEXT_STYLE as ADD-DEVICE flags with FONT_SIZE as KEY-STATE property
  */
 const API_VERSION = '1.4.0'
 
@@ -107,7 +107,7 @@ class ServiceSatellite extends ServiceBase {
 		const streamBitmaps = params.BITMAPS === undefined || !isFalsey(params.BITMAPS)
 		const streamColors = params.COLORS !== undefined && !isFalsey(params.COLORS)
 		const streamText = params.TEXT !== undefined && !isFalsey(params.TEXT)
-		const streamFontSize = params.FONTSIZE !== undefined && !isFalsey(params.FONTSIZE)
+		const streamTextStyle = params.TEXT_STYLE !== undefined && !isFalsey(params.TEXT_STYLE)
 
 		const device = this.surfaces.addSatelliteDevice({
 			path: id,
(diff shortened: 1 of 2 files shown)


krocheck:
Can you as part of this PR, since I’m not aware of any other planned changes to this API, in the service class increment the version to 1.4.0 and document the new param.

elianterelle:
Done! Hope the force push was ok, don't have that much experience with pull requests but thought it would be nicer as a single commit and seems to have updated the pr correctly.

krocheck:
@elianterelle my quick review of the code seems to indicate the COLORS flag sends only bgcolor instead of both bgcolor and [font]color.  Is this accurate and something you would want to address as well?

Our workflow is squash and merge, so cleaning up like that isn’t really necessary.

elianterelle:
That's correct, for my usecase it doesn't matter as i'm using monochrome displays, but i could add this. Then the only missing parameters would be text alignment i think, so if i we add fontcolor, why not add that as well. I'd suggest FONTCOLOR and TEXTALIGN for the parameter names(I'm not a native english speaker but i think that's the correct distinction between Font and Text similar to how it's done in CSS for example)?

krocheck:
@Julusian do you have an opinion here?

Julusian:
I am not certain on this but I am wondering if this should be included as part of the `TEXT=1` option, or perhaps `TEXT-STYLE=1`? 
While both of these would cause us to send more data than may be necessary to clients, it limits the number of options we have at the time of adding the device

elianterelle:
The only issue i see in this is that it could potentially be a breaking change in the sense that the KEY-STATE message gets considerably longer, which is probably irrelevant with dynamically typed languages like javascript, but for example i use the satellite api on a microcontroller programmed in C, so my buffer for incoming messages has a fixed length. On the other side, the satellite api docs already states that more parameters can be added over time.

elianterelle:
That wouldn't be an issue with a new flag like the `TEXT-STYLE` you mentioned.

Julusian:
> Any update on this? Should i go ahead and implement `text-align` and `font-color` in addition to `font-size` with a combined flag `TEXT-STYLE`?

I think it would be good to get it under a `TEXT-STYLE` flag, but the other properties dont need to be defined yet. If theyre not needed, it would be better to leave them be for now, as there is a plan to overhaul the graphics generation which may change how some of these things will work

elianterelle:
Done. I called it `TEXT_STYLE` instead of `TEXT-STYLE` to match the other properties(`PRODUCT_NAME`, `KEYS_PER_ROW`,..) and also changed the property sent to the device from "FONTSIZE" to "FONT_SIZE". Let me know if there's anything else to do on this

Julusian:
Thanks!
I've also pulled this into beta too

Issue #1187, opened 2020-07-31
Expose properties instead of actions


Julusian:
**Describe the feature**
I have been thinking about this for a while, as exposing numeric properties in device modules gets really repetitive, especially when wanting to allow actions to do relative changes and comparison operators in feedbacks.

The current actions architecture works nicely for devices which work in actions, but puts a lot of work on the modules when the devices expose properties. This leads to the functionality and flow being inconsistent even within modules.
The idea of this is not to replace all actions, just some of them. Actions are appropriate for some things (eg performing a cut on an atem), but properties make sense for others (eg setting preview to input1)

As an example lets look at an audio fader.
For this, we want to be able to set it to an absolute value, and to do small changes to bump the level up and down. We also want feedback for that value using a few different comparison operators (<, <=, ==,  !=, =>, >)
In this example, that isnt much to do. But it is still work that is going to want to be replicated for each module which has an audio fader, and some will even have different types of faders and so will need to duplicate it themselves too.

A better approach to handle these would be for modules to instead report back the properties they have which can be set.
Each property could be defined as something like:
```
{
  id: 'ch1',
  label: 'Fader: Channel 1',
  type: 'number',
  min: '-90',
  max: '10',
  getValue: () => {...},
  setValue: () => {...},

  subscribe: () => {...},
  unsubscribe: () => {...},

  variables: [
    {
      id: 'ch1_db',
      label: 'Channel 1 level (db)',
    },{
      id: 'ch1_percent',
      label: 'Channel 1 level (%)',
    }
  ],
  getVariable: () => {...}
}
{
  id: 'ch1-type',
  label: 'Type: Channel 1',
  type: 'dropdown',
  options: [
    { label: 'Something', id: 0 },
    { label: 'Another', id: 1 }
  ],
  getValue: () => {...},
  setValue: () => {...},

  subscribe: () => {...},
  unsubscribe: () => {...},

  variables: [],
  getVariable: () => {...}
}
```
And to go with this, whenever the module has a change to a property it should inform core, which can then calculate where this is used
```
  this.notifyPropertyChange('ch1-type')
```

**Future**
This will serve as a good starting point for when we want to support input devices with faders or dials, this might even start as allowing osc to provide values. 
It could be used to create some more dynamic behaviour too, such as actions which sets the level of channel 1 to match channel 2.

MeestorX:
Can't say I completely understand what you've got in mind, but anything that allows for more flexibility I'm definitely up for.

[...]

PR #2431, opened 2023-04-29
feat: replace variables with properties #1187

Julusian:
This is 'Stage 1' of properties.  
The eventual aim of 'properties' is to avoid the need for modules to create variables, one or more actions, and a feedback for the same 'value' in a model of state. Instead, modules can expose a property, and we will handle generating the actions and feedbacks that are appropriate.

This replaces variables with properties, so far there is very little difference. The only benefit is that properties can have 'instances', in other words they can have multiple values.
So instead of having `atem:me_program_1` and `atem:me_program_2`, we have a single property `atem:me_program` with two 'instances': `atem:me_program:1` and `atem:me_program:2`. A small difference, but it allows for grouping them better.

Properties are also required to provide some info on their type, which will let us know how to interpret them.

These changes will become more meaningful later once:
* properties can be settable
* we auto-generate feedbacks
* we auto-generate actions if settable
* subscriptions for values

For compatability, this is an extension to the module api with the old variables parts translated across. 

### TODO:
- [ ] upgrade 'rules' for 'variables'?
- [ ] subscriptions?


lib/Graphics/Controller.js
```diff
@@ -137,7 +137,10 @@ class GraphicsController extends CoreBase {
 		const variableValue = buttonStyle?.text
 		setImmediate(() => {
 			this.instance.variable.setVariableValues('internal', {
-				[`b_text_${page}_${bank}`]: variableValue,
+				[`b_text_${page}_${bank}`]: {
+					isSingleValue: true,
+					value: variableValue,
+				},
```
Julusian, line 143: I am not keen on this object structure it is using, but as some variables are objects (and might remain that way, for use with jsonpath?), I needed a way to differentiate how to treat them.

Julusian:
I still want to do this, but haven't had the motivation to finish this so it is rather stale and full of conflicts.
I'll reopen whenever I want to pick it up again

PR #2432, opened 2023-05-03
Fix the Elegato plugin reset to default

peternewman:
Just spotted by observation when working on the code. Entirely untested...


Julusian:
good spot :)

krocheck:
Thank you @peternewman

PR #2436, opened 2023-05-04
Add initial support for Contour Shuttle

### LINKED ISSUE #2430: Add support for Contour Shuttle

### Describe the feature

Add support for https://contour-design.co.uk/collections/multimedia-controller/products/multimedia-controller-pro and similar
https://contour-design.co.uk/collections/multimedia-controller/products/multimedia-controller-xpress

I assume you'd accept a PR, looking at xkeys it doesn't look too hard... (famous last words).

### Usecases

Let people with Shuttles use Companion.

peternewman:
Closes #2430 

Still todo:
- [x] Decide the best way to map the devices (see e.g. https://contour-design.co.uk/collections/multimedia-controller/products/multimedia-controller-pro ) there are rows of four, five, three and maybe two buttons depending on how you count them. Should I do each row on a different button row in Companion? I don't really know the pros/cons of this? With/without the encoders embedded in the middle?
- [x] Naming of the module/code - I know it as Contour Shuttle for all of them, they've moved to different names now in the UK versus US ( https://contour-design.co.uk/collections/multimedia-controller/products/multimedia-controller-pro or http://web.archive.org/web/20221020020659/https://contourdesign.fi/collections/multimedia-controller/products/multimedia-controller-pro vs http://web.archive.org/web/20230326122038/https://contourdesign.com/products/shuttle-pro-v2 ). Do we name the module Contour, Shuttle or ContourShuttle?
- [ ] General jog behaviour, how should we integrate it? Should jog and shuttle also do the XKeys behaviour with variables too for anything that uses them, or just the encoder option where supported? - Currently both just generate left/right rotate on different buttons.
- [x] Add a picture of the example config
- [ ] Add example configs for all the different models?
- [x] Guess the mapping for the Xpress and Pro v1 models

~Also what's the best way to test the encoders to see if I've implemented it correctly? Just assigning a rotary action to fire another button? - I couldn't see this on the web UI, but making them set the background colour of the button seemed to work.~ I've roughed up control of my TV's volume via https://github.com/bitfocus/companion-module-samsung-smarttv/ which has been a good test.


lib/Surface/USB/Shuttle.js
```diff
@@ -0,0 +1,202 @@
...
+	static async create(ipcWrapper, devicepath) {
+		const shuttle = shuttleControlUSB
+		shuttle.start()
+		try {
+			let deviceInfo = null
+			let info = null
+			console.log('Device count', shuttle.getDeviceList().length)
+			if (shuttle.getDeviceList().length >= 1) {
+				// TODO(Peter): Fix this to fetch the correct device
+				deviceInfo = shuttle.getDeviceList()[0]
```
peternewman: https://github.com/hopejr/ShuttleControlUSB/pull/5 will fix this to be able to get the correct device.

I'm still worried there could be a race condition though if it hasn't discovered it after shuttle.start() (especially with the delay of 1s in the library on Mac). As @Julusian mentioned something that just explicitly accesses a particular device by path would probably be better.

Julusian: This won't work on macos. macos is different to windows and linux in that HID devices can only be opened in one place at a time.
Even doing two opens of a device from the same process/thread does not work.
Screenshot from 2023-05-07 14-22-50


In the current implementation here, when two devices are connected a different child_process will be spawned for each. Both of those will be running the usb watcher and both will attempt to open the hid devices. Only one will succeed with the opening, and there is no guarantee it will be the one we want it to be.

-----

This either needs to not run the usb hotplug monitoring in each child_process and instead explicitly open a specified path, or it should be running all of the shuttles in a single child_process, and ignore them in the main companion hotplug detection

peternewman: > This won't work on macos. macos is different to windows and linux in that HID devices can only be opened in one place at a time. Even doing two opens of a device from the same process/thread does not work.

Out of interest, is that a core Mac limitation, rather than a node-hid or hidapi specific?

> In the current implementation here, when two devices are connected a different child_process will be spawned for each. Both of those will be running the usb watcher and both will attempt to open the hid devices. Only one will succeed with the opening, and there is no guarantee it will be the one we want it to be.

I assume that just calling `node-hid.devices()` is enough to cause this issue?

> This either needs to not run the usb hotplug monitoring in each child_process and instead explicitly open a specified path, or it should be running all of the shuttles in a single child_process, and ignore them in the main companion hotplug detection

I'm assuming from https://github.com/bitfocus/companion/issues/2430#issuecomment-1528896873 that your preference is the former? I'll see if I can update the package to do that if you're happy with that too @hopejr ?

Julusian: > Out of interest, is that a core Mac limitation, rather than a node-hid or hidapi specific?

Oh, maybe it can be turned off https://github.com/libusb/hidapi/pull/397, but it isnt exposed in `node-hid` currently. I suppose looking into that is an option, but it still feels rather wasteful to be opening them all multiple times.

> I assume that just calling node-hid.devices() is enough to cause this issue?

Its the `new hid.HID(device.path)` call that will be problematic. doing a `hid.devices()` is perfectly fine (in this context)

> I'm assuming from https://github.com/bitfocus/companion/issues/2430#issuecomment-1528896873 that your preference is the former?

Yeah, that would be prefered. It will avoid adding a lot of complexity to companion to handle this, at what I hope is a small cost to the library.  
In develop (3.x), this child_process stuff is being removed, so it might not be as bad, but we still have the same system for scanning for devices.

hopejr: @peternewman if you can come up with a way that's not going to break backwards compatibility, then I have no problem with it. Though does getRawHidDevice() not do what you need? We could always change it to accept path or id (the find function can do `ele => ele.id === id || ele.path === id`).

peternewman: Thanks @hopejr . 

How does this look?:
https://github.com/hopejr/ShuttleControlUSB/pull/8/files

Unfortunately `getRawHidDevice()` on it's own is no good, because all the `new hid.HID(device.path)` @Julusian refers to still get called.

peternewman: We're now not scanning with @hopejr 's library so hopefully this is sorted now.

peternewman:
@Julusian or others I think this is now ready for at least some high level review if possibly not code review until I've had opinions/answers on my queries and implemented them as appropriate.

lib/Data/UserConfig.js
```diff
@@ -42,6 +42,7 @@ class DataUserConfig extends CoreBase {
 		elgato_plugin_enable: false, // Also disables local streamdeck
 		usb_hotplug: true,
 		loupedeck_enable: false,
+		shuttle_enable: true,
```
Julusian: This should default to `false` to be consistent with the other non-streamdeck devices

docs/7_surfaces/images/contour-shuttle.png
Julusian, line 1: I think this layout makes sense

peternewman:
Polite bump if you've got a chance @Julusian or do you have a plan for this to go into a later milestone?

lib/Surface/USB/Shuttle.js
```diff
@@ -0,0 +1,221 @@
...
+		// Should never be fired
+	}
+
+	clearDeck() {
+		this.ipcWrapper.log('debug', 'shuttle.clearDeck()')
+
+		this.shuttle.blankDevice(true, true).catch((e) => {
+			this.ipcWrapper.log('debug', `blank failed: ${e}`)
+		})
+	}
```
Julusian:
```suggestion
		// Not relevant for this device
	}
```
This method is intended to set all the displays/buttons of a surface to black, not relevant for these surfaces

lib/Surface/USB/Shuttle.js
```diff
@@ -0,0 +1,221 @@
...
+			this.ipcWrapper.remove()
+		})
+	}
+
+	async #init() {
+		this.ipcWrapper.log('debug', `Contour Shuttle ${this.deviceInfo.name} detected`)
+
+		// Make sure the first clear happens properly
+		// TODO(Peter): Is this required?
+		//await this.loupedeck.blankDevice(true, true)
```
Julusian: This too can be deleted
```suggestion
```

lib/Surface/USB/Shuttle.js
```diff
@@ -0,0 +1,221 @@
...
+			this.ipcWrapper.rotate(keyIndex, delta == 1)
+		})
+
+		this.shuttle.on('shuttle-trans', (previous, current) => {
+			const keyIndex = rotaryToButtonIndex(this.modelInfo, encoders.SHUTTLE)
+			if (keyIndex === undefined) {
+				return
+			}
+
+			this.ipcWrapper.rotate(keyIndex, previous < current)
```
Julusian: Im not sure about the strategy here, but maybe this will do for now?

Julusian:
> Polite bump if you've got a chance @Julusian or do you have a plan for this to go into a later milestone?

I think this is going to land in 3.1. 3.0 is really close to being released and I think its a bit late to squeeze this in there unfortonately

> Naming of the module/code - I know it as Contour Shuttle for all of them

I think that name makes sense, people will have known them for years by their old name. If it does get confusing it can be renamed later on.

>  General jog behaviour, how should we integrate it? Should jog and shuttle also do the XKeys behaviour with variables too for anything that uses them, or just the encoder option where supported? - Currently both just generate left/right rotate on different buttons.

I think the variable would be a good thing to do initially, so that it can be used anywhere the xkeys already can be.  
I'm not sure that left/right events will be useful currently, I think it has too much risk of getting stuck in a bad/unexpected state. And its behaviour could be replicated with some triggers. But I'm not opposed to leaving it in if you think it will be useful

peternewman:
> > Polite bump if you've got a chance @Julusian or do you have a plan for this to go into a later milestone?
> 
> I think this is going to land in 3.1. 3.0 is really close to being released and I think its a bit late to squeeze this in there unfortonately

No worries, it's not too urgent for me. Do I need to do anything different (change branch etc) or just wait patiently for it to be merged? From your comments it looks like there are only a few minor things to be fixed?

> > Naming of the module/code - I know it as Contour Shuttle for all of them
> 
> I think that name makes sense, people will have known them for years by their old name. If it does get confusing it can be renamed later on.

👍 

> > General jog behaviour, how should we integrate it? Should jog and shuttle also do the XKeys behaviour with variables too for anything that uses them, or just the encoder option where supported? - Currently both just generate left/right rotate on different buttons.
> 
> I think the variable would be a good thing to do initially, so that it can be used anywhere the xkeys already can be. I'm not sure that left/right events will be useful currently, I think it has too much risk of getting stuck in a bad/unexpected state. And its behaviour could be replicated with some triggers. But I'm not opposed to leaving it in if you think it will be useful

Do you know off-hand which xkeys supporting modules might be easy to test with offline? I've not really played with that bit yet, I sort of just assumed it would fire the same as the shuttle wheel.

lib/Surface/USB/Shuttle.js
peternewman: I don't think I necessarily found a better explanation of what I should be doing with the rotate IPC otherwise?

lib/Data/UserConfig.js
peternewman:
```suggestion
		shuttle_enable: false,
```

peternewman: Actually I think I realised I left it like that as the xkeys was true...

webui/src/Wizard/FinishStep.jsx
```diff
@@ -29,7 +29,8 @@ export function FinishStep({ oldConfig, newConfig }) {
 			</ol>
 			{((newConfig.elgato_plugin_enable && oldConfig.elgato_plugin_enable !== newConfig.elgato_plugin_enable) ||
 				(!newConfig.xkeys_enable && oldConfig.xkeys_enable !== newConfig.xkeys_enable) ||
-				(!newConfig.loupedeck_enable && oldConfig.loupedeck_enable !== newConfig.loupedeck_enable)) && (
+				(!newConfig.loupedeck_enable && oldConfig.loupedeck_enable !== newConfig.loupedeck_enable) ||
+				(!newConfig.shuttle_enable && oldConfig.shuttle_enable !== newConfig.shuttle_enable)) && (
```
peternewman: Actually, for consistent naming shouldn't this be like so?:
```suggestion
				(!newConfig.contour_enable && oldConfig.contour_enable !== newConfig.contour_enable)) && (
```

lib/Surface/USB/Shuttle.js
peternewman, line 1: And shouldn't this be called `lib/Surface/USB/Contour.js` ?

Julusian:
@peternewman with 3.0 released, it would be nice to try and get this finished up and merged into 3.1

docs/7_surfaces/contour_shuttle.md
```diff
@@ -0,0 +1,8 @@
+It is possible to use the Contour ShuttleXpress and ShuttlePro v1 and v2 with Companion since v3.0.0.
```
peternewman:
```suggestion
It is possible to use the Contour ShuttleXpress and ShuttlePro v1 and v2 with Companion since v3.1.0.
```

docs/7_surfaces/contour_shuttle.md
```diff
@@ -0,0 +1,8 @@
...
+
+Enable support for it in the Companion settings, and rescan for USB devices.
+
+The layout pretty closely matches what you should expect based on the device.
```
peternewman:
```suggestion
The layout pretty closely matches what you would expect based on the device.
```

lib/Surface/USB/Shuttle.js
```diff
@@ -0,0 +1,213 @@
...
+					info = shuttleProV1Info
+					break
+				case 'ShuttlePro v2':
+					info = shuttleProV2Info
+					break
+				default:
+					this.ipcWrapper.log('debug', `Unknown Contour Shuttle device detected: ${this.deviceInfo.name}`)
+			}
+			if (!info) {
+				throw new Error('Unsupported model ')
```
peternewman:
```suggestion
				throw new Error('Unsupported model')
```

lib/Surface/USB/Shuttle.js
```diff
@@ -0,0 +1,213 @@
...
+	setConfig(config, force) {
+		// No config currently present
+		this.config = config
+	}
+
+	quit() {
+		try {
+			this.clearDeck()
+		} catch (e) {}
+
```
peternewman: Should I just drop this then given clearDeck doesn't exist/do anything for this device?
```suggestion
```

peternewman:
> @peternewman with 3.0 released, it would be nice to try and get this finished up and merged into 3.1

Yeah that would be good, so I think the follow is outstanding which mostly needs input from you or someone else please:

---
> > General jog behaviour, how should we integrate it? Should jog and shuttle also do the XKeys behaviour with variables too for anything that uses them, or just the encoder option where supported? - Currently both just generate left/right rotate on different buttons.
> 
> I think the variable would be a good thing to do initially, so that it can be used anywhere the xkeys already can be. I'm not sure that left/right events will be useful currently, I think it has too much risk of getting stuck in a bad/unexpected state. And its behaviour could be replicated with some triggers. But I'm not opposed to leaving it in if you think it will be useful

Also:
https://github.com/bitfocus/companion/pull/2436/files/1624062d4f0e716d154c082f54acc5b0b5a40712#diff-6a966e7bb48dd10a0fade6c29ca6b3f3600fdf9bced23b5692e47918ff85e805

So if you wanted to simulate the same thing, can you map changes of a variable to trigger actions then?

In terms of getting stuck in unexpected states, when I had it controlling my TV volume, it stopped at the expected time and didn't keep firing if that's what you mean, although obviously it ramped the volume up as you shuttled in one direction, then back down to where it started as the spring brought it back to normal (so not very useful functionally, but it didn't keep making it louder or quieter when in the normal position).

Looking at something like a Blackmagic Hyperdeck, I'd naively assume I could attach the shuttle actions to rotating in either direction, and then use the xkeys variable (when I add it) to populate the speed value and it would then work like a normal shuttle remote.
https://github.com/bitfocus/companion-module-bmd-hyperdeck/blob/master/actions.js

---

Enabled (to match xkeys) or not by default:
https://github.com/bitfocus/companion/pull/2436/files/1624062d4f0e716d154c082f54acc5b0b5a40712#diff-cdb93ec8c3c824195a7b3275187885f4dd81575353b2370e51ce93e103bb288a

---

Naming; xkeys really throws things off, as P.I. Engineering is never used/mentioned ( https://xkeys.com/about.html ) but generally with the other devices there seems to be some randomness with whether the company (e.g. Loupdeck), or the product (e.g. Live), or both are used for files and variables and config etc. Do you want me to try and be consistent one way or the other for Contour Shuttle?

---

Also sort of related, I notice there's now a merge conflict, it looks like info on the Logitech support and the Streamdeck plugin has been removed (seemingly by mistake as it still seems to be supported from what I can see):
https://github.com/bitfocus/companion/pull/2436/conflicts

dnmeid:
Here's my two cents:
>> General jog behaviour, how should we integrate it? Should jog and shuttle also do the XKeys behaviour with variables too for anything that uses them, or just the encoder option where supported? - Currently both just generate left/right rotate on different buttons.

The xkeys variables implementation is a makeshift and we'll change that in the future. Also the rotary actions will get a change. Actually the variables are closer to what it will be in some way, but generally I think it is better to stick with only rotary actions for the moment. Later it will be possible to also have linear values and not only triggers and we can then upgrade.  
For the shuttle this would be a normal rotary. I don't know the internals of Contour, but at the end one rotary event shoud be fired each time the shuttle moved a certain amount of degrees.  
For the Jog it is a bit more complicated to implement, because you would need to constantly fire rotary events and the interval time depends on the amplitude of the jog.

> Enabled (to match xkeys) or not by default

not enabled

> Do you want me to try and be consistent one way or the other for Contour Shuttle?

Try to be descriptive. We want to find code easy. If this allows for some consistency even better, if not, don't worry.

> it looks like info on the Logitech support and the Streamdeck plugin has been removed

I can't tell either why this info has been removed, I also think it should still be there.

lib/Surface/USB/Shuttle.js
Julusian: yep, you can drop that

Julusian:
> So if you wanted to simulate the same thing, can you map changes of a variable to trigger actions then?

For the shuttle, yes I would like to see the same. With it as a variable, I can build a set of triggers to handle the shuttle hitting each value. eg:
image
I think that this being done as rotate events would be harder to handle.  

For jog, doing a rotate as you are now is good. The xkeys approach for jog is definitely a legacy thing, which has a better way of being done now.

> For the shuttle this would be a normal rotary. I don't know the internals of Contour, but at the end one rotary event shoud be fired each time the shuttle moved a certain amount of degrees.

I disagree with doing this for now. If I want to map these shuttle events to a hyperdeck which only has a 'set shuttle speed' action, that would mean doing a bunch of conditionals and state tracking with a custom variable to implement. and if the shuttle gets disconnected while rotated, then companion will need restarting as it wont be possible to recenter it.  

I expect that some modules (vmix?) can interpret the shuttle variable directly, 

> Enabled (to match xkeys) or not by default:

I think disabled. when supporting new surfaces like this, I am always afraid of users getting confused because things started happening that they werent expecting. I know that there will be some users out there using a shuttle directly in vmix on the same machine they are running companion. It would be better to not surprise them, especially on a surface which has no feedback to the user.

> Naming;

I suppose we are generally following some 'brand recognition' here. 
Everyone I know refers to them as xkeys, I didnt know of the company name until many years later.  
For loupedeck, we support multiple ranges of theirs now, and just the product/range name 'live' would be unhelpful.  
In that sense 'shuttle' might make the most sense, but as you also raised that the naming across regions is inconsistent I think it better to be verbose here and have it 'contour shuttle'

> Also sort of related, I notice there's now a merge conflict

The logitech support moved where it is in the ui, so has also moved in the documentation. The plugin support has been reworded to `Enable connected Streamdecks`

dnmeid:
>> For the shuttle this would be a normal rotary. I don't know the internals of Contour, but at the end one rotary event shoud be fired each time the shuttle moved a certain amount of degrees.

> I disagree with doing this for now.

Wait, I mixed up jog and shuttle, meant to be the other way. The endless rotating inner thing is the jog and this should be a rotary.
The outer ring with a dedicated center position is shuttle and here I would go for rotary events in a different frequency.
I see the hyperdeck problem, so it is ok if there is an additional variable. But users have to be beware that this functionality will change.

lib/Data/UserConfig.js
peternewman: Setting to false as discussed elsewhere.

lib/Surface/USB/ContourShuttle.js
```diff
@@ -0,0 +1,218 @@
...
+
+			this.ipcWrapper.rotate(keyIndex, delta == 1)
+		})
+
+		this.contourShuttle.on('jog', (value) => {
+			console.log(`Jog position has changed`, value)
+			this.ipcWrapper.xkeysSetVariable('jog', value)
+//			setTimeout(() => {
+//				this.ipcWrapper.xkeysSetVariable('jog', 0)
+//			}, 20)
```
peternewman: This was copied from the xkeys code, but I don't quite see the logic behind it. Do we actually want the delta here (which seems to be what the xkeys does), rather than the absolute?

lib/Surface/USB/ContourShuttle.js
```diff
@@ -0,0 +1,218 @@
...
+			if (keyIndex === undefined) {
+				return
+			}
+
+			this.ipcWrapper.rotate(keyIndex, delta == 1)
+		})
+
+		this.contourShuttle.on('jog', (value) => {
+			console.log(`Jog position has changed`, value)
+			this.ipcWrapper.xkeysSetVariable('jog', value)
```
peternewman: General note, do you want new IPC stuff for Contour Shuttle, or them renamed to just setVariable or...?

lib/Internal/Surface.js
```diff
@@ -130,11 +130,11 @@ export default class Surface extends CoreBase {
 				name: 't-bar',
 			},
 			{
-				label: 'XKeys: Shuttle position',
+				label: 'XKeys/Contour Shuttle: Shuttle position',
 				name: 'shuttle',
 			},
 			{
-				label: 'XKeys: Jog position',
+				label: 'XKeys/Contour Shuttle: Jog position',
```
peternewman, line 137: I assume you want combined rather than individual variables here, so they'll just work on the existing modules that support them?

lib/Surface/USB/ContourShuttle.js
```diff
@@ -0,0 +1,218 @@
...
+	const index = modelInfo.encoders[info - 1]
+	if (index !== undefined) {
+		return index
+	}
+
+	return undefined
+}
+
+class SurfaceUSBContourShuttle {
+	constructor(ipcWrapper, devicepath, contourShuttle, modelInfo, deviceInfo) {
```
peternewman: `devicepath` seems to be consistently not camelCase across all the device drivers (but is in the main lib/Surface/Controller.js), I don't know if you'd like this fixed?

peternewman (commented):
A few queries...

lib/Surface/USB/ContourShuttle.js
Julusian: camelcase would be better, but I dont really mind

peternewman:
> > For the shuttle this would be a normal rotary. I don't know the internals of Contour, but at the end one rotary event shoud be fired each time the shuttle moved a certain amount of degrees.
> 
> I disagree with doing this for now. If I want to map these shuttle events to a hyperdeck which only has a 'set shuttle speed' action, that would mean doing a bunch of conditionals and state tracking with a custom variable to implement. and if the shuttle gets disconnected while rotated, then companion will need restarting as it wont be possible to recenter it.

> Wait, I mixed up jog and shuttle, meant to be the other way. The endless rotating inner thing is the jog and this should be a rotary. The outer ring with a dedicated center position is shuttle and here I would go for rotary events in a different frequency. I see the hyperdeck problem, so it is ok if there is an additional variable. But users have to be beware that this functionality will change.

My gut feeling is it probably wants both options ideally, which I guess would need to be two actions (or some way to disable the repeat aspect). I think it would want to be two actions, or at least a per rotate action repeat rate option to handle the different types of behaviour.

Some stuff like the Hyperdeck will be clever enough to have a "speed" variable

Whilst for other things it might be that each press makes it run faster (e.g. FF 1x, 2x, 4x with each press incrementing that), so simple triggers will do that (just don't pass in the variable to the action).

So both of the above fire an action once for each increment of the shuttle wheel, either overwriting or repeat pressing to get to the right speed.

Finally if the tool doesn't have a dedicated shuttle type option, you might need to fake it by e.g. jumping forward 10s every second at the slowest rate, but as you wind the shuttle up, a multiplier needs to happen, i.e. it will need to repeat the key press/action once per second at the slowest rate, but seven times at the fastest.

lib/Internal/Surface.js
Julusian: Yes, combined would be better for now

lib/Surface/USB/ContourShuttle.js
Julusian: I personally don't see how an absolute value could be used without first converting it a delta (which is doable in a trigger, but would be annoying). But more importantly, I think it best to match the xkeys behaviour to make it a drop in replacement.

lib/Surface/USB/ContourShuttle.js
Julusian: It isn't worth adding a new method for. I suppose the current one could be renamed, but I am also fine with the name being left as is.

lib/Surface/USB/ContourShuttle.js
peternewman: Yeah fair point, updated.

lib/Surface/USB/ContourShuttle.js
peternewman: I've renamed it now.

lib/Surface/USB/ContourShuttle.js
peternewman: Updated, and serialnumber too.

Julusian:
I dont feel that strongly either way on a repeating thing. That could be a second variable which is pulsed between 0 and a value?

I think the other other thing left is to rename the config field to be `contour_shuttle_enable` or something.

peternewman:
> I dont feel that strongly either way on a repeating thing. That could be a second variable which is pulsed between 0 and a value?

Fair enough, I don't directly need it, so I guess it's a case of wait and see what others want. I kind of figured something in the GUI would be better as it would allow some sort of multiplier option, which a straight variable wouldn't. Plus as per the request in https://github.com/bitfocus/companion-module-bmd-hyperdeck/issues/96 you can't do that with one variable, but you could potentially do something clever for that in the UI.

> I think the other other thing left is to rename the config field to be `contour_shuttle_enable` or something.

Do you mean like this? :wink:
https://github.com/bitfocus/companion/pull/2436/commits/74e3c4d9b7bf98ef199c6769b939241fc241a847

Julusian:
@peternewman this is in the beta builds now!

Would you be able to give it a quick test in the experimental builds (develop branch)? 
There has been a large amount of reworking there, with a few bits impacting how these surfaces interact with companion. I think I made the necessary changes correctly, but there is a chance I missed something. Having confirmation that it still works would be useful

peternewman:
> Would you be able to give it a quick test in the experimental builds (develop branch)? There has been a large amount of reworking there, with a few bits impacting how these surfaces interact with companion. I think I made the necessary changes correctly, but there is a chance I missed something. Having confirmation that it still works would be useful

I tested on Windows (having developed on Linux.

Specifically I ran companion-win64-3.1.0+6124-beta-0f91aeff . It found it fine and I loaded my sample config and all buttons worked fine.

I then switched to companion-win64-3.99.0+6257-develop-abf9c43e, it found the Contour Shuttle fine, but it had a new ID for it, so had the old one disconnected. I've no idea what's causing that bug, but easy enough to work around.

With 3.99.0 with the Contour Shuttle enabled, the launch window doesn't close when I hit quit. If I run it from a command prompt, I get this (it's fine with Shuttle disabled):
```
2023-08-24T22:37:46.725Z info Registry somewhere, the system wants to exit. kthxbai
2023-08-24T22:37:46.727Z error Surface/Handler/contourshuttle:355eef0f4d0c714a778c2ab685d68c58 Contour Shuttle ShuttlePro v2 disconnected
```

With 3.1.0 the error was still thrown but it did exit:
```
2023-08-24T22:43:47.125Z info Registry somewhere, the system wants to exit. kthxbai
2023-08-24T22:43:47.125Z error Surface/Handler/contourshuttle:381cd4026cfffb3ad7b80a032c0773f7 Contour Shuttle ShuttlePro v2 disconnected
2023-08-24T22:43:47.189Z Application: Companion process stopped
```

StockwerkOrange:
Hi, i love the work and i try to use it for scrubbing in Timelines or fast forward/backward Video. Maybe you can give the posibility to aktivate the behavior from jog and shuffel wheel like rotary left right to make it easy to programm them for scrubbing or forward/backward or to do next or previous for cue's and stuff?
Thannks for your work!

PR #2451, opened 2023-05-14
feat: videohub panel server

Julusian:
Replaces https://github.com/bitfocus/companion-module-bmd-videohub-controlpanel

This provides a minimal videohub 'server' implementation, to allow control panels to be connectde and be used as a surface.

Because of the number of buttons on the panels, the last 8 buttons take up a second row on the following page. Once we can make the grid larger, this can be fixed up.
Should this instead split it so each row is on a separate page? Perhaps only occupying a space of 5x4 on each of those pages?

This has been tested with a single panel, there is no reason it shouldn't work just fine with more.

When a panel connects, its settings are modified (requiring the ability for companion to connect back to the panels on port 9991) to be a 40 source, 1 output setup with sequential button order that we need.

The leds on the buttons do not light up currently, this will be hard to do well, so has been omitted for now.

### DIFF
--- a/lib/Data/UserConfig.js
+++ b/lib/Data/UserConfig.js
@@ -61,6 +61,8 @@ class DataUserConfig extends CoreBase {
 
 		emberplus_enabled: false,
 
+		videohub_panel_enabled: false,
+
 		artnet_enabled: false,
 		artnet_universe: 1,
 		artnet_channel: 1,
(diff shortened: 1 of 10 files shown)


krocheck:
I never looked into that module, though I knew it existed.  My thought is that you would use "destination" as the page number and "source" as the bank number.  That would essentially let the user program the pages they want to switch between and executing routes triggers a press/release on a timed structure (similar to how the Rosstalk server works).  That would also allow the use of the panel's macro functionality to punch multiple buttons on a single press.

Although if raw access works and we treat it like a dumb 40 button panel, that would work fine.

willosof:
awesome!

Julusian:
> @krocheck My thought is that you would use "destination" as the page number and "source" as the bank number.

That is a good idea, and is now implemented. By default it will have 0 destination buttons, but that can be changed to be up to 8.

----

I've also re-targeted this to develop, so that it can appear as a single 20x2 surface, without needing to do any special page wrapping.

PR #2455, opened 2023-05-18
buttons page edit

calvinchan0607:
It's possible to add one more page preview on the buttons page?
This will help a lot when you rearrange your buttons/Macro , easy to copy or move. or can copy multiple buttons once

Screenshot_15


Julusian:
Please open an issue, not a pull-request.

PR #2470, opened 2023-06-06
Feat/graphics overhaul

dnmeid:
I feel like the graphics overhaul is now at a point worth thinking about merging in develop.
# New features:
* Move to skia-canvas for all internal graphics generation, that means vector-based drawing and typography is possible. skia-canvas is rust based and comes with native binaries for all our supported plattforms
* Redone all current drawing functions, added some and improved some
* Added truetype fonts for text and color emojis
* More possibilities of png handling
* Dramatically improved image quality for webui by using an internal resolution of 288x288
* improved image quality for all non 72x72 pixel buttons by individual downscaling from larger image
* Added possibility to choose font size freely between 3 and 120px
* Added finer steps for automatic font size
* Improved line breaking algorithm (speed)
* Changed all internal color handling to CSS-Strings, giving possibility for alpha and different color models
* Changed internal image buffers to be RGBA all the way. More data but it turned out that the conversion between RGBA and RGB takes much longer than pushing more bytes

# Caveats and points to be discussed:
* New modules and changes in webui means you should run `yarn update`
* Truetype fonts are quite large, specially the color emoji. All together 48MB
* The truetype fonts have a little different sizing than our old bitmap fonts, so existing configurations may have different linebreaks. I tried to get as close as possible, but there is definitely a difference. Especially the old fixed font sizes had been chosen to each one fit one more line per button. 18pt had been three lines and now it is only two lines.
* The overall speed of the new rendering is not faster by magnitudes. On my machine rendering a basic button is about 0.5ms and the whole stuff from render to idle is about 0.9ms. The times are not very constant the same button which averages at 0.5ms can be finished in 0.3ms and it also can take 1.2ms. Adding more text to the button greatly increases rendering time. A completely filled button with 7pt text can take up to 4ms to render, same button with 'auto' 30ms this is not ideal but also not slower than it was before. Main factor for this increase is the massive amount of checking if text fits. I took different approaches on how to improve this. There is a built-in line breaking functionality in skia-canvas which is a lot faster (ca. factor 3), but it is not accurate and I regulary got longer lines than wanted and it can only break at spaces. For our small buttons I'd prefer to have mor breaking characters. The initial rendering after application startup takes about 600-700ms, initializing and font loading I guess.
* I adjusted the drawing functions for Streamdecks, Loupedeck, Infinitton, Emulator and Webui. I don't have all the devices and can't check them. Especially the SD+ strip I don't know how it will work. I didn't adjust anything for Sattelite or Cloud and this is very likely to break.
* There are still some console.logs and some legacy stuff which needs cleanup but doesn't hurt for the moment.
* I only tested electron builds for intel-mac 

Feel free to give me your comments, should we dare to merge now?


lib/Surface/USB/Infinitton.js
```diff
@@ -104,8 +105,26 @@ class SurfaceUSBInfinitton {
 		key = this.mapButton(key)
 
 		if (key >= 0 && !isNaN(key)) {
-			buffer = rotateBuffer(buffer, this.config.rotation)
-			this.Infinitton.fillImage(key, buffer)
+			let imagesize = Math.sqrt(buffer.length / 4) // TODO: assuming here that the image is square
+			const rotation = translateRotation(this.config.rotation)
+
+			let newbuffer = buffer
+			if (imagesize !== 72 || rotation !== null) {
```
Julusian: I'm not sure we can skip this, as we need to do rgba -> rgb conversion anway.
But this infiniton class is not hooked up, so probably time to delete it..

lib/Surface/USB/ElgatoStreamDeck.js
```diff
@@ -55,16 +56,18 @@ class SurfaceUSBElgatoStreamDeck extends EventEmitter {
 				return
 			} else {
 				try {
-					let image = imageRs.ImageTransformer.fromBuffer(buffer, 72, 72, imageRs.PixelFormat.Rgb)
+					let imagesize = Math.sqrt(buffer.length/4) // TODO: assuming here that the image is square
```
Julusian: perhaps we could pass around an object wrapping the buffer to also contain the dimensions of it?

lib/Graphics/Controller.js
```diff
@@ -30,6 +31,20 @@ class GraphicsController extends CoreBase {
...
+			'Companion-mono': ['assets/Fonts/NotoSansMono-wdth-wght.ttf'],
+			'Companion-symbols': ['assets/Fonts/NotoSansSymbols-wght.ttf', 'assets/Fonts/NotoSansSymbols2-Regular.ttf', 'assets/Fonts/NotoSansMath-Regular.ttf', 'assets/Fonts/NotoMusic-Regular.ttf', 'assets/Fonts/NotoSansLinearA-Regular.ttf', 'assets/Fonts/NotoSansLinearB-Regular.ttf'],
+			'Companion-emoji': ['assets/Fonts/NotoColorEmoji-compat.ttf'],
+			'5x7': ['assets/Fonts/pf_tempesta_seven.ttf']
+		})
+		this.fonts = FontLibrary.families
+		//console.log('companion fonts\n', companionFonts)
+		//console.log('fonts\n', this.fonts)
+		//this.fonts.forEach(f => console.log(FontLibrary.family(f)))
+
```
Julusian:
```suggestion
```

webui/src/Components/ButtonPreview.jsx
```diff
@@ -2,19 +2,40 @@ import React from 'react'
...
+	bmpHeader.writeUInt32LE(0, 46) // colors
+	bmpHeader.writeUInt32LE(0, 50) // importantColors
+	bmpHeader.writeUInt32LE(0x000000ff, 54) // Red Bitmask
+	bmpHeader.writeUInt32LE(0x0000ff00, 58) // Green Bitmask
+	bmpHeader.writeUInt32LE(0x00ff0000, 62) // Blue Bitmask
+	bmpHeader.writeUInt32LE(0xff000000, 66) // Alpha Bitmask
+	bmpHeader.write('sRGB', 70, 4) // colorspace
+
+	console.log('received image size=' + imageSize, 'length='+sourceData.length)
+	console.log(Array.from(bmpHeader).map(i => i.toString(16)))
```
Julusian:
```suggestion
```

webui/src/Components/ButtonPreview.jsx
```diff
@@ -2,19 +2,40 @@ import React from 'react'
...
+	const imageSize = Math.sqrt(sourceData.length / 4)
 
-	const convertedData = Buffer.alloc(sourceData.length)
-	for (let i = 0; i < sourceData.length; i += 3) {
-		// convert bgr to rgb
-		convertedData.writeUInt8(sourceData.readUInt8(i), i + 2)
-		convertedData.writeUInt8(sourceData.readUInt8(i + 1), i + 1)
-		convertedData.writeUInt8(sourceData.readUInt8(i + 2), i)
-	}
+	const bmpHeaderSize = 138
```
Julusian, line 14: It looks like this was the only usage of `PREVIEW_BMP_HEADER`. If so it would make sense to remove it.
Or another way to go would be to convert `PREVIEW_BMP_HEADER` into a function so that it can be reused

webui/src/Components/ButtonPreview.jsx
```diff
@@ -2,19 +2,40 @@ import React from 'react'
...
+	bmpHeader.writeUInt32LE(3, 30) // compress
+	bmpHeader.writeUInt32LE(sourceData.length, 34) // data size
+	bmpHeader.writeUInt32LE(2835, 38) // hr
+	bmpHeader.writeUInt32LE(2835, 42) // vr
+	bmpHeader.writeUInt32LE(0, 46) // colors
+	bmpHeader.writeUInt32LE(0, 50) // importantColors
+	bmpHeader.writeUInt32LE(0x000000ff, 54) // Red Bitmask
+	bmpHeader.writeUInt32LE(0x0000ff00, 58) // Green Bitmask
+	bmpHeader.writeUInt32LE(0x00ff0000, 62) // Blue Bitmask
+	bmpHeader.writeUInt32LE(0xff000000, 66) // Alpha Bitmask
```
Julusian, line 36: Nice! I like that this avoids the need to do the pixel reordering. 
I guess I never looked at the bmp header spec enough to figure out this was possible

lib/Graphics/Renderer.js
```diff
@@ -68,46 +78,48 @@ export default class GraphicsRenderer {
...
 
-			img.fillColor(img.rgb(15, 15, 15))
+			img.fillColor(colorDarkGrey)
 
 			if (options.page_plusminus) {
-				img.drawChar(30, 20, options.page_direction_flipped ? charCodeMinus : charCodePlus, colorWhite, 0, 1)
+				img.drawTextLine(31, 20, options.page_direction_flipped ? '–' : '+', colorWhite, 18)
 			} else {
-				img.drawChar(26, 20, 'arrow_up', colorWhite, 'icon')
+				img.drawPath([[46,30], [36,20], [26,30]], colorWhite, 2)
```
Julusian:
```suggestion
				// Arrow up
				img.drawPath([[46,30], [36,20], [26,30]], colorWhite, 2)
```

lib/Graphics/Renderer.js
```diff
@@ -68,46 +78,48 @@ export default class GraphicsRenderer {
...
 
-			img.fillColor(img.rgb(15, 15, 15))
+			img.fillColor(colorDarkGrey)
 
 			if (options.page_plusminus) {
-				img.drawChar(30, 40, options.page_direction_flipped ? charCodePlus : charCodeMinus, colorWhite, 0, 1)
+				img.drawTextLine(31, 36, options.page_direction_flipped ? '+' : '–', colorWhite, 18)
 			} else {
-				img.drawChar(26, 40, 'arrow_down', colorWhite, 'icon')
+				img.drawPath([[46,40], [36,50], [26,40]], colorWhite, 2)
```
Julusian:
```suggestion
				// Arrow down
				img.drawPath([[46,40], [36,50], [26,40]], colorWhite, 2)
```

dnmeid:
I found another problem: dragging presets now shows the large image during drag as a preview. This is not a problem of the graphics itself. I've seen that it actually is the image which is draggable and not the component. It should be the div around the image. If I set the drag ref to the div, another error kicks in: it looks like it should, but the image in the div is not shown the first time you drag the div.
Another point why I don't like react-dnd.

lib/Graphics/Renderer.js
```diff
@@ -270,16 +288,16 @@ export default class GraphicsRenderer {
...
+		img.fillColor(colorDarkGrey)
+		img.drawTextLineAligned(36, 36, `${num}`, colorWhite, 44, 'center', 'center')
 		return img.bufferAndTime()
 	}
 
 	static drawPincodeEntry(code) {
-		const img = new Image(72, 72)
-		img.fillColor(img.rgb(15, 15, 15))
-		img.drawAlignedText(0, -10, 72, 72, 'Lockout', colorButtonYellow, 14, undefined, 44, 'center', 'center')
+		const img = new Image(72, 72, 4)
```
Julusian, line 331: should these pincode bitmaps also be created at higher resolution?

lib/Graphics/Image.js
```diff
@@ -552,134 +962,145 @@ class Image {
...
-			// 	char,
-			// 	String.fromCharCode(parseInt(char))
-			// )
-			return 0
+		// create HTML compatible imageData object
+		let imageData
+		try {
+			imageData = new ImageData(buffer, width, height)
+		} catch (error) {
+			console.log('new ImageData failed', error);
```
Julusian:
```suggestion
			console.log('new ImageData failed', error);
			return
```
It would be good for any permanent `console.log` to be replaced with the `logger`

lib/Graphics/Image.js
```diff
@@ -552,134 +962,145 @@ class Image {
...
-				if (dummy == false) {
-					for (let len = 0; len < gfx[pixel][2]; len++) {
-						this.pixel(x + gfx[pixel][0], y + gfx[pixel][1] + len, color)
-					}
-				}
-			}
-		}
-
-		return maxX
+		this.context2d.drawImage(imageCanvas, x, y)
```
Julusian, line 873: will this do the correct scaling and alignment?

lib/Graphics/Image.js
```diff
@@ -552,134 +962,145 @@ class Image {
...
 
 	toBase64() {
 		return this.buffer().toString('base64')
 	}
 
 	buffer() {
-		return Buffer.concat(this.canvas)
+		//return Buffer.concat(this.canvas)
+		//const buffer = new Buffer.from(this.context2d.getImageData(0, 0, this.width, this.height).data.filter((_elem, index) => { return (index + 1) % 4 }))
+		const buffer = new Buffer.from(this.context2d.getImageData(0, 0, this.width, this.height).data)
```
Julusian:
```suggestion
		const buffer = Buffer.from(this.context2d.getImageData(0, 0, this.width, this.height).data)
```

lib/Graphics/Image.js
```diff
@@ -261,55 +291,428 @@ class Image {
...
-	int spacing: how much space should be between letters, leave undefined for spacing of font
-	int size: font size multiplier
-	string halign: horizontal alignment left, center, right
-	string valign: vertical alignment top, center, bottom
-	bool dummy: don't actually draw anything if true, just return if the text fits
-
-	returns true if text fits
-*/
-	drawAlignedText(
+	drawAlignedText_old(
```
Julusian: it would be good to remove these old methods, and the large blob of font.js that this depends upon

Julusian (approved):
It would be good to remove all the commented out testing log lines

I haven't read much of the Image.js code in any detail, as long as it works Im happy :)

lib/Surface/USB/Infinitton.js
dnmeid: Good catch, did the shortcut when it was still RGB buffer

lib/Surface/USB/ElgatoStreamDeck.js
dnmeid: Yep, this is only the first step. Graphics-overhaul will get more stuff.

lib/Graphics/Renderer.js
dnmeid: Is there a reason why not? They are not much update traffic. The real thing happens at startup when all button graphics are rendered.

lib/Graphics/Image.js
dnmeid: This draws the buffers, e.g. from feedbacks or the cloud icon, which do not reflect scaling. They get drawn as they are and will look exactly like they have been before.
The upsampling is taken into account, but nothing else.

lib/Graphics/Image.js
dnmeid: Yes, like I wrote, cleanup is to be done.

lib/Graphics/Renderer.js
dnmeid: I see the point that you think they are only at the buttons and the buttons do only have 72x72-100x100 anyway.
True, but it is better to downscale with some headroom and a huge point is that skia-canvas is really bad at drawing small text. Really, really bad. So it is better to draw text large and downscale the result.
Also the emulator benefits from it.

dnmeid:
@haakonnessjoen I don't know the impact on the cloud system. The little cloud symbol will be drawn correct, but do you consume image buffer somwhere? the image buffer is now RGBA and can have different sizes.

haakonnessjoen:
> @haakonnessjoen I don't know the impact on the cloud system. The little cloud symbol will be drawn correct, but do you consume image buffer somwhere? the image buffer is now RGBA and can have different sizes.

No, it only uses the object data. So it uses for example png64 etc. It is companion that is generating the button on both sides. Do we support sending raw graphics in the "bank object"? Or only png64? If we support raw data we might need to make sure that companion cloud only runs on versions that comes after this feature. It's not released yet, so its still time to not support anything before your feature.

dnmeid:
at the moment we support two ways of getting images on the buttons.
1. pgn data from the ui, this will be stored in png buffer in the bank object. the change with the new graphics handling is that images now can have a wider range of sizes.  Previously they had been scaled on upload and had been stored in 72x72. Now only very large images are downscaled on upload and only to an intermediate size. This is stored in the bank and will be scaled to fit on demand.
2. images coming from feedbacks, they are argb buffers and are inserted dynamically. At the moment we do not support higher resolutions or scaling here, so they will look at the buttons like they did before but in the webui they will look degraded.

dnmeid:
The bank object itself is not changed yet, but will be in one of the next steps. There will be more styling options.

lib/Graphics/Renderer.js
Julusian: Ignore me, I didn't realise that the 3rd parameter is for oversampling. I read this as still being rendered at 72x72, not at the higher resolution

lib/Graphics/Image.js
```diff
@@ -38,466 +40,767 @@ class Image {
...
+	 * @param {string} text
+	 * @param {string} color CSS color string
+	 * @param {number} fontsize Em height
+	 * @param {boolean|undefined} dummy Don't actually draw anything just return the text width
+	 * @returns {number} width of the line
+	 */
+	drawTextLine(x, y, text, color, fontsize, dummy) {
+		if (text === undefined || text.length == 0) return 0
+
+		if (parseInt(fontsize) == NaN) return 0
```
Julusian, line 351: This check will always return false, as `NaN` is not equal to itself. Use `isNaN` instead

bryce-seifert:
Screenshot 2023-06-12 at 8 07 43 PM
Screenshot 2023-06-12 at 8 07 47 PM
 src="https://github.com/bitfocus/companion/assets/50299051/5e1697ad-6aaf-4e90-ae7c-3816a77068eb">
Not sure if this is related to this commit, but when using the develop build I'm now not able to see buttons in Safari. Chrome and Firefox seem to work fine.

Julusian:
@bryce-seifert fixed!

PR #2480, opened 2023-06-14
feat: use crisper image rendering when buttons are upscaled

leolabs:
When viewing the button grid on a larger screen, by default browsers try to smoothen out the images which makes them look blurry. This PR adds a query that forces browsers to use nearest-neighbour upscaling when buttons exceed the 72px width that images are rendered at.

IMO, this makes the text easier to read because it has more contrast against its background. What do you think?

Before:
CleanShot 2023-06-14 at 11 49 49@2x

After:
CleanShot 2023-06-14 at 11 49 38@2x

### DIFF
--- a/webui/src/scss/_bank.scss
+++ b/webui/src/scss/_bank.scss
@@ -8,6 +8,7 @@
 	height: fit-content;
 	display: inline-block;
 	border: 1px solid transparent;
+	container-type: inline-size;
 
 	&.clickable {
 		cursor: pointer;
@@ -50,6 +51,10 @@
 		margin: 0 auto;
 
 		-webkit-user-drag: none;
+
+		@container (min-width: 72px) {
+			image-rendering: pixelated;
+		}
 	}
 }
 


dnmeid:
Thank you for your contribution, but buttons will look much better in the next version. Check out development branch / expermental build.

leolabs:
Oh, that sounds promising! Looking forward to testing it out 🚀

leolabs:
@dnmeid I just checked out the new button rendering in the develop branch and it looks great! As a module developer, would it make sense to scale my icons to 288x288 pixels as well so they don't look more pixelated than the text surrounding them?

Currently, they look a bit out of place:
CleanShot 2023-06-23 at 12 48 57

dnmeid:
> As a module developer, would it make sense to scale my icons to 288x288 pixels as well so they don't look more pixelated than the text surrounding them?

This is currently not possible. The imageBuffer doesn't have a associated image size and assumes a fixed width of 72px.
I'm also working on improvements here. e.g. svg rendering, then we are resolution independent and users may be able to choose fill colors for images.

leolabs:
Alright, thanks for letting me know! SVG icons sound sweet as well. My icons are all SVG anyway, so that would remove one conversion step to add them to Companion.

Issue #2345, opened 2023-02-23
Extend expressions to action options


dnmeid:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

Button text can be entered either as text (with inline variable support) or as an expression with numerous calculations and functions available. For the modules on the other hand there is the option to allow users to enter variables in textinput options and then parse the variables in the text.    
The goal of this feature suggestion is to extend the expressions functionality to all action options.

Aspects:

- The options get a new property, like `enableExpression: boolean`
- When the option is set, the input element gets a toggle button to switch between the default input (ordinary textinput, dropdown, checkbox, number slider...) and an expression textinput
- When the action is executed the result of the expression is calculated and typecasted to what the option needs
  - textinput -> result as string
  - chekbox -> result is truthy value
  - number -> result parsed as number obeying step width
  - dropdown -> when result is number then index of options, when text then matching item
  - multidropdown -> when result is array of numbers or strings then multiple items, when number or string then single item
  - variable -> result as string
- failed expressions should lead to the options default value, module developers need to make sure that default value makes sense

Possibilities **tbd**:

- additional property like `expressionFallbackValue` to give control to module what happens if the expression fails
- extend expressions also to feedback options, but not sure about the usecases at the moment. It would require to track feedbacks with expressions and then recalculate them if any used variable in the expression changes 

### Usecases

Using variables or custom variables in actions and at the same time retaining the convenience of input elements like dropdown menu for manual selection.

Julusian:
> failed expressions should lead to the options default value, module developers need to make sure that default value makes sense

I think it would make more sense for the action to be skipped. Maybe `expressionFallbackValue` would be a good idea to opt into the action to still be called.

> extend expressions also to feedback options, but not sure about the usecases at the moment. It would require to track feedbacks with expressions and then recalculate them if any used variable in the expression changes

As of https://github.com/bitfocus/companion/pull/2251 we handle this for feedbacks for basic variable parsing, it should be possible to extend to expressions safely.

-----

I started playing with this almost a year back in https://github.com/bitfocus/companion/pull/2032, and have a very long note with my thoughts https://github.com/orgs/bitfocus/projects/2?query=is%3Aopen+sort%3Aupdated-desc&pane=issue&itemId=19678623.

[...]

Issue #1909, opened 2022-02-13
Add variables in the scope of a button


BryanCrotaz:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

Variables:
$(button:text)
$(button:enabled)
$(button:disabled)
$(button:page)
$(button:number)
$(button:col)
$(button:row)

These variables are only valid within a button (not within a trigger for example)

### Usecases

External system sets up button names and wants to know when each one is pressed.

Set up a page of buttons with OSC actions with a path of /my-action/$(button:text) or /my-action/$(button:page)/$(button:number
Copy paste now works without post-editing the buttons.

Then using OSC or http, push all the available names into the page. When the button is pressed, you know which one it was

In my example it's a sports team and the OBS script loads video stingers from a folder and puts the filename on a button. When that button is pressed it looks up the video by name and plays it. With 32 on each team it takes AGES to set up the companion page manually.

jsjoen7896:
I would LOVE for this to be added to the stable version of Companion. My use case would be to use the page number and button number to automate variable actions.

I have a more complicated use case, but a simple example is as follows:
I am using vMix to cut cameras/inputs/etc., and also have feedbacks to change the color of the button to green or red depending on if it is in preview or program. I use the button number to substitute my manually entered number. This way, I can copy and paste the button and it will automatically be using the current page/button numbers to trigger/feedback the respective input.

My expression would look like this:
($(internal:this-page)-1 * 32) + $(internal:this-button)
That way, I can copy and paste to any page and it would increment by 1/32. And of course you can do any other calculation to get more custom results!

PS:
I love that companion has the modulo operator available! It would be so cool if they could somehow add ternary operators, and possibly conditional operators too in the future!

sstaub:
I would also like to handle page numbers as variables. In my case I want to step around pages with one! button action.
e.g. I have three pages I want to step trough, starting with page 1, the next press should go to page 2 ... when in page 3 the next press should go to page 1 again.

Julusian:
I have a big spanner to throw into this idea, that I don't have a solution for currently.

In 3.1/3.2 I am planning to do an overhaul of pages, and while doing so I want to allow for a looser form of #27,  #1676 and some similar things.  
[...]

PR #2499, opened 2023-06-24
feat: groundwork for grid size

Julusian:
This lays the groundwork for being able to change the grid size, with the following key changes:
* Rework controlIds to be generated as `bank:<uuid>` instead of `bank:<page>:<bank>`
* At places where we need a `bank` number, that has been replaced with a coordinate string (eg 0x0)
* The PagesController now contains a mapping of what controlIds are at what coordinates.
* Rework surfaces to operate in xy coordinate instead of the `bank` number
* Internal actions have been updated to use coordintes instead, maintaining backwards compatibility
* Modules are no longer informed of the page & bank their actions & feedbacks are on, both values will now be zero. These properties always been marked as deprecated in the new module api and are only there for the legacy module wrapper.

This is a very sprawling change, and it is likely to have broken things in the process.  
With this change, there should be no technical reason for being limited to 8x4. 

All of the protocols (including cloud) are still using the old bank numbers, further changes will be needed to allow for clients to use coordinates instead.


krocheck:
I'll start looking through this.  I like the coordinates system compared to letters for rows like we discussed a long time ago.  I see v4 DB; that's logical.  Good work!

I know I've hard-coded different grid sizes to test how the software reacts, and I'd think it'll go quick to add the user ability to manipulate that.  The UI starts to need zoom and scrolling, not unlike what Dorian and you have called out in roadmap discussion.  I think ultimately we should auto-detect a minimum grid size on newer installs to match the grid of the largest panel connected (up to the XLE-128), even automatically or prompting when a larger panel is detected to ask to grow the grid.

Do you have a breakdown where in the software execution that `bankControlId` vs `page.coordinate` is used?  I know from my experience investigating implementing this exact PR that the entanglement of `page,bank` was such that it would end up being easier to lean on `page.coordinate` in a lot of places even though we'd ultimately prefer `bankControlId` for the long-term goal of placing those controls persistently in coordinates or across multiple pages ad hoc.  I'm curious how you navigated that.

dnmeid:
I did try it and for me the coordinate formatting of XxY feels somehow unnatural. I see that we are used to coordinates like A1, B3... in Spreadsheets, but with a numeric system I'd prefer the row in first position and the column in second position. This is more like our normal reading habits: first lines, next positions. X/Y is more the mathematical order, but if we'd follow the mathematical rules, we would have to go in the positive X direction to the right and positive Y upwards.
Also the "x" doesn't look right to me, I always have the association with a size and not a position. Most often vector components are separated by a "/". Then the new standard would be P.Y/X[.S]
First I thought to stick with the dot, but a different character than the "." would be beneficial for API use, because 2.4 can be misinterpreted as a float.
In the option labeling I would prefer "row" and "column" over "X" and "Y", this is less technical and more descriptive.
Concerning the internal actions and feedbacks, I feel they are not uniform. In the actions there is this mode dropdown and in the feedback a checkbox for this...
We have the feature requests for making all options expressionable (#2345) and for variables in the scope of a control (#1909). With these features we could simplify and unify the internal actions and feedbacks.

@krocheck I thought the coordinate and controlId are two separate things, a control gets an id at creation and keeps this id for its lifespan. It can be relocated to a different coordinate and still keep the id. Also I think in the future we will see controls spanning more than one grid cell. And we could use the same control at different places. The coordinates are not a property of the control, but the controlId is a property of a grid cell at coordinates.

lib/Data/Upgrades/v3tov4.js
```diff
@@ -0,0 +1,94 @@
...
+function addControlIdsToPages(db) {
+	const controls = db.getKey('controls', {})
+	const pages = db.getKey('page', {})
+
+	const maxButtons = 32
+	const perRow = 8
+
+	for (const [number, page] of Object.entries(pages)) {
+		if (!page.controls) {
+			page.controls = {}
```
dnmeid, line 17: I wonder if it would be better to continue the logic of the pages and split this in nested objects for rows and columns.
e.g. `"page":{"1":{"0": {"4": "bank:Xy3fZt4"}}}` for a control at row 0 and column 4 of page 1.

lib/Data/Upgrades/v3tov4.js
```diff
@@ -0,0 +1,94 @@
...
+				page: Number(match[1]),
+				bank: Number(match[2]),
+			}
+		}
+	}
+
+	return undefined
+}
+
+function convertImportToV4(obj) {
```
dnmeid, line 52: old rotate actions objects can also be deleted

lib/Data/Upgrades/v3tov4.js
Julusian: A lot of the backend logic is using the 'coordinate' as a string, for simplicity of keeping it one parameter. which makes storing it in the db as a single level deep be simpler.

But I am wondering about whether instead of this string, whether passing the coordinate as an object would be messy. the page number could even be combined into that object when appropriate

Julusian:
> I see that we are used to coordinates like A1, B3... in Spreadsheets

Perhaps one thing to confirm is whether we want this grid to be able to grow in both directions, and not just down and to the right like spreadsheets. If we were to limit it to that direction then we could keep it simpler and go with A1 etc.  
I do think it will be useful to be able to grow in all directions, but I am open to being convinced otherwise.

> I'd prefer the row in first position and the column in second position.

I admit that at points I have been getting it the wrong way around

> Also the "x" doesn't look right to me, I always have the association with a size and not a position. Most often vector components are separated by a "/". Then the new standard would be P.Y/X[.S]

I don't have a strong preference on the character here. Perhaps we should change the separator used between the other parts too and make it `P/Y/X[/S]`?


> In the option labeling I would prefer "row" and "column" over "X" and "Y", this is less technical and more descriptive.
Concerning the internal actions and feedbacks, I feel they are not uniform. In the actions there is this mode dropdown and in the feedback a checkbox for this...

Im happy to refer to them as row and column.  
The difference between the actions and feedbacks is simply because the feedbacks dont have the 'expression' mode. and a dropdown with 2 options felt weird, but I didnt think about the lack of consistency.

What are your thoughts on the row & column being separate fields in the actions? or should that be a single field where the user should write in `0x0`? One benefit to the single field is that it could support the old bank indexes more easily than the `_legacy` actions I created for it.

> I know I've hard-coded different grid sizes to test how the software reacts, and I'd think it'll go quick to add the user ability to manipulate that. 

I've been trying to avoid doing anything like that for now, just to keep the scope managable. But once this is merged it should be a lot easier to gradually implement the missing support into more areas. Putting a very crude resize option in the experimental settings would be a good place to start (accepting that the button editor grid will become visually broken from this)

> Do you have a breakdown where in the software execution that bankControlId vs page.coordinate is used? I know from my experience investigating implementing this exact PR that the entanglement of page,bank was such that it would end up being easier to lean on page.coordinate in a lot of places even though we'd ultimately prefer bankControlId for the long-term goal of placing those controls persistently in coordinates or across multiple pages ad hoc. I'm curious how you navigated that.

It varies depending on the area, and depending on the intent. If something is tied to a particular location, then it is using the coordinte. If it should be tied to a partcular control then it will use the controlId.

In general the actions, feedbacks and controls are only aware of the controlId. The exception being during drawing, and internal actions/feedbacks are aware of the location. 
For now, all the internal actions/feedbacks that allow for targetting another button use the page+coordinate, I expect in the future alternate versions will be added which will target a controlId.

In the surfaces, they mostly consider the coordinate, until they get to the point of needing to do something (like a press) where they lookup the controlId at a coordinate.

The ui (and all the commands sent back and forth while editing a button), is controlId based. Except when streaming the pages of button renders, when we instead do it by coordinate. But a single preview is done by controlId instead.

Does that answer your question?

lib/Data/Upgrades/v3tov4.js
dnmeid: I think the most challenging part of this database is not storing data or deleting it (which is more mork with a nested structure), it is retrieving data. Today we have the need to get either the control on one specific coordinate, which is eqaully easy on all versions, or to get a complete page. I think in the future we will need an efficient way of retrieving a rectangular cutout. E.g for what is currently shown in UI or what is in the window.
Let's say we want all controls from 1/1 to 5/5. With the combined string we would need an Array.find() going over all the keys of the controls of the potentially huge page. Or we generate all existing coordinates in the range and test if there is a control at the position.
My mind's not made up for a specific solution yet, I'm just thinking loud.

dnmeid:
> I do think it will be useful to be able to grow in all directions, but I am open to being convinced otherwise.

All directions is my preference too.

> make it P/Y/X[/S]

It will be not as eye pleasing, but yes, I think that's the best solution.
There is one more aspect to think of. If we introduce windows in the future, will it be W/P/Y/X[/S]? Even if I do not expect that positions will exceed the three-digit range, the text is getting longer and longer. Maybe we need settings options to finetune what is shown in the topbar.

> What are your thoughts on the row & column being separate fields in the actions? or should that be a single field where the user should write in 0x0?

Yes, in the end I think the single field will be more efficient. Most users will want to address a particular control and typing `4/3`will be the most straightforward way (by the way for the entry we could accept any delimiter except "-"). "this row" or "this column" can be solved with scoped variables once they are made.
The page now is a separate option, we can override the page option, if data is entered in form of `1/2/3`, this would make it even more easy to use data from variables.
Only thing which would change the existing action is, currently the position field is an expression, which is fine because a single number evaluates to a valid result. A string in an expression has to be quoted tough. What would be a good solution here? I think we should change the field to a textinput with variables parsing by default. I assume that #2345 will be done when we release this and then we could have a more sophisitcated upgrade function turning the expression on if it actually was used.

Julusian:
> There is one more aspect to think of. If we introduce windows in the future, will it be W/P/Y/X[/S]?

Will that be shown here? Aren't windows related to a specific surface, and not the control/grid? 

> Yes, in the end I think the single field will be more efficient. 

Oh, right yeah so the 'problem' with doing it as a single field with variables for this, is that will be much harder to show the preview of the target button. But as we are setting up a subscription to load the render we need, maybe it wouldnt be too bad to make that subscription resolve the expression..

> The page now is a separate option, we can override the page option, if data is entered in form of 1/2/3, this would make it even more easy to use data from variables.

hmm possibly. If we want to allow that we will need to give the old bank index a prefix/suffix so that we can differentiate them. Other than that I think it would work ok? Presumably if the page is omitted, use the current page.

> A string in an expression has to be quoted tough. What would be a good solution here? I think we should change the field to a textinput with variables parsing by default.

I think this is a general thing we could improve with the expression sytnax. Im also not keen on how variables have to be written as `${$(aaaa)}`.  
Maybe we could allow the outer backticks to be omitted? If parsing the string fails, we could try again wrapped in backticks?  
Seeing how the valid characters in variable names is really strict now, maybe we could allow skipping the `$(` wrapper around variables in expressions? Because of the presence of the colon, it will probably be easily identifiable from other tokens?
Or maybe the parser could be made to accept `$()` being used in the string interpolation.
Combined it would mean that `${abc}/${def}` could be accepted by an expression. Not 100% the same as the other syntax, but just as short and easy.
The parser is a very plugin based library, and our variable syntax is already its own plugin

dnmeid:
>> There is one more aspect to think of. If we introduce windows in the future, will it be W/P/Y/X[/S]?

> Will that be shown here? Aren't windows related to a specific surface, and not the control/grid?

Correct, but the question is if and how to show windows on the surface. How can I tell from my streamdeck buttons which button comes from which window? Is this information relevant for me and it should be always available? Is it opt in? Is it opt out?
I think with a more dynamic control assignment to buttons, it will be more important to know where a particular control comes from.

> I think this is a general thing we could improve with the expression sytnax. Im also not keen on how variables have to be written as `${$(aaaa)}`.

Yeah, I do think too that the expression syntax should be more focused on variables, operators and functions and it would be really an advantage to write variables just like aaa:bbb.  Unfortunately I don't see a good method to make expression input automatically compatible with a string input. 6/3 is a valid expression and evaluates to 2. We can improve a little bit if we return the plain input if the expression can't be evaluated, maybe also doing a best guess type casting. But if we want to tackle the coordinate input that way, we need a delimiter which is not used otherwise. .-+*/%:^&|\=$ are out. `_` and `#` would be candidates. Actually I think the `#` would also make at least a little iconographic sense.

lib/Data/Upgrades/v3tov4.js
Julusian: I don't think splitting it into another layer will help at all with that. Either way it will require the same level of searching.

But I am probably going to change this as part of other refactoring to stop storing a string coordinate.

Julusian:
I've gone through and changed it so that instead of passing around 'coordinate' as a string, there is now a location object looking like:
```
{
  pageNumber: number
  column: number
  row: number
}
```

This was cleaner than I was originally expecting (when I chose the other route a couple of months ago). And this means that presenting that as a string is 99% done in the ui (plus a few points of logging), so we can be less decided on the format and order of the values.

Julusian:
OK, those internal actions and feedbacks are reworked a bit more. And this includes a crude and limited functionality of 'local' variables, so that `$(this:page)` works. But it will only work in these specific actions and feedbacks for now.


Screenshot from 2023-06-28 23-41-37
Screenshot from 2023-06-28 23-41-32

> I assume that https://github.com/bitfocus/companion/issues/2345 will be done when we release this and then we could have a more sophisitcated upgrade function turning the expression on if it actually was used.

The internal actions in 3.0 are able to use expressions, but not regular modules. This doesnt change that

dnmeid:
>> I assume that #2345 will be done when we release this and then we could have a more sophisitcated upgrade function turning the expression on if it actually was used.

> The internal actions in 3.0 are able to use expressions, but not regular modules. This doesnt change that

I'm not decided what to do next. #2345 is one option, proceeding with graphics overhaul is the other one. Graphics-overhaul would next need a few bigger changes in the gui and a few more thoughts about backwards compatability.
What do you think, does it make sense to do #2345 independent from this branch?

Julusian:
I dont have a preference on which you do next.

I don't think we need to do #2345 at the same time as this. While it would be nice to for uniformity, most modules will lag behind anyway, and some serious thought needs to be put into the UX of 2345.
3.0.0 already has a couple of actions/feedbacks which use expressions, and I think it is ok to have the internal ones have special abilities to begin with.

dnmeid:
The Loupedeck CT is working now with one exception: when a button gets redrawn, it will only update buttons on the upper half and the first button on the lower half. On page change, everything is drawn correctly.

I've seen that the location row and column sometimes is a number and sometimes is a string. Maybe that's the problem.
Also I'm really often am confused by the order x,y. Row,column feels much more natural when 0,0 is in the top left.

Julusian:
Thanks for fixing up the CT implementation

> I've seen that the location row and column sometimes is a number and sometimes is a string. Maybe that's the problem.

Yeah, Im not surprised. This is an issue we already have with the page & bank. Sounds solvable, but they could come from anywhere...

> Also I'm really often am confused by the order x,y. Row,column feels much more natural when 0,0 is in the top left.

Yeah, I dont feel strongly that it shouldnt change. Trying to avoid doing too much at once, so now that the way I was handling 'locations' has changed, I think this can be tackled. Im not sure if it should change when they are two parameters in the code (still referred to as x&y), or just in user facing places?

dnmeid:
Ok, the number fix also fixed some of the CT problems (=== comparisons now work).
There are still pagenumbers as strings coming from webUI.

>> Row,column feels much more natural when 0,0 is in the top left.

> Yeah, I dont feel strongly that it shouldnt change.

I think row,column will be definitely better. Internal xy-stuff can be left xy, but I have allready found off by one errors.

Julusian:
> I think row,column will be definitely better. Internal xy-stuff can be left xy, but I have allready found off by one errors.

ok, I have swapped them everywhere that I think needs it.

Other than some additional testing, I think this is back to being ready

Julusian:
@dnmeid @krocheck let me know if you want me to hold off on merging this for a bit. Otherwise I shall merge it in a couple of days

dnmeid:
My only concern at the moment is, that everytime I switch back from this branch, Companion won't start because of a too new database. I thought it would just ignore newer database formats and work with the old one.

Julusian:
> My only concern at the moment is, that everytime I switch back from this branch, Companion won't start because of a too new database. I thought it would just ignore newer database formats and work with the old one.

That will happen going between this and develop, but it doesn't happen between this and beta

PR #2537, opened 2023-07-29
feat: Add render cache for bank images to improve CPU usage

### LINKED ISSUE #2536: [BUG] High CPU Usage When Button Value Changes Often

- [X] I believe this to be a bug in companion

### Describe the bug

This is less of a bug and more of a missed performance optimization, I think.

The latest version of the AbleSet module exposes some variables that can change up to 30 times per second, like the current timecode of the song. I noticed that when I add a button with the current frame to my surface, Companion's CPU usage increases significantly.

My guess is that this is caused by the graphics renderer re-rendering the button image every time it changes. This might be something that could be optimized by introducing a render cache that stores the last few (~50-100 maybe) images in memory so they don't have to be re-rendered so often.

If this is something you think would be useful, I could try to make a PR for it.

Let me know what you think!

### Steps To Reproduce

_No response_

### Expected Behavior

_No response_

```markdown
- OS: macOS 13.3 (22E252)
- Browser: Safari 16.4 (18615.1.26.11.22)
- Companion Version: 3.99.0 (3.99.0+6181-develop-b7c5f516)
```

### Additional context

_No response_

leolabs:
This PR aims to fix #2536 by introducing an LRU render cache for bank images. The 100 most recently used bank images are kept in memory so they don't have to be re-computed every time they're used.

Before:
CleanShot 2023-07-29 at 18 04 23

After:
CleanShot 2023-07-29 at 18 05 06

There might be other places that can be optimized but this seemed like the best first approach.


Julusian (changes requested):
I think it would be better to do this caching as part of the GraphicsController rather than GraphicsRenderer. Replacing this line: https://github.com/bitfocus/companion/blob/develop/lib/Graphics/Controller.js#L154

I don't like this static class now having state, the other solution would be to make it be a stateful (non-static) class, but we are already doing a layer of caching in the GraphicsController, so it would make more sense to do this extra layer there instead.  
Additionally, doing it here means that any preset drawing will be done with the cache too. I don't think it is a good idea to put the preset images into this cache, as it is highly likely they will be shown once and then discarded, which will make it essentially wipe this cache for no benefit.

dnmeid:
I'm actually surprised that a cache at this point gives a difference of 20% in CPU load. Actually the whole render process is only done if something changes on the button. A cache is definitely a benefit if the change is reverted later to a cached state, so it is a benefit for buttons which change background color or so. What consumes the most power at button rendering is definitely the text, especially when set to "auto". So here the cach can help a lot if there is a known set of possible texts. If you show a timecode variable in the text, the cache won't help because there will be more than 100 different texts.
The real key to bring load down is improoving text rendering.

leolabs:
Hey @Julusian and @dnmeid, thank you for reviewing my PR!

I've just pushed a commit that moves the caching logic into the controller. Does this look good to you?

Regarding some of the points you made in your comments:

> I don't think it is a good idea to put the preset images into this cache, as it is highly likely they will be shown once and then discarded, which will make it essentially wipe this cache for no benefit.

Since the cache always removes the least recently used item when it's full, the items that are being removed due to the previewing of buttons are probably not in use anymore anyway. If we notice that preview images are purging button images that are still in use, then we could increase the cache size.

On the other hand, button previews can reference variables that change often, which would give them the same performance boost as the buttons on the surface using the new cache.

---

> If you show a timecode variable in the text, the cache won't help because there will be more than 100 different texts.

I might have worded my initial PR description badly. In my case, the timecode is split across four buttons, like this:

CleanShot 2023-08-01 at 23 44 08

So the "frames" button quickly goes from 00 to 29 over and over again which is a perfect scenario for the cache.

---

> The real key to bring load down is improoving text rendering.

That would be great! It's been a while since I worked with the Canvas API so my knowledge on this is a bit rusty unfortunately, and I don't know how the skia-canvas library you're using performs compared to browsers, but here's a few thoughts I have on this:

I remember that setting the `context2d.font` value on every render, even if it hasn't changed, has a heavy performance impact. This might be worth checking out, maybe we can keep a cache of one canvas per font family / font size so they don't have to be re-initialized each time a button is rendered.

Regarding the wrapping of text, there's a library called [canvas-hypertxt](https://github.com/quicktype/canvas-hypertxt) which aims to be the most performant solution for rendering text by using smart caching so it can predict text wrapping after a few runs without having to actually try to draw the text to see if it fits. This might be useful here as well.

I'm looking forward to your feedback!

Julusian:
> I've just pushed a commit that moves the caching logic into the controller. Does this look good to you?

You had some bad luck with timing, causing a merge conflict, but having fixed that for you it looks good :)

---
> Since the cache always removes the least recently used item when it's full, the items that are being removed due to the previewing of buttons are probably not in use anymore anyway. If we notice that preview images are purging button images that are still in use, then we could increase the cache size.
>
> On the other hand, button previews can reference variables that change often, which would give them the same performance boost as the buttons on the surface using the new cache.

By presets I mean the previews shown when opening the preset panel for a connection. So they aren't real buttons yet, as such we don't live stream the renders, they are generated and sent to the client once when opening the panel.

However, those don't go through `#drawAndCacheBank`, so they are no longer being cached (which I think is a good thing).  
Every other generation of a image for a real bank/button does go through `#drawAndCacheBank`, so this cache will be used for the ui/surfaces/whatever just fine.

----

@dnmeid 
> I'm actually surprised that a cache at this point gives a difference of 20% in CPU load.

I read those numbers as a 11% improvement, which is still a pretty noticable improvement.  

True, this won't help always, but the diffing looks cheap enough to not impact new drawings. This will help especially in cases where text isnt changing, but for example buttons flashing (which can be done with a trigger), or ones which are toggled between an on and off state often.  
Caching is definitely an area we can explore a lot more in the future. If we really tried to tackle various scenarios, we could probably avoid a lot of redraws.
A good one could be to have a cache per bank, but only if it doesnt use any variables and only uses boolean feedbacks. In that scenario it will have a known number of states, so the cache will have known maximum size. Maybe don't cache them all, or maybe use a ttl as a limit instead.  
But of course this could become problematic for users with less memory, so not something I want to rush into.

I think we are already limiting how often a button is redrawn, but its quite generous (50-100fps) so perhaps another easy win would be to more aggressively limit this. From memory streamdecks cant draw that fast anyway.

Julusian (approved):

dnmeid:
@Julusian I think the current redraw frequency limitation is the debounce function with 10ms

@leolabs I allready had a look at canvas-hypertext and many other libraries. I couldn't find one which ticks all the marks. Because we have very limited space at the buttons, we have some extra line-breaking characters, which no other library offers.
But the implementation of the canvas based drawing like it is now is very new and not fully optimized yet.
Thanks for the tip with the font calls. At the moment the font is only changed when it is actually needed. Once for the topbar and once for the text. Exception is when font size is "auto" because we try-fit many fontsizes.
I wonder how this could be improoved. Currently a new canvas is created for every bank image. Maybe we create one dedicated canvas for every size, use that only for rendering text and put the final text on the resulting bank.

Issue #409, opened 2018-12-16
Feature Request: Button layout rotation on programming page


BasilChessman:
**Describe the feature**
When rotating buttons on a surface, have the button layout page rotate as well.

**Is this platform dependent (windows, mac, ..)?**
no

**If documentation is required to implement, do you know where to find it?**
no

**Usecases**
Would make programming easier if the button layout rotated to match rotation changes made in surface settings.

haakonnessjoen:
How would you want to do this. As you could have several stream decks connected with different orientations?

krocheck:
I had a couple thoughts on this:
1. In the button tab you can add a rotate button which cycles the layout through the 4 orientations but is simply visual and has no actual configuration implications
2. Add a page option to force an orientation and have the button tab layout follow

davidjoshuaford:
With the excellent updates to the multi-Emulators in version 3 of Companion ahead of us, I can see real value in implementing a vertical mode, particularly for phone usage – holding a phone naturally needs to be vertical, and for one handed operation when walking around a venue. 

There are two things to add here: adding rotation to the emulators, just like the Stream Decks currently have, and

Adding an option to rotate the button building page.

Julusian:
Once we have #1607, perhaps this issue as it stands is no longer necessary?

Instead, it could be that it is possible to define the width & height of an emulator. You could then configure it to be 3x6 or whatever works well for your phone. I almost did this when implementing the emulator styling, but decided to leave it out initially

And perhaps the current way of rotating streamdecks should be deprecated (probably cant be removed for at least a few releases), replacing it with a rotation mode that actually rotates the whole streamdeck, not each button individually. 
Meaning that to rotate a XL, you will need to resize the grid to be at least 8 tall, then once rotated it will take 4x8 of the buttons (rather than the current fixed 8x4)

The combination of these two means that the current orentiation in the UI will always be correct, so we don't have to think about rotating it

davidjoshuaford:
That makes sense for longer term flexibility with multiple devices. Would buttons be renamed into a spreadsheet cell terminology? ie A1, A2..... B1, B2 etc? 

I've been playing around with 4UP stream deck arrays, and also Dashmaster2k windows, and even with four of the new emulators can get a 16x8 grid like this: 
[...]

PR #2543, opened 2023-08-07
feat: improved surface rotation

Julusian:
Related: #409

Instead of rotating purely the bitmaps inside of each key, this makes it possible to rotate everything about the panel.
For example, a 15 key streamdeck normally occupied a 5x3 area in the grid and once rotated will now occupy a 3x5 area instead.
PXL_20230807_201806284

For backwards compatibility, the old rotation modes have not been removed, that can be done in a later release.

Additionally, xkeys has been reworked, so that it no longer needs to span multiple pages or do other mangling to fit the panels into the grid.
PXL_20230807_194059424
This is controlled by a setting, so that existing xkeys users can choose when to switch their configurations to the new layout.
I have only tested this with a XK-24 and a XK-12 (Jog&Shuttle)
With this new layout, rotation works as you would expect. It is also offered when using the old layout, but makes configuring even more confusing.

loupedeck ct now also uses a single page, with a suitable grid size

### DIFF
--- a/lib/Data/Upgrades/v3tov4.js
+++ b/lib/Data/Upgrades/v3tov4.js
@@ -32,6 +32,12 @@ function addControlIdsToPages(db) {
 /** do the database upgrades to convert from the v3 to the v4 format */
 function convertDatabaseToV4(db, logger) {
 	addControlIdsToPages(db)
+
+	// If xkeys was previously enabled, then preserve the old layout
+	const userconfig = db.getKey('userconfig', {})
+	if (userconfig.xkeys_enable && userconfig.xkeys_legacy_layout === undefined) {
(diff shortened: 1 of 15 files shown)


Julusian:
@dnmeid could you verify I haven't broken the loupedeck ct implementation (again)? 
It should now reside in a single page. I havent bothered with backwards compatibility, it doesnt feel worth implementing
Its not directly related to this PR, but makes sense to do at the same time.

dnmeid:
I can check CT stuff at the weekend.

dnmeid:
Done a few fixes, Loupedeck CT is happy now.

Julusian:
Thanks @dnmeid!

Issue #2547, opened 2023-08-10
'Invert' checkbox as default for all boolean feedback definitions


leonreucher:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

For the modules I developed I added an option to invert the boolean value in all boolean feedback definitions. Probably it would be a nice idea, to implement this feature for all feedbacks which are boolean based?

![Bildschirmfoto 2023-08-10 um 23 35 59](https://github.com/bitfocus/companion/assets/19686873/f79ea121-aaa8-403d-bcba-d565f3542143)

### Usecases

Invert the logic for already designed buttons, without having the need to change the design.

Julusian:
This is now in the beta builds. See #2549 for details on how it works, it has also been added to the module docs https://github.com/bitfocus/companion-module-base/wiki/Feedbacks#inverting-boolean-feedbacks

Issue #1649, opened 2021-07-02
Add Trigger when Feedback is False


istnv:
**Describe the feature**
The current boolean feedback provides an automatic trigger to fire when the feedback is True/On. 
There are no options to trigger actions when that feedback changes to False/Off.
This request is to add the ability to trigger actions when the feedback is False/Off.

**Usecases**
Certain module feedbacks indicate the status of the connected device and can be altered outside of Companion. Watching the 'Mute' status of an audio mixer will allow muting/unmuting other channels depending on the status. To be fully useful, a trigger should be available for both states (on/true or off/false).

Julusian:
I think this would be best achieved by having 2 sets of actions on each trigger. A set of 'on' actions and a set of 'off' actions.
That will be the most flexible and avoid the need to create almost identical triggers to handle the different directions

jmcc94:
Not every scenario will be met be just having 2 sets actions on each trigger - it would still be good if a separate trigger condition could be when a feedback is False/Off

Eg. A complex trigger consisting of when feedback A & feedback B are both false, or when feedback C is true & feedback D is false.

Usecase - using triggers to turn on 'CAM LIVE' signs when certain sources in OBS are in PGM, and turning it off when they're not.

BryanCrotaz:
I've been noodling on the same problem for my recent project where I've ended up with about 20 triggers for a boardroom with 6 devices. In your use case, actions for true/false would work perfectly if you can have an ORed set of feedbacks. This can be modelled as If/Else:

If (src1 in pgm OR src4 in pgm OR src7 in pgm)
  turn light on
Else
  turn light off

Julusian:
From the discussion in #2549, it sounds like this is solved

PR #2549, opened 2023-08-13
feat: boolean feedback invert #2547

Julusian:
This shows an 'Invert' checkbox for all boolean feedbacks, unless the feedback already has one (simple check for field of certain id).  

This will apply to all modules, the configurability requires an updated @companion-module/base

Shown here is a feedback with the new field applied, and one which already had an invert checkbox:
image

The behaviour of this is handled internally, with modules only being aware of this in their upgrade scripts. This is so that they can migrate users to/from this builtin invert field.

Modules are able to turn this invert field on and off, by setting the `showInvert` property on their feedbacks. By default we will auto-detect whether to show the field, but if set to true/false we will respect the specified value. Module developers will need to make sure to not break user configs when changing this.

Example upgrade snippet:
```js
	for (const feedback of props.feedbacks) {
		if (feedback.feedbackId === FeedbackId.DSKOnAir && feedback.options['invert'] !== undefined) {
			feedback.isInverted = !!feedback.options['invert']
			delete feedback.options['invert']
			result.updatedFeedbacks.push(feedback)
		}
	}
```


istnv:
This solves #1649 as well.

Julusian:
I feel like #1649 is already possible with the event `On condition becoming false`. But if not then yeah this will probably help

istnv:
Both work. I think this is simpler/cleaner by acting directly on `true/false` instead of testing an expression.
Similar to `if (ok) {}` or `if (!ok) {}`vs `if (ok == true) {}` or `if (ok == false) {}`

krocheck:
Since this is a static feature of booleans, I feel like my preference is to see it on the left under the learn option rather than inline with other options.  Happy for others to disagree, just my 2 cents.

Julusian:
I have no problem with that.
I do think it will make things a little less consistent, as for a while some feedbacks will have an invert in the left and some in the right (depending on whether they define the invert field themselves). But going forward that should decrease as modules remove any invert fields they have added.
And we can always move it back later if it does prove too confusing.

Ill get that moved across. I shall also look into providing a helper function for upgrade scripts, to make it easier for modules to update their feedbacks. Maybe I should look into finding all the modules which implement invert fields already, so we can suggest they remove them.

PR #2569, opened 2023-09-10
Fix companion_headless.sh to work with symlinks

mdeboer:
The companion_headless.sh script was not working when it was symlinked, this fixes that 😄


Julusian:
Thanks!

PR #2573, opened 2023-09-12
feat: Creates a shortcut to buttons internal variable

riccefarias:
Allow to use a shortcut $(button:variable_name) to work with a variable isolated on button.
image

### DIFF
--- a/lib/Controls/ActionRunner.js
+++ b/lib/Controls/ActionRunner.js
@@ -110,12 +110,38 @@ export default class ActionRunner extends CoreBase {
 	 * @access private
 	 */
 	#runAction(action, extras) {
+		//console.log("action run",action,extras);
+
+		const reg = /\$\(([^:$)]+):([^)$]+)\)/g
+
+		const bank = extras.page + '_' + extras.bank
+		let actionClone = {}
+
+		let tmp = {}
+		for (var option in action.options) {
+			let value = action.options[option]
+
+			let matches
+			while ((matches = reg.exec(value))) {
+				const fullId = matches[0]
+				const instanceId = matches[1]
+				const variableId = matches[2]
+
+				if (instanceId === 'button') {
+					value = value.replace(fullId, '$(internal:custom_bank_' + bank + '_' + variableId + ')')
+				}
+			}
+
+			tmp[option] = value
+		}
+		actionClone = { ...action, ...{ options: tmp } }
+
 		if (action.instance === 'internal') {
-			this.internalModule.executeAction(action, extras)
+			this.internalModule.executeAction(actionClone, extras)
 		} else {
 			const instance = this.instance.moduleHost.getChild(action.instance)
 			if (instance) {
-				instance.actionRun(action, extras).catch((e) => {
+				instance.actionRun(actionClone, extras).catch((e) => {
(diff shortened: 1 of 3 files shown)


Julusian:
I'm not sure about this approach to the problem. I suspect it will have issues in some scenarios:
* This will not work for feedbacks
* If the reference to `$(button:inputId)` is made inside a variable it will not be parsed correctly (eg, button text is `$(internal:custom_test)` and that custom variable has  value of `$(button:inputId)`. )
* This will 'pollute' the list of custom variables with these scoped ones. It will be possible to use these as normal custom variables, which means that moving them to somewhere else later on will break peoples configs

To make it work everywhere/consistently, I think it will need to hook into the `parseVariablesInString` function (this is being done in develop/experimental for a couple of select internal actions), but that is not possible to do without changes to modules.

I am also wondering if these local variables should be persistent or if they should only exist within the execution of the action list. I feel like persistent does make sense so that you can set it in the down actions, and it still has a value for the up actions.

riccefarias:
Hi, thanks for your response...

- Yes i was forgotten to work with feedbacks.
- The ideia here is all internal button variables to be "changed"  $(button:inputId) will be $(internal:custom_bank_X_Y_inputId)
- You're right, the ideal would be to have a specific handle for the button, I thought about using CustomVariables as a base and duplicating the class to recreate the specific function for the buttons


Initially we was tested with "parseVariablesInString", but here we not have reference for button bank/page, and will require work on all submodules to fix/change its reference.

In my poor solution, we change it inside de main base classes to apply in a generally way.


I will work more on that, if you have any suggestion about using "parseVariablesInString" will be welcome.

Julusian:
In theory there is already half of the flow for the `parseVariablesInString` in place. In the call to https://github.com/bitfocus/companion/blob/master/lib/Instance/Wrapper.js#L634, msg may contain `controlId` which tells us where it is. But that relies on modules using the newer `parseVariablesInString` which only a couple do.
Providing a parameter such as https://github.com/bitfocus/companion/blob/develop/lib/Internal/Util.js#L57-L61 will inject the variables and make them available to be used.

The reason I havent implemented this myself, is the lack of module support. And because of the lack of module support, it will become even more confusing than today for users on whether a field supports variables, and what variables it will support.  
So before this is done, I think we need a way in the ui to show users what each input field supports. Perhaps some icons next to the label for the field. Such as a `$` which on hover will say `Supports variables`, and similarly something to say it supports scoped variables. Once we have these ui hints, the lack of module support doesnt bother me, as it will be clear to the user whether it should or shouldnt work.

For feedbacks, they should already be using the correct `parseVariablesInString` method internally, if they are not then they wont be updating when the variables they reference change. So it should be a lot simpler to support there.

-----

For the actual implementation of the parsing, it might be enough to modify https://github.com/bitfocus/companion/blob/master/lib/Instance/Wrapper.js#L634 to do (in pseudocode):
```
let injectedVariables = {}
if msg.controlId {
	const control = .....
   injectedVairables = control.getScopedVariables() // This method will need creating
}
``
and pass that into the `parseVariables` call.

The 'internal' module doesnt use this method, so will need a bit more work to do the same thing, ideally with some absraction to avoid it being copied and pasted around everywhere.

PR #2583, opened 2023-09-20
feat: indiciate variables support on input field

Julusian:
This is a pretty small change, that I think will help users. I am not 100% sold on the look, if anyone has ideas on how to improve it please do say

A simple indicator to show that the input field supports variables. On hover it shows `Supports variables`. It is using the `useVariables` property to decide to show this, which matches how we decide whether to suggest variables while typing.

I am hoping that we can add more icons to both this field type and others over time to indicate support for other features that modules need to implement support for

image

### DIFF
--- a/webui/src/Controls/OptionsInputField.jsx
+++ b/webui/src/Controls/OptionsInputField.jsx
@@ -132,13 +132,15 @@ export function OptionsInputField({
 		control = <CInputGroupText>Unknown type "{option.type}"</CInputGroupText>
 	}
 
+	const featureIcons = []
+	if (features.variables)
+		featureIcons.push(<FontAwesomeIcon key="variables" icon={faDollarSign} title={'Supports variables'} />)
+
 	return (
 		<CFormGroup style={{ display: visibility === false ? 'none' : null }}>
 			<CLabel>
 				{option.label}
-				{features.variables && (
-					<FontAwesomeIcon className="feature-icon" icon={faDollarSign} title={'Supports variables'} />
-				)}
+				{featureIcons.length && <span className="feature-icons">{featureIcons}</span>}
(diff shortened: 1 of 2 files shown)


MeestorX:
Can you also add this icon to dropdowns when `allowCustom` is enabled, as it allows for variables too.
Another request to PLEASE add variable support for ALL input types. Numbers, Checkboxes, etc.

Julusian:
> Can you also add this icon to dropdowns when allowCustom is enabled, as it allows for variables too.

Any suggestion on what icon should be used for `allowCustom`? 

> Another request to PLEASE add variable support for ALL input types. Numbers, Checkboxes, etc.

The architectural challenges blocking this don't have a solution yet

MeestorX:
> > Can you also add this icon to dropdowns when allowCustom is enabled, as it allows for variables too.
> 
> Any suggestion on what icon should be used for `allowCustom`?

I think the same icon would be fine.

> 
> > Another request to PLEASE add variable support for ALL input types. Numbers, Checkboxes, etc.
> 
> The architectural challenges blocking this don't have a solution yet

Copy that. NP.

dnmeid:
@MeestorX your request is already on my todo list for a long time: https://github.com/bitfocus/companion/issues/2345

Julusian:
> > > Can you also add this icon to dropdowns when allowCustom is enabled, as it allows for variables too.
> > 
> > Any suggestion on what icon should be used for allowCustom?
> 
> I think the same icon would be fine.

I'm not sure it should be the same, as there is no guarantee that `allowCustom` means that variables can be used.  
But an `allowVariables` option could be added to the dropdowns to show that icon, with a different icon shown if it supports `allowCustom`, once we figure out what that icon should be

Issue #2079, opened 2022-07-28
Formalise/tidy up the remote protocols


Julusian:
### Describe the feature

The HTTP, TCP, UDP and OSC protocols are a bit of a mess

They vary in consistency, and how well defined they are. There is no form of security or access control, some of which will need to be engrained into the protocol

Additionally, the http implementation is using names at the root level. This means that the ui has to be careful to not use the same name or it will clash and cause issues. Everything here should be moved to `/api/` or something.

This task should spend some time writing a proper definition of the protocols, and make a from scratch reimplementation of them. We should allow them to be run side by side with the old implementations (on different ports), and after a couple of releases remove the old implementation.

* Write a proper definition (something like https://github.com/bitfocus/companion/wiki/Satellite-API)
* (Optional): Should we implement some kind of api-token security in tcp, udp and http?
* (Optional): How hard would it be to do ip based allow/deny lists.

otelom:
> * Should we implement some kind of api-token security in tcp, udp and http?

@Julusian, that is exactly what I was coming here to suggest now, might be easy to implement?

Alsso, what about disabling emulator and tablet pages at all?
Or maybe add basic auth to them?
What do you think? Is it easy to implement? Should I create an issue for this feature requests?

Julusian:
@otelom please open a new issue for that

Julusian:
These apis were rewritten in 3.2, with the old apis put behind config flags.
Security and firewalling will be covered by #1925

PR #2616, opened 2023-10-20
Add a get route to the request for a value of an custom variable

uniforx:


uniforx:
There was written "You have agreed to the CLA for bitfocus/companion" at the CLA .. but the state isn´t updated here.

lib/UI/Express.js
```diff
@@ -251,6 +251,20 @@ class UIExpress extends Express {
...
+			res.header('Access-Control-Allow-Origin', '*')
+			res.header('Access-Control-Allow-Methods', 'GET,OPTIONS')
+			res.header('Access-Control-Allow-Headers', 'Content-Type, Authorization, Content-Length, X-Requested-With')
+
+			this.logger.debug(`Got HTTP /get/custom-variable/ name ${req.params.name}`)
+			const result = this.registry.instance.variable.custom.getValue(req.params.name)
+			if (result) {
+				res.send(result)
+			} else {
+				res.send('not found')
```
Julusian, line 264: This should set the return status to 404, so that the caller can know that it wasnt found
```suggestion
				res.status(404).send('not found')
```

lib/UI/Express.js
```diff
@@ -251,6 +251,20 @@ class UIExpress extends Express {
...
 		})
 
+		this.get('^/get/custom-variable/:name', (req, res) => {
+			res.header('Access-Control-Allow-Origin', '*')
+			res.header('Access-Control-Allow-Methods', 'GET,OPTIONS')
+			res.header('Access-Control-Allow-Headers', 'Content-Type, Authorization, Content-Length, X-Requested-With')
+
+			this.logger.debug(`Got HTTP /get/custom-variable/ name ${req.params.name}`)
+			const result = this.registry.instance.variable.custom.getValue(req.params.name)
+			if (result) {
```
Julusian, line 261:
```suggestion
			if (result === undefined) {
```
A variable could have a value of an empty string, which would be valid. Being `undefined` is what we want to catch here as it implies it doesnt exist

lib/Instance/CustomVariable.js
```diff
@@ -270,6 +270,15 @@ export default class InstanceCustomVariable extends CoreBase {
 		}
 	}
 
+	getValue(name) {
+		if (this.custom_variables[name]) {
+			this.logger.silly(`Get value "${name}"`)
+			return this.custom_variables[name]
+		} else {
+			return 'Unknown name'
```
Julusian, line 278: It would be better to return `undefined` here rather than a value. That will let the caller decide what to do when it doesnt exist, instead of thinking it has a value of `Unknown name`

lib/Instance/CustomVariable.js
```diff
@@ -270,6 +270,15 @@ export default class InstanceCustomVariable extends CoreBase {
 		}
 	}
 
+	getValue(name) {
+		if (this.custom_variables[name]) {
+			this.logger.silly(`Get value "${name}"`)
+			return this.custom_variables[name]
```
Julusian, line 276: I don't think this is correct for retrieving the value, this object describes the various options and details about the variable, everything except for its current value

Julusian (changes requested):

Julusian:
> There was written "You have agreed to the CLA for bitfocus/companion" at the CLA .. but the state isn´t updated here. And if I click on the badge, the message switch to already agreed too. So what can I do for the next step?

I think it is confused because the commit doesn't have an email address associated with it.
If you feel up to trying to amend the commit to have that set that would resolve it, otherwise as the name matches it is enough for you to confirm that you are the author

Julusian:
I am currently working on #2079, and have reworked this change as part of that, so it might not be worth doing anything more here

Julusian:
This is included in #2639, which will land in beta in the next few days

PR #2639, opened 2023-11-08
feat: support grid size in protocols

### LINKED ISSUE #2299: Add post request for button png backgrounds.

### Describe the feature

This is related to #2097 

I had a very bad experience using png background upload through url queries. I mostly get 431 errors (very long url request).
It is trivial to create another router endpoint (POST) just for the png. Provide there the base64 string and solve the issue.

### Usecases

It would be very easy for users to automate the change of buttons image background.

Julusian:
Replaces #2616
Fixes #2299

Extend the TCP, UDP, OSC, Emberplus and HTTP protocols to support the new button location syntax, and improves the consistency of what is available in each API.  
Additionally, this reworks the protocols to be structured in a more RESTful structure, tackling part of #2079

Also includes unit tests for the new sections. The old sections are marked as deprecated and have been moved to behind config fields to discourage their use.  
Perhaps in a few releases time companion should forcefully disable the legacy protocols (and remove documentation) as a reminder of them being removed soon, and a couple of releases after that remove those portions entirely?  
Keeping them around doesn't have much cost, but it does add noise to the docs and code


dnmeid:
I think with the current handling of colors it is possible to allow any CSS color string (except color names, it is possible to use color names for setting colors, but the color picker in the ui doesn't know how to show them).

Julusian:
@dnmeid good point. I've made sure that the one protocol which exposes that info (emberplus) will handle it, but I have left everything setting the value as a number as that is what it was doing before.
We can change it to use hex or something at the same time as elsewhere, but as we need to validate the string is valid it may as well keep doing the conversion fo rnow

davidjoshuaford:
Hey @Julusian , not sure if this is the best place to ask this, but I was wondering if there's a reason that the new button names need to begin at 0/0?

If they began at 1/1 then it would be faster to identify the total grid size (ie 4/8) when dealing with larger grids, rather than having to -1 from each row and column. 

Screenshot 2023-11-22 at 9 15 39 AM

Julusian:
> but I was wondering if there's a reason that the new button names need to begin at 0/0?

It is because they can go negative too. The idea being that you can grow the grid in every direction, and have a central 'home position' rather than only being able to grow to downwards and to the right.

davidjoshuaford:
Ok thanks!

davidjoshuaford:
To that end, is it possible for me to custom build a grid where the home position is 1/1?

PR #2655, opened 2023-11-18
feat: surface groups

### LINKED ISSUE #570: Enhancement: Grouping of Surfaces for internal actions
**Describe the feature**
The ability to assign both physical surfaces (stream decks) and virtual surfaces (web interface, emulator) to groups so that internal actions apply to all surfaces within a group.

**Usecases**
Have 2+ stream decks for one operator that would lock simultaneously and unlock as a group, while also having one stream deck that can lockout via timeout (for client use/etc.). 
Having the ability to lockout the tablet/web buttons would also be handy (to prevent unauthorized use on the network).
Another option would be to allow multiple surfaces in a group to change pages together using the stock page select buttons on any deck in the group.

A further enhancement would allow you to designate one 'Master' deck in a group that would have page buttons, and 'Slave' decks without page buttons that would respond to the master's page button presses.

### LINKED ISSUE #1642: Synced device page navigation
**Describe the feature**
Link 2 or more Stream deck devices so that a page change on one triggers a page change on the others.

**Usecases**
Useful for when pages are not in the standard navigation configuration and the nav buttons may not show on one of the devices. For example a Stream Deck XL synced with a standard or mini could change pages together.

Julusian:
Closes #570 
Closes #1642
Solves a request from @davidjoshuaford that I cannot find for being able to address surfaces by 'name'

This adds a new layer to surfaces; 'surface groups'. It does a few key things:
* A group can have any number of surfaces in it (none, one or multiple)
* All surfaces in a group will be on the same page at all times (intended to have different offsets within the page)
* All surfaces in a group will have the same locked/unlocked state at all times.
* Internal actions/feedbacks can reference these groups instead of an individual surface

This addresses #570 and #1642 as they are asking for this synced control. They may have wanted this to work with offsets or something, but as the grid can be resized now it would be a trivial (but tedious to do that many button moves) change to their configuration to arrange the buttons such that they will work.

For @davidjoshuaford's request, you can use these groups as your persistent 'names', and move surfaces in and out of them to get the behaviour desired.

Internally, every surface is now in a group (it makes the logic flow simpler), and we automatically create a group for attached surfaces which are not explicitly placed into a group. So all the internal actions/feedbacks now operate on groups rather than individual surfaces.

Q) Does this solve the same thing the 'surface by index' was to solve? (If so we should deprecate the old actions)  
I never liked the index flow because the order was not controlled by the user, so plugging in a new surface could break the order and behaviour.

Updated surfaces tab:
I am not 100% sure on how the table now looks, I think it needs work, but I am not 100% sure what
image

Surface group config:
image

Surface config showing when solo or part of a group:
image
image

Internal action referencing groups:
image


dnmeid:
> Q) Does this solve the same thing the 'surface by index' was to solve?

No, surface by index is to address users which are taking their configurations to installations with different hardware. In professional rental environments, you often get two or three streamdecks out of a rental pool and if your configuration is set to serial numbers you have to adjust every action. Also maybe your device will break and with the index programming you can replace it within seconds without any additional configuration.
At that time we were considering addressing names or index and agreed on index being faster and less configuration. The focus was more on replacing and changing surfaces but staying with the same amount of surfaces of each type. The index sorting was always emulator first, then device type alphabetically, then device serial number. So devices are naturally "grouped" by type.  
So the surface groups won't be a 1:1 replacement as you would have to manually move your surface in a group to make it work.

A solution I could imagine is, moving all freshly connected devices which never have been assigned to a group by the user automatically to a group named after the device type. E.g. all Streamdecks XL automatically get in "Streamdeck XL". If we then improove the by index actions to be "index within group" ready, we would retain this functionality and also improove it because then you would also be able to add additional surfaces without breaking your configuration logic.

Another remark about the general functionality of surface groups:
Actually the idea of the windows is to make all of this possible and in a even more dynamic way.
A window will be able to span a page over several surfaces or parts of surfaces. The difference to the surface groups is that the association to the shown source is on a geometric location basis (within window) per button and not on a logic basis (within group) per surface. Also windows will be able to be positioned and opened/closed dynamically.

So in maybe is it more useful to start with implementing the windows rather soon and not having groups at all? Or do you think it is ok to have groups now as they are nearly finished and later it is ok to have two very similar concepts?
In the long run I think organization of surfaces will be done more graphical, i.e. by moving them around on the grid. I can't estimate how long it takes for windows to be implemented and if groups will be still beneficial then or not.

Julusian:
> At that time we were considering addressing names or index and agreed on index being faster and less configuration. The focus was more on replacing and changing surfaces but staying with the same amount of surfaces of each type. The index sorting was always emulator first, then device type alphabetically, then device serial number. So devices are naturally "grouped" by type.
So the surface groups won't be a 1:1 replacement as you would have to manually move your surface in a group to make it work.

One big flaw I see in the index flow is that if you have two streamdecks in use, with serial numbers 1 and 2. 1 dies, so you swap it out for 3. Now your streamdecks are in reverse so you will need to swap their positions on you desk to get it correct again. This gets more painful the more streamdecks you have (especially so if some are connected via satellite or usb extensions).  
But I agree that in theory it is a quicker swap with no reconfiguration needed inside companion.

> A solution I could imagine is, moving all freshly connected devices which never have been assigned to a group by the user automatically to a group named after the device type. E.g. all Streamdecks XL automatically get in "Streamdeck XL". If we then improove the by index actions to be "index within group" ready, we would retain this functionality and also improove it because then you would also be able to add additional surfaces without breaking your configuration logic.

While that would resolve a small subset of cases where indices change, it wouldn't solve many of the more common ones (like the scenario I described above)
Plus, groups are intended so that all members are on one page, so without scrapping that (reverting the portions which solve  #1642) this will be very confusing to existing users.


> So in maybe is it more useful to start with implementing the windows rather soon and not having groups at all? Or do you think it is ok to have groups now as they are nearly finished and later it is ok to have two very similar concepts?

I forget about the windows concept sometimes...  
Is there anything here that you think we wouldn't be able to automatically migrate to be using windows once that is implemented?   
As windows doesnt have a timescale for being implemented (and has already been floating around as an idea for a few years) I don't think we should reject this in favour of windows.  
It feels like it should be able to translate across, but perhaps groups would still be relevant for stuff like grouped locking?

Julusian:
I can't think of any big blockers that would stop this being possible with windows later, even if every surface is disconnected, we know the location of it and the group it belongs to. If we start tracking the dimensions too then we will easily be able to generate a window per group (or however that will work), and translate any actions addressing the group to the new window.

I don't know if the synced lockouts will translate to windows, but I suspect that whole system will need revisiting at some point, and perhaps even having 'lockout groups' along with windows would make sense? I dont know, to be decided as it is done

Issue #2003, opened 2022-04-22
More than 99 Pages


Madcliff:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

Greetings, can we please get the ability to enable more than 99 pages. 

### Usecases

Companion is perhaps the most mature solution for Streamdeck on the linux Platform. I find that my usecase of the streamdeck is predominantly for expansive use of unicode with diacritics in classical languages and for macro functionalities. 
As part of this requirement, I find it helpful to be able to have increasingly complex nested pages to help me control and have intuitive access to the particular scripted commands that generate the respective unicode character.

gerbenvandekraats:
+1

davidjoshuaford:
+1 for this feature too! 99 pages is not enough... :) Does companion get sluggish or unstable if it were to exceed 99?

josephdadams:
It’s like trying to go past Warp 9. Weird stuff happens

MikeTechnical:
Hi there,
When will more than 99 pages be available?
I have 3.3 running and still don't see the pages option.
Am I missing it somewhere, or is it just not implemented yet?

Julusian:
This is now in the betas

davidjoshuaford:
Will check out warp 9!

Issue #976, opened 2020-02-02
Feature Request: Ability to rearrange existing pages


jarodwsams:
As a Companion user, I would like the ability to re-arrange pages of buttons that I've already programmed but maybe need/want in a different order.

This should be platform-agnostic.

Example Use Case:
I already have a fully-configured collection of pages, but then I add a new piece of equipment. For any number of reasons, I would like to have the control buttons for that new piece of gear closer to page 1. Ideally, I would be able to either: inject a new blank page between pages 1 and 2, or move another page to the number 2 spot (pushing the original page 2 to page 3, and so on).

josephdadams:
This sounds like #610, but since yours is so much more thoughtfully written out, I'm going to close that one and keep this one. :)

Hennyr:
Josephadams,
What is the status of issue/request #976?  I am looking for this feature. I am very pleased with companion.

poucedeleon:
Joseph, you commented to the status of 976 "not Implemented", does that mean still being considered or the request was turned down?

Julusian:
@poucedeleon it will be done, and is waiting for some larger work on being able to change the number of pages to be done

metawops:
I'm also looking forward to this feature. Any news? Will it be implemented on the 2.x branch (which I'm still using) or will it be a feature of Companion 3.x? 😳 Thanks so much! 😊

Julusian:
This will probably be in 3.1, and definitely wont be in an earlier release

istnv:
You can also do a full export, import a single page with a different destination page.

Julusian:
This is now in the betas

PR #2689, opened 2024-01-07
feat: more than 99 pages #2003

Julusian:
Support more or less than 99 pages.

Fresh companion installations will start with 1 page, it is intended that users add as many pages as they need. Existing installations are unchanged, to avoid risking impacting users without them expecting it.

There is a new 'pages' tab in the right panel, which lists the pages with links to jump to a page, as well as to insert/add a page (at any position), and remove any page.
image

When adding a page, a modal is shown, to allow easily adding multiple and to encourage giving each new page a name

image

If when removing a page, a surface ends up out of bounds, it will be pushed to be within the new page count.

### TODO:
- [x] Importing a single page needs to be able to import to a new page (potentially just at the end for now?)
- [x] When adding a page, open a prompt to ask how many pages and the name to give each page

### Future:
* Some work should be done to make surfaces more sticky to pages when they are removed/reordered
* The actions/feedbacks should be improved to be able to reference a page by id. this is so that as the index of a page changes, actions and feedbacks can 'follow' that moving. (Import will need to take care to fix up these ids properly)
*  [Feature Request: Ability to rearrange existing pages #976](https://github.com/bitfocus/companion/issues/976)

### DIFF
--- a/docs/3_config/buttons.md
+++ b/docs/3_config/buttons.md
@@ -2,7 +2,7 @@ In the Buttons tab, you can configure the buttons for your Stream Deck.
 
 ![Buttons Page](images/buttons.png?raw=true 'Buttons Page')
 
-The Buttons layout has 99 pages that can be navigated using the grey left/right arrows, or the dropdown box. You can give your pages a unique name by replacing the word _PAGE_ right next to the page number.
+The Buttons layout has numerous pages that can be navigated using the grey left/right arrows, or the dropdown box. You can give your pages a unique name by replacing the word _PAGE_ right next to the page number.
 
 If you hold down the SHIFT key on your keyboard, you can trigger a button directly by clicking on it.
 
--- a/lib/Controls/Controller.js
+++ b/lib/Controls/Controller.js
@@ -1593,6 +1593,7 @@ class ControlsController extends CoreBase {
 
 	/**
 	 * Create a control
+	 * Danger: This will not delete an existing control from the specified location
 	 * @param {import('../Resources/Util.js').ControlLocation} location Location to place in the grid
 	 * @param {string} newType The type of the new control to create (if any)
 	 * @returns {string | null} controlId
--- a/lib/Data/ImportExport.js
+++ b/lib/Data/ImportExport.js
@@ -243,7 +243,7 @@ class DataImportExport extends CoreBase {
 			if (isNaN(page)) {
 				next()
 			} else {
-				const pageInfo = this.page.getPage(page, true)
+				const pageInfo = this.page.getPageInfo(page, true)
 				if (!pageInfo) throw new Error(`Page "${page}" not found!`)
 
 				const referencedConnectionIds = new Set()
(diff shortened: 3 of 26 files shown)


bryce-seifert:
@Julusian Just played around with this, seems to be working well. I'm not sure if it's in the scope of this pull, but the new Pages tab has me thinking about the following:

- Ability to bulk add pages. IE add 10 new pages at a time as opposed to individually, especially if all new instances are starting at 1
- Ability to reorder pages https://github.com/bitfocus/companion/issues/976 I feel like that could potentially be worked into this new Pages tab with up/down arrows?

And less of a priority, bit I'm wondering if there's a way to add a more obvious add / delete animation on the pages tab. It sort of pops on currently, but I think a quick fade in / out, with green or red background/text could help see exactly where a page was added/removed

Julusian:
> @Julusian Just played around with this, seems to be working well. 

Thanks for trying it, Im glad it is working for you :)

> * Ability to bulk add pages. IE add 10 new pages at a time as opposed to individually, especially if all new instances are starting at 1

Yeah that would be reasonable. I guess a button next to the existing add at end button (or the insert ones too?) which will open a prompt. or should all the existing buttons prompt for how many to add?

> * Ability to reorder pages [Feature Request: Ability to rearrange existing pages #976](https://github.com/bitfocus/companion/issues/976) I feel like that could potentially be worked into this new Pages tab with up/down arrows?

Yeah, I'm leaving that out of this to try and avoid this PR becoming huge. It is one of the next things to do though

> And less of a priority, bit I'm wondering if there's a way to add a more obvious add / delete animation on the pages tab. It sort of pops on currently, but I think a quick fade in / out, with green or red background/text could help see exactly where a page was added/removed

that sounds beyond my ability of creating something that will look acceptable :laughing:  

but this does make me wonder, perhaps adding a page should always open a modal and it should prompt the user to give the new pages names. In the past the page names haven't been used for all that much (although their use has been growing gradually over time), perhaps we should be nudging the user to give them useful names?   
Then at the same time it would be easy to prompt the user to choose how many pages to insert

bryce-seifert:
>  I guess a button next to the existing add at end button (or the insert ones too?) which will open a prompt. or should all the existing buttons prompt for how many to add?

My thought is using the existing button(s) with prompt, and if we page names are added there it could streamline that setup

Julusian:
After leaving this alone for too long, I am satisfied that this chunk is complete.
There is still more to do to improve the support for dynamic page counts (see the 'future' heading in the description), which shall follow and should be done before this makes it into the betas

smudge1977:
I'm not a Node or js person but it looks like the API still all runs on numbers for pages which I would guess would be the order pages are added in rather than been the name - so for example the common bank-press - I'm using UDP broadcasts to do the work:
https://github.com/bitfocus/companion/blob/13d9a2a738b287ef142ac615b4d916f96eabe065/companion/lib/Service/TcpUdpApi.js#L112

Julusian:
@smudge1977 The api's have not been changed for this currently, I'm not sure how those changes should look currently.
But it will respect the order they are shown in the ui and the rest of the system

reidwwall:
Hello,
I Cannot seem to find this feature in companion now.  I have tried looking in 3.2.2 AND and 3.3.0.  Am I missing something?

Thank You!

dnmeid:
> Hello, I Cannot seem to find this feature in companion now. I have tried looking in 3.2.2 AND and 3.3.0. Am I missing something?

@reidwwall It is in the development branch, that means the experimental builds.

PR #2743, opened 2024-02-12
Fix typos and missing operators.

joezimjs:
For a long time I thought you didn't have a ternary operator because it wasn't listed here. And since I was in here adding it, I fixed a couple typos too.

I know, I know. If I had just read the things above this list, I would have realized it, but I'm a skimmer.


Julusian:
Thanks!

Issue #2559, opened 2023-08-27
add function "split" to variable expressions


Fred-DTV:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

I am receiving an array of numbers as a variable from obs and was trying to split them into single variables with the js split function, but the companion log says that this function is not yet implemented.

Would be great to see this added.

### Usecases

see above

Julusian:
I'm not sure if the expression syntax currently allows for arrays at all. That will need to be verified/implemented at the same time as doing this

Fred-DTV:
Thanks :)

Julusian:
This is now possible in the betas

Coupe70:
Sorry for trying it in this old post, but maybe I can reach out to the experts here...
I have a hard time finding out how to use arrays in Companion as there is no documentation or example.
I know how to use 'split' in a programming language, but when I try split("one#two#three", "#") in Companion 
my custom variable just contains the string "one,two,three" afterwards. So yes, it splits, but I don't get an array
to access the single elements.
How do I set up an array in Companion (and split to it) and how do I access an element of an array afterwards?
Probably the latter will work as I would expect as soon as I have a proper array...

PR #2782, opened 2024-03-02
feat: Add expression array join function

peternewman:
To go with the split one requested in #2559

### DIFF
--- a/shared-lib/lib/Expression/ExpressionFunctions.js
+++ b/shared-lib/lib/Expression/ExpressionFunctions.js
@@ -40,6 +40,9 @@ export const ExpressionFunctions = {
 	split: (str, separator) => {
 		return (str + '').split(separator)
 	},
+	join: (arr, separator = ',') => {
+		return arr.join(separator)
+	},
 	concat: (...strs) => ''.concat(...strs),
 	includes: (str, arg) => {
 		return (str + '').includes(arg)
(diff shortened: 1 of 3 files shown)


shared-lib/lib/Expression/ExpressionFunctions.js
```diff
@@ -40,6 +40,9 @@ export const ExpressionFunctions = {
 	split: (str, separator) => {
 		return (str + '').split(separator)
 	},
+	join: (arr = [], separator = ',') => {
+		return (arr.constructor === Array ? arr : [arr]).join(separator)
```
Julusian, line 44: I think this could use `Array.isArray(arr)` to check if it is an array

Julusian (approved):

shared-lib/lib/Expression/ExpressionFunctions.js
peternewman: This sounded like it thought the equality was faster, but I'm happy to switch for readability if you'd prefer?
https://stackoverflow.com/a/26633883

Julusian: I'm not sure that we should be worried about speed here, these expression executor is going to be rather suboptimal by design (its an interpreter running inside an interpreter/jit)
But I don't feel that strongly about how this should be done. you thought this was better, which is enough for me

peternewman: Heh very true. Maybe readability should have won out, feel free to just push a change. There didn't seem to be much else in it compatibility wise etc.

Issue #2605, opened 2023-10-15
Add Custom Names to Step Tabs


davidjoshuaford:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

I have some buttons that have many Steps – it is difficult to recall the sequence of events when they are just numbered.

It would be helpful to be able to add a custom name to each tab number so that parameters that have been built in the past are easy to understand when you come back to the button later on. 

Bonus if these custom names were searchable and could be referenced/ triggered. 

<img width="443" alt="Screenshot 2023-10-15 at 11 59 54 AM" src="https://github.com/bitfocus/companion/assets/76959770/535abeee-689f-40b0-8c0a-15f1cf2d64f7">

Stagge:
Hi,

We are a group of 5 students at KTH whom would like to solve this issue as part of a course in Software Engineering!

@Pqformeln @FelixSjogren @Stagge @isadorawinter @BDidace

PR #2783, opened 2024-03-04
feat: Custom names for Step Tabs

Stagge:
This is work related to #2605.

The Tabs for Steps when editing a button can now get custom names by double clicking the tab to edit name. 

More work could be done with persisting the names. When the application in restarted the names of the steps dissapear. This is probably because the backend is not updated correctly.

### DIFF
--- a/companion/lib/Controls/Fragments/FragmentActions.js
+++ b/companion/lib/Controls/Fragments/FragmentActions.js
@@ -43,6 +43,12 @@ export default class FragmentActions extends CoreBase {
 	options
 
 	/**
+	 * @type {string | undefined}
+	 * @access public
+	 */
+	name
+
+	/**
 	 * Commit changes to the database and disk
 	 * @type {(redraw?: boolean) => void}
 	 * @access private
@@ -551,5 +557,15 @@ export default class FragmentActions extends CoreBase {
 		}
 
 		return changed
+	}
+
+	/**
+	 * Rename this control
+	 * @param {string} newName the new name
+	 * @returns {void}
+	 * @access public
+	 */
+	rename(newName) {
+		this.name = newName
 	}
 }
--- a/companion/lib/Controls/Fragments/FragmentActions.js
+++ b/companion/lib/Controls/Fragments/FragmentActions.js
@@ -43,12 +43,6 @@ export default class FragmentActions extends CoreBase {
 	options
 
 	/**
-	 * @type {string | undefined}
-	 * @access public
-	 */
-	name
-
-	/**
 	 * Commit changes to the database and disk
 	 * @type {(redraw?: boolean) => void}
 	 * @access private
@@ -566,6 +560,6 @@ export default class FragmentActions extends CoreBase {
 	 * @access public
 	 */
 	rename(newName) {
-		this.name = newName
+		this.options.name = newName
 	}
 }
--- a/webui/src/Buttons/EditButton.tsx
+++ b/webui/src/Buttons/EditButton.tsx
@@ -40,11 +40,12 @@ import React, {
 	useState,
 	useMemo,
 	FormEvent,
+	useReducer,
 } from 'react'
 import { nanoid } from 'nanoid'
 import { ButtonPreviewBase } from '../Components/ButtonPreview.js'
 import { GenericConfirmModal, GenericConfirmModalRef } from '../Components/GenericConfirmModal.js'
-import { KeyReceiver, LoadingRetryOrError, socketEmitPromise, SocketContext, MyErrorBoundary } from '../util.js'
+import { KeyReceiver, LoadingRetryOrError, socketEmitPromise, SocketContext, MyErrorBoundary, CompanionSocketType } from '../util.js'
 import { ControlActionSetEditor } from '../Controls/ActionSetEditor.js'
 import jsonPatch, { Operation as JsonPatchOperation } from 'fast-json-patch'
 import { ButtonStyleConfig } from '../Controls/ButtonStyleConfig.js'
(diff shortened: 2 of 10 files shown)


webui/src/Buttons/EditButton.tsx
```diff
@@ -506,17 +568,15 @@ function TabsSection({ style, controlId, location, steps, runtimeProps, rotaryAc
...
 									else if (isCurrent) linkClassname = 'only-current'
 								}
 
 								return (
-									<CNavItem key={k} className="nav-steps-special">
-										<CNavLink data-tab={`step:${k}`} className={linkClassname}>
-											{i === 0 ? (keys.length > 1 ? 'Step ' + (i + 1) : 'Actions') : i + 1}
-										</CNavLink>
-									</CNavItem>
+									<TabStep selectedKey={selectedKey} k={k} i={i} name={name} inputState={{showInputEle:false}} linkClassname={linkClassname} socket={socket} controlId={controlId} ></TabStep>
```
Julusian: the `key` property should remain at this level, rather than being pushed inside. It is an 'id' field for react to handle arrays/lists of similar components https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

webui/src/Buttons/EditButton.tsx
```diff
@@ -372,6 +373,59 @@ interface TabsSectionProps {
...
+	)
+	
+	function onChangeHandler(e: React.ChangeEvent<HTMLInputElement>){
+		setState({showInputEle: state.showInputEle});
+		if(selectedKey){
+			renameStep(selectedKey, e.target.value);
+		}
+	}
+
+	const [state, setState] = useState(inputState);
```
Julusian: I think this state could be simplified to be a single boolean value, with a more descriptive name `const [showInputEle, .....`. The default value also doesn't make sense to be a prop to this component, as it always gets passed the same value.

webui/src/Buttons/EditButton.tsx
```diff
@@ -372,6 +373,59 @@ interface TabsSectionProps {
...
+					onChange={(e)=>{onChangeHandler(e)}}
+					onKeyDown={(e)=>{e.key === 'Enter' && setState({showInputEle: false})}}
+					onBlur={(e)=>setState({showInputEle: false})}
+					autoFocus
+				>			
+				</input>
+				</CNavLink>
+			) : (
+				<CNavLink onDoubleClick={(e)=>setState({showInputEle: true})} data-tab={`step:${k}`} className={linkClassname}>
+					{name && name !== "" ? name : (i === 0 ? 'Step ' + (i + 1) : i + 1)}
```
Julusian: Would it make sense to do `name (${i + 1})` when there is a name set?
I suspect that if you have something like 8 named steps, figuring out which is step 4 will be tedious without us keeping the number at the end

webui/src/Buttons/EditButton.tsx
```diff
@@ -372,6 +373,59 @@ interface TabsSectionProps {
...
+					value={name}
+					onChange={(e)=>{onChangeHandler(e)}}
+					onKeyDown={(e)=>{e.key === 'Enter' && setState({showInputEle: false})}}
+					onBlur={(e)=>setState({showInputEle: false})}
+					autoFocus
+				>			
+				</input>
+				</CNavLink>
+			) : (
+				<CNavLink onDoubleClick={(e)=>setState({showInputEle: true})} data-tab={`step:${k}`} className={linkClassname}>
```
Julusian: This is something for me to think on rather than something for you to resolve 

We don't currently do double click to edit anywhere, It might be good to have some visual hint to help users discover this. Perhaps a `title` would be enough?
Double click to edit could also make sense elsewhere too.

Or discoverability could be solved by showing an edit pencil for just the current step? Would that look too busy?

companion/lib/Controls/Fragments/FragmentActions.js
```diff
@@ -42,6 +42,12 @@ export default class FragmentActions extends CoreBase {
 	 */
 	options
 
+	/**
+	 * @type {string | undefined}
+	 * @access public
+	 */
+	name
```
Julusian: I wonder if this would make more sense as a property inside of the `options` object above?
I think that would solve the issue of it being forgotten. The real issue with that is that the loading flow for properties on this class is non-obvious, and it happens https://github.com/bitfocus/companion/blob/da425be3a3ba07c0d543220149edde7d2d58c0ab/companion/lib/Controls/ControlTypes/Button/Normal.js#L602. The other solution to it not being loaded at startup would be to add a similar loading line there

webui/src/Buttons/EditButton.tsx
```diff
@@ -372,6 +373,59 @@ interface TabsSectionProps {
...
+
+	const [state, setState] = useState(inputState);
+	return (
+		<CNavItem key={k} className="nav-steps-special">
+			{state.showInputEle? (
+				<CNavLink className={linkClassname}>
+				<input
+					type="text"
+					value={name}
+					onChange={(e)=>{onChangeHandler(e)}}
```
Julusian: It would be nice to see these functions wrapped in `useCallback` like the one above is. That will help ensure react is performant by not passing it new functions each time it renders.
Saying that, it is probably overkill to do somewhere like this

webui/src/Buttons/EditButton.tsx
```diff
@@ -372,6 +373,59 @@ interface TabsSectionProps {
...
+				console.error('Failed to rename step:', e)
+			})
+		},
+		[socket, controlId]
+	)
+	
+	function onChangeHandler(e: React.ChangeEvent<HTMLInputElement>){
+		setState({showInputEle: state.showInputEle});
+		if(selectedKey){
+			renameStep(selectedKey, e.target.value);
```
Julusian: The use of `selectedKey` seems strange here. Shouldn't it always be renaming `k`, as that is what this component is?
I know that 99% of the time it should only be possible to open the editor when `selectedKey` and `k` are the same value, but it would simplify things to use `k` to confirm the intent

Julusian (commented):
Thanks for doing this, this is looking good!

I have a few nitpicking comments/suggestions that I hope make sense. Let me know if you have any questions, I've tried to give reasons for everything

Could you provide a screenshot of how it looks while editing the name of a step? I'm curious how it looks

Stagge:
Thanks for the review and the great feedback!

I've implemented your suggestions. The name property is now under "options", and names are no longer forgotten :)
There is still no visual hints for double clicking though, but I do agree that it might be needed.

Take a peek!
Screenshot 2024-03-07 at 21 29 35

Julusian:
Thanks, this looks good to me :)

PR #2784, opened 2024-03-04
[WIP] Feat/issue 2581/list of variables are used

tgadler:
Fixes issue 2581 for CustomVariables. Left to do is to extend it to also include other Variables. Furthermore I would appreciate some feedback on the layout.


Julusian:
This needs some work before it can be merged, and as the author hasn't responded to the request I am going to close this for now.

We are not opposed to this idea, so if someone is willing to address our concerns they are welcome to open a new PR

Issue #2714, opened 2024-01-24
[BUG] Waiting for Webserver and high RAM use


Cdonhou:
### Is this a bug in companion itself or a module?

- [X] I believe this to be a bug in companion

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Installed 3.2.0 over existing working 3.1.2 install. After install the app launches but it hangs at “waiting on web server”. It never loads and clicking any item in the app window causes the system to spike and use almost all available RAM. Uninstall and reinstall same version doesn’t work. Did an uninstall and rolled back to 3.1.2 and everything works again

I

### Environment (please complete the following information)

```markdown
- OS:windows 10
- Browser:
- Companion Version:3.2.0
```

Julusian:
could you export your config and sent it to me? git@julusian.co.uk
It launches fine on my windows machine, perhaps it is something specific to certain configs/modules

Cdonhou:
I can but the only module is a Black Magic hyperdeck mini HD.

Julusian:
hmm ok. still, it could be something in there that isnt being handled correctly.

I should give the release a try on some other windows machines too , maybe there is something special about mine that makes it work

Cdonhou:
I’m not the only person with the issue. There another report on the slack channel.

giorgiodutto:
Same for me, i've tested on many pc also without upgrade but directly the 3.2 version.
the log speak about fonts......
the problem is that if i unistall the 3.2 ,to reinstall the old version, alsa the old version don't work.
the temporary solution for me was to install the old version in a clean machine

Julusian:
thanks for the config, which unsurprisingly did not help to reproduce the issue.
I've tried 3.2 on 4 windows machines now, none of them had any issues with running it.

Could you send the log files? That might have some more info as to what is going wrong. There will be a folder of them at `%APPDATA%/companion/logs` I think

premultiply:
I also have got a Windows 10 LTSC system where this issue occures.

It almost instantly takes gigabytes of RAM but never come up completly.

@Julusian Maybe I can offer you a remote access to the system for further analysis if it would help.

giorgiodutto:
![2](https://github.com/bitfocus/companion/assets/99798875/c7ffa948-d185-4319-977f-a074d1bb5651)
![1](https://github.com/bitfocus/companion/assets/99798875/8d115df1-534d-41ea-b808-bd31975f4b30)

[...]

PR #2785, opened 2024-03-04
fix: change canvas lib

Julusian:
~~I feel like this should go to 3.3, but it might be needed to fix some bugs in 3.2~~
This is now planned to land in 3.4 immediately after the release of 3.3

The skia-canvas lib we are using is problematic. There was a bug where loading the emoji font on windows could consume all available memory and crash #2714 , and another bug where it (also on windows) never returns when rendering certain unicode characters.

Also while at one point the library looked promising, it now looks to be abandoned. While there might be some viable forks of that library, this proposed change is definitely maintained better. (Long term, we may want to consider `canvas`, but that currently wont work due to not being node-api based yet).

This needs some more testing, to verify it doesn't have the same flaws on windows.
And some testing to verify the drawing in more scenarios. My quick testing shows it to be something like a pixel off sometimes (pixel rounding?), but is otherwise identical to 3.2.2


dnmeid:
When I started the canvas stuff, I actually tried first with napi-rs/canvas, but unfortunately I couldn't get it to work by then. Also skia-canvas has some nice features on top of skia which napi-rs/canvas is missing. Mainly I was interested in the built in word-wrap/multiline support.
At the end I found that the word wrap didn't had the flexibility we need with some custom valid line breaking characters and I stayed with own solution. So if now skia-canvas is loosing maintanance and napi-rs/canvas does work, I'm fine with the change. 

Regarding the pixel preciceness, with skia-canvas I had often to position lines at .5 positions to fit exactly one pixel. This is also the correct way like it is in skia.

I have now a Windows machine at hand and can give the PR a try in the next few days.

Julusian:
So far I haven't done any refinement other than fixing one very noticable issue with line heights, which was a simple swapping of some property names.
My main interest is whether it has the same 'fatal' bugs as the previous lib did, I will need to ask someone who had those issues previously to test and verify.

And I haven't confirmed that this works in the built files that we distribute yet, but I don't expect that to be impossible, just needs some config tweaking to get setup

Issue #2820, opened 2024-03-27
GET value of a custom-variable via new HTTP API


LHasenbein:
### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

I have tested the Experimental 3.99.0+6806. I want to GET the value of a custom-variable via HTTP API like it's described in the doc. I've tried it with Javascript and php. I always get an html page back:

<!doctype html>
<html lang=\"en\">
	<head>
		<meta charset=\"utf-8\" />

		<link href=\"/img/icons/256x256.png\" rel=\"shortcut icon\" />

		<link rel=\"manifest\" href=\"/manifest.json\" />
		
		<meta name=\"theme-color\" content=\"#000000\" />
		
		<link rel=\"apple-touch-icon\" sizes=\"16x16\" href=\"/img/icons/16x16.png\" />
		<link rel=\"apple-touch-icon\" sizes=\"32x32\" href=\"/img/icons/32x32.png\" />
		<link rel=\"apple-touch-icon\" sizes=\"48x48\" href=\"/img/icons/48x48.png\" />
		<link rel=\"apple-touch-icon\" sizes=\"128x128\" href=\"/img/icons/128x128.png\" />
		<link rel=\"apple-touch-icon\" sizes=\"256x256\" href=\"/img/icons/256x256.png\" />

		<meta name=\"viewport\" content=\"width=device-width, initial-scale=1.0 , maximum-scale=1.0, user-scalable=0\" />
		<meta name=\"language\" content=\"en\" />
		<meta name=\"google\" content=\"notranslate\" />

		<meta http-equiv=\"X-UA-Compatible\" content=\"IE=edge\" />
		<meta http-equiv=\"Content-Language\" content=\"en\" />

		<meta name=\"theme-color\" content=\"#000000\" />
		<meta name=\"description\" content=\"Bitfocus Companion - Admin\" />
		<title>Bitfocus Companion - Admin</title>
		<script type=\"module\" crossorigin src=\"/assets/index-CbmXG7xe.js\"></script>
		<link rel=\"stylesheet\" crossorigin href=\"/assets/index-TlWH-Hzx.css\">
	</head>

	<body>
		<noscript>You need to enable JavaScript to run Companion.</noscript>
		<div id=\"root\"></div>
	</body>
</html>

Maybe it's my fault. But I tried it as described in the documentation.

### Environment (please complete the following information)

[...]

PR #2821, opened 2024-03-29
fix: routes precedence #2820

dnmeid:
So far the express routes of the new http API have been pushed to the end of the routes of express app. This results in bug #2820 where the default catch all get requests and return index is served before the get of the api.
This PR changes the routing a little bit:
The api router is invoked with a base url from the app, so all urls are relative to the base url there. The router itself does not define a base url any more.
Legacy API routes and the new API routes are set up in one shared router. The router is added dynamically to the express app and can be swapped at runtime to make the logic consistent with the controller.
The new api routes are now parsed before the legacy routes.
The catch all 404 route for the api is not a part of the legacy or the new api routes any more but always at the end of all api routes.

### DIFF
--- a/companion/lib/Service/Controller.js
+++ b/companion/lib/Service/Controller.js
@@ -38,8 +38,7 @@ class ServiceController {
 	 * @param {import('../Registry.js').default} registry - the application core
 	 */
 	constructor(registry) {
-		this.httpApi = new ServiceHttpApi(registry, registry.ui.express.legacyApiRouter)
-		this.httpApi.bindToApp(registry.ui.express.app)
+		this.httpApi = new ServiceHttpApi(registry, registry.ui.express)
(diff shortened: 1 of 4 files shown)


dnmeid:
Ah, I see the tests are failing. Didn't remember that there are unit tests for the api and they are broken now.

Julusian:
@dnmeid This has broken the legacy/old http api, it has moved those old urls to be under `/api` instead.

> Ah, I see the tests are failing. Didn't remember that there are unit tests for the api and they are broken now.

Yeah, I wrote some unit tests for the new api while writing it, as part of a gentle nudge into getting better test coverage. :)

dnmeid:
Wait, our old api was not under `/api`. Crazy. Ok, I'll fix it.

PR #2824, opened 2024-04-02
fix: attach legacy http api at root

dnmeid:


dnmeid:
I also added at least one unit test for the legacy http api, didn't want to put too much effort in it.

Issue #2498, opened 2023-06-22
Feature Request - Select a custom color for GUI's Menu bar color


davidjoshuaford:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

Hopefully an easy feature to implement in the Settings page: I would like to be able to select a custom color to replace the red in the title menu bar in the GUI: 

<img width="935" alt="Screenshot 2023-06-22 at 5 50 00 PM" src="https://github.com/bitfocus/companion/assets/76959770/3e5c0827-42b1-48e1-aff6-3270972dbce2">

### Usecases

When using multiple companion servers on the same network, it's easy to edit the wrong instance if you have multiple tabs open and don't check the IP. 

Being able to set the top bar color would provide quick visual feedback that you're working on the right one. 

For instance, my main companion installation could remain red, but I could set my CompanionPi color to purple, so that if I flick between pages, it's immediately apparent which instance I am working on.

Julusian:
At the same time as doing this, it would be good to also have a customisable 'name' shown where the version is. (perhaps version should move to the right side of that space instead?)

Making that colour customisable is very doable, but will require some changes to how the page loads, otherwise it will be red until the 'syncing' is complete then will switch to the target colour. Not terrible if it isnt fixed, but not ideal

davidjoshuaford:
Great idea - custom name would be helpful, particularly in team environments to communicate clearly.

dnmeid:
I back the idea of a customisable server name.
For the color part I'd suggest an adjustable theme, this would serve more purposes then just tinting the bar on top.

PR #2866, opened 2024-05-18
feat: add installName userconfig

### LINKED ISSUE #2813: Allow changing of browser page/tab title

### Describe the feature

Allow the browser page/tab title of the admin UI to be edited via the admin UI.
Perhaps use the hostname of the machine companion is running on and/or make the title editable.

### Usecases

In a situation where multiple instances of companion admin are open in different tabs this would make identifying the desired instance easier.

krocheck:
This adds a `userconfig` field called `installName` that defaults to "Bitfocus Companion" but when modified changes the browser titles and adds the name to the left of the version in the Admin header.

The functionality can be useful in larger show networks when users might be configuring or utilizing Companion on more than one machine.

image

Closes #2813 and partially responds to #2498

TODO:
- [ ] Add documentation


Julusian (approved):
I'm wondering what the default value should be.
Do we need to bake in 'Bitfocus Companion' as the default value, or would it be better to have an empty string, with that default inserted at the place of using it?
I've never been that keen on how the default page name is 'PAGE', giving that value some special properties.

Part of me wants this when unset to use the hostname of the machine, but I'm not sure that would be a good idea

krocheck:
I can switch this to be empty by default and replace/override 'Bitfocus Companion' in the title bar instead if that's preferred.

webui/src/Emulator/Emulator.tsx
```diff
@@ -22,6 +22,8 @@ import {
 	EmulatorImageCache,
 } from '@companion-app/shared/Model/Common.js'
 import { Operation as JsonPatchOperation } from 'fast-json-patch'
+import { UserConfigStore } from '../Stores/UserConfigStore.js'
+import { useUserConfigSubscription } from '../Hooks/useUserConfigSubscription.js'
 
 export function Emulator() {
```
Julusian:
```suggestion
export const Emulator = observer(function Emulator() {
```

This needs wrapping in observer, so that it re-renders whenever the watched properties change

webui/src/Emulator/Emulator.tsx
```diff
@@ -65,6 +67,13 @@ export function Emulator() {
 		}
 	}, [retryToken, socket, emulatorId])
 
+	const userConfigStore = useMemo(() => new UserConfigStore(), [])
+	useUserConfigSubscription(socket, userConfigStore)
+
+	useEffect(() => {
+		document.title = `${userConfigStore?.properties?.installName} - Emulator`
+	}, [userConfigStore])
```
Julusian:
```suggestion
	}, [userConfigStore.properties?.installName])
```

the `userConfigStore` reference won't ever change after the first render, so this useEffect needs to be more specific and target the property it needs instead.

an alternate option would be to do something more mobx based so that the `useEffect` gets re-run whenever the observable changes, but thats fiddly to do

krocheck:
@Julusian I think this is ready to go now if you want to give a fresh look.

Julusian (approved):

PR #2867, opened 2024-05-18
v3.4: Wizard Update

krocheck:
This adds Videohub Panel listener enable to the Surfaces tab and a new Button Grid size tab.  It also adds a couple UI bits to the Settings tab to show the total number of rows and columns in the grid and cleans up the design of the Surfaces tab.

image
image
image
image


haakonnessjoen:
> From Keith:
> Here's an update.  I changed the logic so it detects if minRow or minCol are not 0 and bypasses the step in that case.
>
> image

Looks good, but I think it would be nice to add some text to clarify that it can be changed later. Or else the user might think this is something that has to be decided right as they see this screen.

krocheck:
@haakonnessjoen 
image

Issue #2812, opened 2024-03-20
Improve variable naming consistency


mark-epstein:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

Depending where you are in Companion a custom variable is referred to as 
`$(internal:custom_myVar)` {variable substitutions}
`A custom variable (internal:custom_myVar)` {trigger events, conditions}
`A custom variable (myVar)` {trigger actions}

I would prefer that they were all a bit more consistent (and as brief as possible without introducing ambiguity. I'd suggest:
`$(custom:myVar)` for the substitution one and
`A custom variable (myVar)` for everywhere that's a drop-down option

The "internal" designation seems redundant in all cases because it appears that all custom variables, by default, are internal.

### Usecases

just makes things a bit cleaner, easier to read

dnmeid:
That's not going to happen like you suggest. There is a logic behind variable naming and the syntax of using variable names.
Every variable is issued by a module, that is the part in front of the `:`. The module internal offers some variables which are defined by the application and the user can add custom variables. To make them distinguishable custom variables start always with `custom_`. So a complete custom variable needs all the parts internal:custom_myVar
If you want to address a variable it depends on the context of where you are what you need to enter. E.g. in a text environment we need to distinguish between the literal text internal:custom_myVar and the call to that variable. This is done by surrounding variable names with $(). If we offer a dropdown, an action or an API call where only custom variable names are allowed we can shorten it to the last part as all custom variables are issued by internal and start with custom_
I admit that this sometimes is not intuitive at first and maybe there could be improvements in GUI and/or documentation to make variable selection more easy.
If you have any suggestions how we can improve here, please feel free.

I also admit that the naming syntax itself is not the best stuff. 
In expressions you have to use variables with $() again just for consistency. Then there is the template string syntax with ${} which can be confusing with $(). As of today I would prefer being able to use variables in expressions without $() and also I would prefer the js-like notation of `internal.custom.myVar`. This would give modules a standardized way of grouping variables in a hierachical tree. Also if we would have variable replacement in expressions without $(), it would be good to replace the $()-notation in text with ${} notation. By doing so, we could make the standard text entry a template string and give it more functionality and more consistency with an expression at the same time. 
[...]

PR #2868, opened 2024-05-18
feat: move custom variables to $(custom:...) with backward compatibility

krocheck:
Responds, in part to #2812 by moving custom variables from `$(internal:custom_...)` to `$(custom:...)`.  Checks are added to maintain backwards compatibility with legacy names.

_If we want to move forward I'll get the rest of the action items executed._ 

My main reason for attempting this is that with the addition of `$(this:...)`, it seems logical to me to make this change.  The fact that `custom` isn't a reserved word in v2.x is concerning, but I think is quite the edge case for someone to completely rename an instance to `custom`.  But that can be investigated for upgrade mitigation ... though scrubbing the variables might become difficult.

TODO:

- [x] Validate basic functionality and backwards compatibility
- [ ] Validate custom variable internal actions
- [ ] Validate triggers
- [ ] Update documentation
- [ ] Update any labels, like indicated in #2812
- [ ] Investigate `custom` instance name mitigation in v2-v3 upgrade

### DIFF
--- a/companion/lib/Instance/CustomVariable.js
+++ b/companion/lib/Instance/CustomVariable.js
@@ -430,7 +430,7 @@ export default class InstanceCustomVariable {
 	 */
 	syncValueToDefault(name) {
 		if (this.#custom_variables[name]) {
-			const value = this.#base.getVariableValue('custom', fullname)
+			const value = this.#base.getVariableValue('custom', name)
 			this.#logger.silly(`Set default value "${name}":${value}`)
 			this.#custom_variables[name].defaultValue = value ?? ''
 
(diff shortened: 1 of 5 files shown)


Julusian:
I have no problem with this being done.
Perhaps for a release or two (or forever?) we should keep the `internal:custom_` around for compatibility? But without the definition, so that they won't show up in the ui.
They could be implemented so that `internal:custom_X` has a value of `$(custom:X)`, to simplify setting the values (we use this 'trick' elsewhere, to get the `$(this:page_name)` and `$(this:step)` to be correctly reactive)

PR #2880, opened 2024-05-24
feat: Time Expression Functions

thedist:
Adds `timeOffset` and `timeDiff` Expression functions, as well as change the Expression Function documentation to create a `Time operations` section to group up these, and the previous time related functions.

`timeOffset` allows taking in a timestamp an offset, and optionally a 12 hour clock boolean, and outputting the time with that offset in the same format it came in with (eg, if hh:mm:ss is provided, then hh:mm:ss is returned, if seconds are omitted then likewise in the return value).

`timeDiff` compares 2 times, either in `hh:mm:ss` style format, or a Date Time String, eg `YYYY-MM-DDTHH:mm:ss.sssZ`, with the return value in seconds. This allows for easy creation of countdowns, formatting of the seconds can easily be done with the existing `msToTimestamp` Expression Function, and because the return value is a number it also makes it easy to do things like triggers when the value is less than 30 seconds, or when it becomes negative to show when over-running etc....

These sort of basic time functions should help with the growing use of Expression Funcitons and Triggers for more complex automation. I had considered an additional function that would allow a timezone instead of an offset, but with the frequent confusion around time zones due to DST (eg, people referring to EST instead of EDT, or GMT instead of BST) using a numerical or timestamp offset in `timeOffset` should be a more reliable option.

### DIFF
--- a/docs/4_secondary_admin_controls/expressions/functions.md
+++ b/docs/4_secondary_admin_controls/expressions/functions.md
@@ -51,18 +51,6 @@ Finds the largest of the provided values.
 
 Finds the smallest of the provided values.
 
-**unixNow()**
-
-Get the current unix time in milliseconds.
-
-**timestampToSeconds(timestamp)**
-
-Convert a timestamp of format 'HH:MM:SS' into the number of seconds it represents.
-
-eg `00:10:15` gives 615
-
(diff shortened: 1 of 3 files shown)


Julusian (approved):
I haven't read t

PR #2889, opened 2024-06-01
Some updates to satellite API

dnmeid:
This PR:
- allows use of more than 32 buttons
- adds the text color if client requests colors (unfortunately the original API uses `color` for background color instead of `bgcolor`, so I use `textcolor` for text color instead of `color`)
- adds the option to read back colors in css rgb notation
- allows optionally to use button numbers in row/column format. The surface buttons start at 0/0 in that aspect, streamed button numbers are unchanged

API version is increased to 1.6.0  
as clients are required to actually parse parameter names and not only rely on positions, I consider the addition of textcolor parameter not a breaking change

This PR also has a bunch of completely unrelated formatting

### DIFF
--- a/companion/lib/Service/Satellite.js
+++ b/companion/lib/Service/Satellite.js
@@ -13,8 +13,13 @@ import LogController from '../Log/Controller.js'
  * 1.4.0 - Add TEXT_STYLE as ADD-DEVICE flags with FONT_SIZE as KEY-STATE property
  * 1.5.0 - Specify desired bitmap size with BITMAPS parameter when adding device
  * 1.5.1 - Remove surface size limit
+ * 1.6.0 - Allow for row/column notation,
+ * 		 - add streaming of text color when color is requested,
+ * 		 - allow choice of returned color format
+ * 		 - compatibility with internal CSS colors,
+ * 		 - allow buttons > 32
  */
-const API_VERSION = '1.5.1'
+const API_VERSION = '1.6.0'
 
 /**
  * Class providing the Satellite/Remote Surface api.
@@ -125,7 +130,10 @@ class ServiceSatellite extends ServiceBase {
 			}
 		}
 
-		const streamColors = params.COLORS !== undefined && !isFalsey(params.COLORS)
+		const streamColors =
+			params.COLORS && ['01', '1', 'true', 'hex', 'rgb'].includes(params.COLORS.toString().toLowerCase())
+				? params.COLORS.toString().toLowerCase()
+				: false
 		const streamText = params.TEXT !== undefined && !isFalsey(params.TEXT)
 		const streamTextStyle = params.TEXT_STYLE !== undefined && !isFalsey(params.TEXT_STYLE)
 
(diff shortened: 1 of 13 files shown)


companion/lib/Service/Satellite.js
```diff
@@ -125,7 +130,10 @@ class ServiceSatellite extends ServiceBase {
 			}
 		}
 
-		const streamColors = params.COLORS !== undefined && !isFalsey(params.COLORS)
+		const streamColors =
+			params.COLORS && ['01', '1', 'true', 'hex', 'rgb'].includes(params.COLORS.toString().toLowerCase())
```
Julusian: I think it would be better to change this so that it either produces one of `'hex' | 'rgb' | null`. that will make for an unambiguous interpretation of what this value means.

It might be worth creating a help function for `parseStringParamWithBooleanFallback(['hex', 'rgb'], 'hex', params.COLOR)` which will try to match the parameter against the list, then fallback to the `isFalsey` check, but always producing a string or null.

companion/lib/Service/Satellite.js
```diff
@@ -292,31 +300,41 @@ class ServiceSatellite extends ServiceBase {
...
+			Array.isArray(keyParse) &&
+			Number(keyParse[1]) < device.device.gridSize.rows &&
+			Number(keyParse[2]) < device.device.gridSize.columns
+		) {
+			this.emit('click', Number(keyParse[2]), Number(keyParse[1]), pressed)
+			socket.write(`KEY-PRESS OK\n`)
+		} else if (
+			!isNaN(keynum) &&
+			keynum <= device.device.gridSize.columns * device.device.gridSize.rows &&
+			keynum >= 0
```
Julusian: one of these <= or >= looks wrong, this would accept 0-32 which is 33 buttons

companion/lib/Service/Satellite.js
```diff
@@ -292,31 +300,41 @@ class ServiceSatellite extends ServiceBase {
...
+			socket.write(`KEY-PRESS ERROR DEVICEID="${id}" MESSAGE="Missing PRESSED"\n`)
 			return
 		}
-
 		const pressed = !isFalsey(params.PRESSED)
 
-		const id = `${params.DEVICEID}`
-		const device = this.#devices.get(id)
+		const keynum = Number(params.KEY)
+		const keyParse = params.KEY.toString().match(/^\+?(\d+)\/\+?(\d+)$/)
```
Julusian: It would be nice to make a helper method which does this parsing of KEY. Perhaps it should always return `[x: number, y: number] | null` as well as doing the validation.
Then `doRotate` and `doButton` can be changed to accept that xy value instead of parsing it themselves.

companion/lib/Service/Satellite.js
```diff
@@ -292,31 +300,41 @@ class ServiceSatellite extends ServiceBase {
...
 
-		if (device && device.socket === socket) {
-			device.device.doButton(key, pressed)
+		if (
+			Array.isArray(keyParse) &&
+			Number(keyParse[1]) < device.device.gridSize.rows &&
+			Number(keyParse[2]) < device.device.gridSize.columns
+		) {
+			this.emit('click', Number(keyParse[2]), Number(keyParse[1]), pressed)
```
Julusian: I think these emits won't work, there isn't anything listening to them on this class. 
It would work if they were made `device.device.emit(....`, but another comment should negate the need to do this

companion/lib/Service/Satellite.js
```diff
@@ -13,8 +13,13 @@ import LogController from '../Log/Controller.js'
  * 1.4.0 - Add TEXT_STYLE as ADD-DEVICE flags with FONT_SIZE as KEY-STATE property
  * 1.5.0 - Specify desired bitmap size with BITMAPS parameter when adding device
  * 1.5.1 - Remove surface size limit
+ * 1.6.0 - Allow for row/column notation,
+ * 		 - add streaming of text color when color is requested,
+ * 		 - allow choice of returned color format
+ * 		 - compatibility with internal CSS colors,
+ * 		 - allow buttons > 32
```
Julusian, line 20: does this actually change this?
I'm pretty sure that 1.5.1 already does this, but does keep using the same index system. But the indexes used in this api are always relative to the size of the grid it reported when adding the device, not the usual 1-32 that companion used to use internally

dnmeid: There was still a check to abort if key number > LEGACY_MAX_BUTTONS
That's why I picked this up, I'm using the satellite API in a current project.

Julusian: Oh right yeah I can see that now.
I guess I didn't test that change very well..

dnmeid:
all done and tested

companion/lib/Surface/IP/Satellite.js
```diff
@@ -57,8 +57,9 @@ class SurfaceIPSatellite extends EventEmitter {
 	 */
 	#streamBitmapSize
 	/**
-	 * Whether to stream button colors to the satellite device
-	 * @type {boolean}
+	 * Whether to stream button colors to the satellite device and which format
+	 * can be false, true or 'hex' for hex format, 'rgb' for css rgb format.
+	 * @type {string | boolean}
```
Julusian, line 62:
```suggestion
	 * Whether to stream button colors to the satellite device and which format
	 * can be null, 'hex' for hex format, 'rgb' for css rgb format.
	 * @type {string | boolean}
```

companion/lib/Service/Satellite.js
Julusian, line 1: I don't have a strong argument for the order of these checks in this, but I think my reasoning for the existing order was first validating that a message was correctly formed (had all the parameters), then validating the content of the message.

Julusian (approved):

PR #2893, opened 2024-06-03
fix: Swap icons for global and local variables

bfish510:


Julusian:
I'm not sure about this.
It was done this way because supporting global variables is common, and we use the `$` elsewhere to represent these.
In some locations where global variables can be used, local variables can also be used.

Perhaps just the local varaibles icon can be changed to something else?

bfish510:
Having the context in mind that `$` was used for globals prior, I agree it makes sense to keep that as such.

I'll take a look through font-awesome when I get a chance and see if there's a local variable icon that would work.

Julusian:
I'm going to close this for now, but feel free to open a new pr when you get a chance to figure out some better icons

PR #2943, opened 2024-07-08
fix: various UI tweaks and cleanup

bryce-seifert:
Some fixes after coreui update, and some other general ui tweaks

### DIFF
--- a/webui/src/Triggers/index.tsx
+++ b/webui/src/Triggers/index.tsx
@@ -329,10 +329,10 @@ function TriggersTableRow({ controlId, item, editItem, moveTrigger, isSelected }
 			</td>
 			<td className="action-buttons">
 				<CButtonGroup>
-					<CButton color="white" onClick={doClone} title="Clone" style={{ padding: 4 }}>
+					<CButton color="white" onClick={doClone} title="Clone">
 						<FontAwesomeIcon icon={faClone} />
 					</CButton>
-					<CButton color="gray" onClick={doDelete} title="Delete" style={{ padding: 4 }}>
+					<CButton color="gray" onClick={doDelete} title="Delete">
 						<FontAwesomeIcon icon={faTrash} />
 					</CButton>
 					<CButton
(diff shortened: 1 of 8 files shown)


webui/src/Triggers/index.tsx
```diff
@@ -329,10 +329,10 @@ function TriggersTableRow({ controlId, item, editItem, moveTrigger, isSelected }
 			</td>
 			<td className="action-buttons">
 				<CButtonGroup>
-					<CButton color="white" onClick={doClone} title="Clone">
+					<CButton color="white" onClick={doClone} title="Clone" style={{ padding: 4 }}>
```
Julusian: I've pushed a change that should negate the need for these https://github.com/bitfocus/companion/commit/8fddb40ba120c1f228ec6fa58ca2e5f253c5f4e0

Julusian:
personally, this area feels a bit bare now (for a stable release). maybe doesn't need to be changed as part of this though
image

webui/src/Triggers/index.tsx
bryce-seifert: Got it, I just reverted that.

bryce-seifert:
Good point, I agree. I mostly tested it with in tandem with an Installation Name, so it looks more full there. But I bet currently the amount of people using a custom name is low.

Julusian:
I think having the build visible is really useful for beta builds, just saying '3.4.0' will make it hard to tell what is a stable release and what is a beta. So I've put that back for any non-stable build

Julusian:
> Good point, I agree. I mostly tested it with in tandem with an Installation Name, so it looks more full there. But I bet currently the amount of people using a custom name is low.

I've been wondering for a while about simply pushing it to the right corner instead. But not enough to try it

Julusian:
github is complaining about a merge conflict, other than that I am happy to merge this

bryce-seifert:
Should be good now, looks like we both had the same idea about the zoom panel closing

PR #2944, opened 2024-07-08
feat: initial VEC Footpedal support

### LINKED ISSUE #2919: Add support for VEC Footpedal

### Describe the feature

Add support for VEC footpedal https://veccorp.com/foot-controls.html

I did a PR to add it to the xkeys module Companion is already using (so no code changes required to Companion), but unfortunately that's been rejected:
https://github.com/SuperFlyTV/xkeys/pull/89

So I can either do as they suggest and make a library based on xkeys (not-xkeys?) by stripping out most stuff apart from my PR from a fork of their repo:
https://github.com/SuperFlyTV/xkeys/pull/89#issuecomment-1790274754

Or something a lot simpler/lightweight like https://www.npmjs.com/package/shuttle-control-usb

Any particular preference there?

### Usecases

Let people with VEC Footpedals use Companion in the same way as xkey footpedal owners can already.

peternewman:
Closes #2919

### DIFF
--- a/assets/linux/50-companion-desktop.rules
+++ b/assets/linux/50-companion-desktop.rules
@@ -39,3 +39,7 @@ SUBSYSTEM=="usb", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0030", MODE:="666"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0010", MODE:="666", TAG+="uaccess"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0020", MODE:="666", TAG+="uaccess"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0030", MODE:="666", TAG+="uaccess"
+
+# vec footpedal
+SUBSYSTEM=="usb", ATTRS{idVendor}=="05f3", ATTRS{idProduct}=="00ff", MODE:="666", TAG+="uaccess"
+KERNEL=="hidraw*", ATTRS{idVendor}=="05f3", ATTRS{idProduct}=="00ff", MODE:="666", TAG+="uaccess"
--- a/assets/linux/50-companion-headless.rules
+++ b/assets/linux/50-companion-headless.rules
@@ -39,3 +39,7 @@ SUBSYSTEM=="usb", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0030", MODE:="666"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0010", MODE:="666", GROUP="companion"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0020", MODE:="666", GROUP="companion"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0030", MODE:="666", GROUP="companion"
+
+# vec footpedal
+SUBSYSTEM=="usb", ATTRS{idVendor}=="05f3", ATTRS{idProduct}=="00ff", MODE:="666", GROUP="companion"
+KERNEL=="hidraw*", ATTRS{idVendor}=="05f3", ATTRS{idProduct}=="00ff", MODE:="666", GROUP="companion"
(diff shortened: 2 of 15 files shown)


assets/linux/50-companion-desktop.rules
```diff
@@ -39,3 +39,7 @@ SUBSYSTEM=="usb", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0030", MODE:="666"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0010", MODE:="666", TAG+="uaccess"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0020", MODE:="666", TAG+="uaccess"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0030", MODE:="666", TAG+="uaccess"
+
+# vec footpedal
+SUBSYSTEM=="usb", ATTRS{idVendor}=="05f3", ATTRS{idProduct}=="00ff", MODE:="666", TAG+="uaccess"
+KERNEL=="hidraw*", ATTRS{idVendor}=="05f3", ATTRS{idProduct}=="00ff", MODE:="666", TAG+="uaccess"
```
peternewman, line 45: These technically overlap with the xkeys lines above, but are more specific...

assets/linux/50-companion-headless.rules
```diff
@@ -39,3 +39,7 @@ SUBSYSTEM=="usb", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0030", MODE:="666"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0010", MODE:="666", GROUP="companion"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0020", MODE:="666", GROUP="companion"
 KERNEL=="hidraw*", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0030", MODE:="666", GROUP="companion"
+
+# vec footpedal
+SUBSYSTEM=="usb", ATTRS{idVendor}=="05f3", ATTRS{idProduct}=="00ff", MODE:="666", GROUP="companion"
+KERNEL=="hidraw*", ATTRS{idVendor}=="05f3", ATTRS{idProduct}=="00ff", MODE:="666", GROUP="companion"
```
peternewman, line 45: Again these technically overlap with the xkeys lines above, but are more specific...

companion/lib/Surface/Controller.js
```diff
@@ -26,6 +26,7 @@ import { getStreamDeckDeviceInfo } from '@elgato-stream-deck/node'
 import { usb } from 'usb'
 // @ts-ignore
 import shuttleControlUSB from 'shuttle-control-usb'
+import vecFootpedal from 'vec-footpedal'
```
peternewman, line 30: This causes:

> lib/Surface/Controller.js:29:26 - error TS7016: Could not find a declaration file for module 'vec-footpedal'. '/home/runner/work/companion/companion/node_modules/vec-footpedal/index.js' implicitly has an 'any' type.
>   Try `npm i --save-dev @types/vec-footpedal` if it exists or add a new declaration (.d.ts) file containing `declare module 'vec-footpedal';`
> 
> 29 import vecFootpedal from 'vec-footpedal'
>                             ~~~~~~~~~~~~~~~
> 
> 
> Found 1 error in lib/Surface/Controller.js:29

But I can't work out how e.g. Contour Shuttle doesn't cause the same thing...

Julusian: The `// @ts-ignore` on line 27 (the line before the import) is what allows it to work.
It would be really good to have typings enabled, but I guess that isn't required for this PR seeing as that one doesn't have it either.
```suggestion
// @ts-ignore
import vecFootpedal from 'vec-footpedal'
```

peternewman: Great thanks, couldn't see the wood for the trees!

If there's an example typing (for a non TS library) I can steal I can add one for one or indeed both in due course (but yes ideally after this PR is in). Presumably ideally the typing gets added to the source package so everyone benefits?

Julusian: Maybe https://github.com/bitfocus/node-infinitton-idisplay/blob/infinitton-idisplay/index.d.ts or https://github.com/node-hid/node-hid/blob/master/nodehid.d.ts are useful examples?
Ideally the `EventEmitter` would be typed too, I think that can be done without a library now, but I haven't tried it

Issue #2189, opened 2022-11-20
Button view for Stream Deck +


sstaub:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

The buttons layout view in the "Buttons" tab should reflect the new Stream Deck + console if attached.
Rotary encoder options should also enabled by default for Stream Deck +.
If there is no Steam Deck attached it should possible to change the layout view manually.

Issue #1352, opened 2020-12-31
Accessibility: Display correct Stream Deck model button layout


ghost:
**Describe the feature**
Display the correct layout with the appropriate number of buttons (6, 15, or 32) under **Buttons** based on the connected Stream Deck model.

Don't know if this is a bug or the way I have it setup...
But Companion **always** displays 32 buttons! Why? I am visually impaired and would prefer to deal with only 15 buttons at a time, because that's the Stream Deck model I'm using. The extra buttons are an accessibility issue for me. I use Stylus to add the following CSS to remove the buttons (do I *really* need to do this, or is there some other reason that "emulator" is the first in my list of Surfaces, followed by my actual Stream Deck, and there are no settings to customize the emulator layout that I can tell?):
```css
#pagebank .bank {
    width: 20% !important;
}
#pagebank .bank:nth-child(8n) {
    display: none !important;
}
#pagebank .bank:nth-child(8n-1) {
    display: none !important;
}
#pagebank .bank:nth-child(8n-2) {
    display: none !important;
}
#pagebank .bank:nth-child(n+24) {
    display: none !important;
}
```

**I tried stopping/restarting the server, deleting the database files, etc. to no avail.**

**Is this platform dependent (windows, mac, ..)?**
Mac OS Catalina. I don't know about others.

**If documentation is required to implement, do you know where to find it?**
See CSS above. I am not familiar with the code/framework of this app.

**Usecases**
Well, there are currently (as of this writing) 3 models of Stream Deck:
- 6-button
- 15-button
- 32-button

Would be better UX for the UI to display the correct number of buttons.

josephdadams:
This is a very “one stream deck” mindset. What about those of us with multiple decks connected to Companion?

tgrushka:
@josephdadams, what does my request have to do with the number of decks connected? I don't see the "connection" between displaying the correct number of buttons and the number of decks. I suppose there should be a way to select the "active" deck you are working on? The Buttons screen should then update based on the "active" deck you are editing ... but that would seem to be a different feature/issue?

JeffreyDavidsz:
The buttons you are programming are used on all the models. It is independent of which model you use. This is why you can connect different models at the same time. For the 15 deck there is an outline on the programing page, but it might be hard to see. It would be possible to create 3 emulators, that's no problem. Only someone needs to find time to do it.

istnv:
I have all 3 sizes of streamdeck connected at the same time. They are on different pages for different functions. 
[...]

PR #2945, opened 2024-07-09
feat: view as surface

Julusian:
Related: #1352 #2189

This adds the ability to view the button grid as a particular surface.
For now this is limited to reducing the buttons shown to the ones displayed by a surface, it will later be extended to support showing 'custom' layouts and graphics for surface types that define them.

The selection is remembered across page reloads/refreshes.

There is a toggle button added above the grid to quickly enable/disable this for the current surface, the main configuration for this tool is a panel on the right.

The 'Surface' dropdown offers every surface as shown in the surfaces tab of companion. We already cache the dimensions of these, so know what size grid to show, even when they are not connected

image

In some cases, we don't have dimensions (surface hasn't been connected since we started storing that), then a warning is displayed and no size change takes effect
image

You can also choose 'Custom' in the 'Surface' dropdown, which enabled the 'Surface Type' dropdown which is configured with every surface type that companion natively supports.
You can also manually adjust the x&y offsets to position as desired.

image

### DIFF
--- a/companion/lib/Surface/Controller.js
+++ b/companion/lib/Surface/Controller.js
@@ -1,4 +1,3 @@
-// @ts-check
 /*
  * This file is part of the Companion project
  * Copyright (c) 2018 Bitfocus AS
@@ -41,6 +40,7 @@ import SurfaceIPVideohubPanel from './IP/VideohubPanel.js'
 import FrameworkMacropadDriver from './USB/FrameworkMacropad.js'
 import CoreBase from '../Core/Base.js'
 import { SurfaceGroup } from './Group.js'
+import { SurfaceLayoutRegistry } from './LayoutRegistry.js'
 
 // Force it to load the hidraw driver just in case
 HID.setDriverType('hidraw')
@@ -49,6 +49,13 @@ HID.devices()
 const SurfacesRoom = 'surfaces'
 
 class SurfaceController extends CoreBase {
+	/**
+	 * @type {SurfaceLayoutRegistry}
+	 * @access private
+	 * @readonly
+	 */
+	#surfaceLayouts
+
 	/**
 	 * The last sent json object
 	 * @type {Record<string, ClientDevicesListItem> }
@@ -111,6 +118,8 @@ class SurfaceController extends CoreBase {
 	 */
 	constructor(registry) {
 		super(registry, 'Surface/Controller')
+
+		this.#surfaceLayouts = new SurfaceLayoutRegistry()
 
 		this.#surfacesAllLocked = !!this.userconfig.getKey('link_lockouts')
 
@@ -536,6 +545,10 @@ class SurfaceController extends CoreBase {
 
 			return group.groupConfig
 		})
+
+		client.onPromise('surfaces:get-layouts', () => {
+			return this.#surfaceLayouts.getLayouts()
+		})
 	}
 
 	/**
@@ -658,10 +671,7 @@ class SurfaceController extends CoreBase {
 				location: null,
 				xOffset: config.config?.xOffset ?? 0,
 				yOffset: config.config?.yOffset ?? 0,
-				layout: {
-					rows: 0,
-					columns: 0,
-				},
+				layout: config.layout ?? null,
 			}
 
 			if (surfaceHandler) {
@@ -672,6 +682,9 @@ class SurfaceController extends CoreBase {
 				surfaceInfo.configFields = surfaceHandler.panel.info.configFields || []
 
 				surfaceInfo.layout = {
+					id: surfaceInfo.id,
+					name: surfaceInfo.displayName,
+
 					rows: surfaceHandler.panelGridSize.rows,
 					columns: surfaceHandler.panelGridSize.columns,
 				}
--- a/companion/lib/Surface/Controller.js
+++ b/companion/lib/Surface/Controller.js
@@ -658,6 +658,10 @@ class SurfaceController extends CoreBase {
 				location: null,
 				xOffset: config.config?.xOffset ?? 0,
 				yOffset: config.config?.yOffset ?? 0,
+				layout: {
+					rows: 0,
+					columns: 0,
+				},
 			}
 
 			if (surfaceHandler) {
@@ -666,6 +670,11 @@ class SurfaceController extends CoreBase {
 
 				surfaceInfo.location = location || null
 				surfaceInfo.configFields = surfaceHandler.panel.info.configFields || []
+
+				surfaceInfo.layout = {
+					rows: surfaceHandler.panelGridSize.rows,
+					columns: surfaceHandler.panelGridSize.columns,
+				}
 			}
 
 			return surfaceInfo
--- a/companion/lib/Surface/Controller.js
+++ b/companion/lib/Surface/Controller.js
@@ -674,6 +674,18 @@ class SurfaceController extends CoreBase {
 				layout: config.layout ?? null,
 			}
 
+			// If the surface has a cached grid size, a crude layout can be generated
+			if (config.gridSize && !surfaceInfo.layout) {
+				surfaceInfo.layout = {
+					id: '__auto__',
+					name: surfaceInfo.displayName,
+
+					type: 'grid',
+					rows: config.gridSize.rows,
+					columns: config.gridSize.columns,
+				}
+			}
+
 			if (surfaceHandler) {
 				let location = surfaceHandler.panel.info.location
 				if (location && location.startsWith('::ffff:')) location = location.substring(7)
@@ -682,7 +694,7 @@ class SurfaceController extends CoreBase {
 				surfaceInfo.configFields = surfaceHandler.panel.info.configFields || []
 
 				surfaceInfo.layout = {
-					id: surfaceInfo.id,
+					id: '__auto__',
 					name: surfaceInfo.displayName,
 
 					type: 'grid',
--- a/companion/lib/Surface/Controller.js
+++ b/companion/lib/Surface/Controller.js
@@ -685,6 +685,7 @@ class SurfaceController extends CoreBase {
 					id: surfaceInfo.id,
 					name: surfaceInfo.displayName,
 
+					type: 'grid',
 					rows: surfaceHandler.panelGridSize.rows,
 					columns: surfaceHandler.panelGridSize.columns,
 				}
--- a/companion/lib/Surface/Controller.js
+++ b/companion/lib/Surface/Controller.js
@@ -422,6 +422,11 @@ class SurfaceController extends CoreBase {
 			for (let surface of this.#surfaceHandlers.values()) {
 				if (surface && surface.surfaceId == id) {
 					surface.setPanelConfig(config)
+
+					setImmediate(() => {
+						this.updateDevicesList()
+					})
+
 					return surface.getPanelConfig()
 				}
 			}
@@ -651,6 +656,8 @@ class SurfaceController extends CoreBase {
 				isConnected: !!surfaceHandler,
 				displayName: getSurfaceName(config, id),
 				location: null,
+				xOffset: config.config?.xOffset ?? 0,
+				yOffset: config.config?.yOffset ?? 0,
 			}
 
 			if (surfaceHandler) {
--- a/companion/lib/Surface/LayoutRegistry.js
+++ b/companion/lib/Surface/LayoutRegistry.js
@@ -41,6 +41,12 @@ export class SurfaceLayoutRegistry {
 				name: 'Stream Deck: XL',
 				rows: 4,
 				columns: 8,
+			},
+			{
+				id: 'streamdeck-mini',
+				name: 'Stream Deck: Mini',
+				rows: 2,
+				columns: 3,
 			}
 			// TODO - add more layouts
 		)
--- a/companion/lib/Surface/LayoutRegistry.js
+++ b/companion/lib/Surface/LayoutRegistry.js
@@ -25,31 +25,137 @@ export class SurfaceLayoutRegistry {
 	#layouts = []
 
 	constructor() {
+		this.#addCountourShuttleLayouts()
 		this.#addStreamdeckLayouts()
+		this.#addLoupedeckLayouts()
+		this.#addInfinittonLayouts()
+		this.#addVideohubLayouts()
+		this.#addXKeysLayouts()
+
+		// Sort by name
+		this.#layouts.sort((a, b) => a.name.localeCompare(b.name))
+	}
+
+	#addCountourShuttleLayouts() {
+		// TODO
 	}
 
 	#addStreamdeckLayouts() {
 		this.#layouts.push(
 			{
 				id: 'streamdeck-15',
-				name: 'Stream Deck: Original',
+				name: 'Elgato Streamdeck Original',
+				type: 'grid',
 				rows: 3,
 				columns: 5,
 			},
 			{
 				id: 'streamdeck-xl',
-				name: 'Stream Deck: XL',
+				name: 'Elgato Streamdeck XL',
+				type: 'grid',
 				rows: 4,
 				columns: 8,
 			},
 			{
 				id: 'streamdeck-mini',
-				name: 'Stream Deck: Mini',
+				name: 'Elgato Streamdeck Mini',
+				type: 'grid',
 				rows: 2,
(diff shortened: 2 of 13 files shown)


bryce-seifert:
I like this idea personally, I think it will be helpful as I think button position is often something users can struggle with trying position surfaces on an infinite grid. I like the sanity check of "let me see exactly what's on this surface as the user sees it" while programming buttons.
Did some testing, and everything seems to be working as intended.

One possible suggestion, and there may be a more elegant way than mocked up below, but potentially some more visible flag that says "you're viewing as this surface" that way people don't confuse the view for missing buttons.
Screenshot 2024-07-09 at 10 01 19 AM

dnmeid:
> some more visible flag

From my side, even more NOs. Sorry to be the bad guy here, but that is a UX nightmare. Modes are bad UX.

What about this mockup:


Would this be easy enough to know where to put your button?

Julusian:
> I like this idea personally, I think it will be helpful as I think button position is often something users can struggle with trying position surfaces on an infinite grid. I like the sanity check of "let me see exactly what's on this surface as the user sees it" while programming buttons.

This is exactly the use case I am aiming for here.  
Particularly the next portion of 'complex' layouts will target the case of "I am using surface X which doesn't conform to a grid pattern, and don't know/remember where each button falls in the grid"


> I'm not sure how much I like this. I always try to inline developments with https://github.com/bitfocus/companion/discussions/2387 and the UX there does not relate to single surfaces. 

I don't think this is in contrast with some of the ideas proposed in #2387. It is aiming to solve a similar scenario as that screenshot of yours.
but instead of forcing everything into a grid shape, it embraces that surfaces aren't perfect grids.  
It isn't intended to be the only way you view the grid, but a way of checking alignment/positioning and making tweaks for a specific surface. 

Im curious, how would you display this in that grid? It doesn't quite align to a grid nicely, so I think will be confusing to display like that
image
This is actually already supported in satellite

dnmeid:
1. I can't see how the limitation to a part of the grid would embrace non-grid surfaces
2. I thought the grid vs. non-grid discussion is done. We talked a lot about how non-grid surfaces can be transfered to Companion's grid and how this could be made visible to the user. see #2387
3. Nobody seems to disagree that the setup tab is too complex for the problem it should solve and generally: please not another tab somewhere just because it has to go somewhere.

Don't get me wrong, I'm a big fan of every functionality, but I think this it the wrong UI approach to that problem.

Julusian:
> I can't see how the limitation to a part of the grid would embrace non-grid surfaces

For now it is showing just the portion of the grid, and it may remain doing so for some surface types. But stage 2 of this change is to add 'full' layout definitions so that when viewing as a surface you get a view like this:
image
This view will be able to position the buttons however it wants to, so when showing a non-grid surface the buttons can be aligned like the physical device.

I would argue that this simple mode still has value, both for consistency so that this works for every surface, even if we don't have a full drawing definition. And as a way to easily see what portion is used by a surface.
For that latter part, the other approach would be to as you have drawn, and a way to 'jump to surface X'. But when you have a bunch of surfaces potentially in overlapping, I do wonder how hard it will be to interpret.

> I thought the grid vs. non-grid discussion is done. 

Yes, we have agreed that surfaces will choose something for how they map to the grid, but I don't see a conclusion on how the user should know what that is other than through trial and error or interpreting some semi to scale, semi stretched representation of the surfaces.
This is an alternate approach to present that grid in a friendlier and easier to interpret way only when the user wants to see it like that.

> Nobody seems to disagree that the setup tab is too complex for the problem it should solve and generally: please not another tab somewhere just because it has to go somewhere.

I almost did this as a popover thing like the zoom control, but thought it would be too long/complex to be reasonable to put there. I could try doing that if you would prefer?

dnmeid:
> Yes, we have agreed that surfaces will choose something for how they map to the grid, but I don't see a conclusion on how the user should know what that is other than through trial and error or interpreting some semi to scale, semi stretched representation of the surfaces.

One of the points of my proposal is, that controls don't need to be 1x1. Let's map your Xencelabs. I think it fits 2x7 quite good.
image
The top button/dial is covered by four cells, the bottom button is covered by two cells. When you place a control on one of those cells it will interact with the overlapping surface hardware element. The user can either place his Companion control on any of the touching cells without guesswork or enlarge it to 2x2 for the top dial or 2x1 for the bottom button. I think this is simple, straight forward, and quite reusable on different surfaces. 

> I almost did this as a popover thing like the zoom control, but thought it would be too long/complex to be reasonable to put there. I could try doing that if you would prefer?

If the problem that is to be solved is: "I want to see my surface and how it will look" or "I only have one streamdeck and I get lost if I see more cells than there are buttons, I think your stage 2 can be a solution. But I would take a different aproach. I see this as an surface emulator, something which renders like the real device and behaves like the real device. I can imagine to have surface emulators also inline in the GUI but not replacing the button grid but being an equivalent tab. The button grid by default selects buttons for editing and has a hot-press option. The emulators default to press and can have an select option. Maybe what is standard and what is option can also be a user preference for the emulator. Then this would be complete inline with the existing emulators and no need of additional configuration or additional modes.

dnmeid:
And by the way I think placing controls is difficult if you have a few different surfaces with different offsets and that is the point where you need to see all of them and not one of them.

Julusian:
> I think this is simple, straight forward, and quite reusable on different surfaces.

I kind of agree but not fully.
That is the case (to users) until multiple of those buttons it overlaps has contents. And once that happens, it will be some guesswork as to which one will be used. (Maybe we execute the actions for all of them, but only one will win on feedback)

And for the author of the surface integration, they will need to define those overlaps. 
Either each integration will need to handle the merging/selection for presses & drawing or companion itself needs to become aware of this to handle it for them.

>  I see this as an surface emulator, something which renders like the real device and behaves like the real device.

Unless rotaries and things work in that, it can't really be called an emulator. A viewer maybe, but for those with rotaries it would only be useful as a programming tool not for operation. (Unless someone has a plan on how rotaries should work in an emulator)
The old emulator has not looked like a streamdeck in quite a while, there were many requests to change that because of the wasted screen space, and very few if any to restore the old styling, which suggests that an operation tool like that isn't in demand. While there havent been many requests for something like this, I think this will be very useful when working with weirder surfaces in the future.

> And by the way I think placing controls is difficult if you have a few different surfaces with different offsets and that is the point where you need to see all of them and not one of them.

I had thought yesterday that it would be nice to be able to select a surface-group in this view, rather than just individual surfaces, but didn't want to tackle that yet.  
So I am definitely open to this being continued to view multiple at once. Whether that is a rigidly defined thing, or becomes a more freeform 'desk simulator' I don't know.

This is reading to myself like I am fighting the concept of a grid, but I'm not intending to. And its not changing any of the technical details behind things, just an alternate way of viewing the grid.

dnmeid:
> Unless rotaries and things work in that, it can't really be called an emulator. A viewer maybe, but for those with rotaries it would only be useful as a programming tool not for operation. (Unless someone has a plan on how rotaries should work in an emulator)

That shouldn't be a real problem. Rendering like it looks on the device, use mouse or touch rotation movement to rotate it. Extra bonus: use scroll wheel to rotate. Done.

> The old emulator has not looked like a streamdeck in quite a while, there were many requests to change that because of the wasted screen space, and very few if any to restore the old styling, which suggests that an operation tool like that isn't in demand. While there havent been many requests for something like this, I think this will be very useful when working with weirder surfaces in the future.

The current emulator should not be called an emulator in that regard, I'd rather call it a webpanel.

> So I am definitely open to this being continued to view multiple at once. Whether that is a rigidly defined thing, or becomes a more freeform 'desk simulator' I don't know.

What would be the difference then from my surface view? Only that you don't see unused cells or anything else? In my proposal the surface view actually has the main purpose of positioning the surfaces, but I'm open for a simulator view. Overlapping surfaces is a challenge for that, maybe one could select / deselect the surfaces or groups to show on simulator. I don't think we should use too much effort for this feature because Companion is about using equipment with real surfaces or virtual surfaces and the main GUI is for programming that stuff and not for running it. A simulator can help while programming but the main focus of the GUI should stay programming.

bryce-seifert:
Do we have any telemetry on how many people actually use Companion with multiple surfaces / overlapping surface layouts?

Maybe I’m the odd one out, but I don’t utilize surfaces in that way and would consider it more of an advanced use case. The Surface mockup with them all overlayed is confusing to me, in terms of how I think about the grid and organize my Companion systems. but maybe seeing an actual implementation would make it make more sense. 

And data could prove me dead wrong, but my suspicion is that lots of people use companion with just a single Streamdeck, or at least a single type of Streamdeck.

dnmeid:
Overlapping surfaces have been the standard ever since. When we started this workflow it was a little awkward to many people and still is today. Often beginners ask why both of their surfaces show the same buttons or how they can address the second surface individually. The programming style with a side-by side layout is only made possible recently with the larger grid size (not counting multiple streamdeck minis).
My assumption is, that a side-by-side layout will get more and mode popular with the current and coming GUI changes and at some point will be the prevalent layout style.
But overlapping surfaces will never die. I see a lot SD mini or legacy with Satellite for some limited remote access e.g. from a lectern or director. Overlap is what you want to have there so you can use same programming on different surfaces.

In my surrounding (professional events industry), I'd say 50% of the people are using 1x XL, 30% more than one XL, 20% mixed surfaces. The reason is that we like to add streamdecks to the genuine consoles. In the streamer and more budget-oriented world, the combination of XL and SD+ is very popular because it gives some rotaries and good amount of buttons. They usually replace the consoles completely with Streamdecks.

Speaking about my Mockup: What does confuse you? Keep in mind that this is only a quick Mockup, done over a year ago and does not 100% fit the context here. Have you read #2387 ?

bryce-seifert:
That's interesting anecdotal evidence of what devices / how people are using them, but I would be interested in the raw data of the userbase as a whole.

> Speaking about my Mockup: What does confuse you? Keep in mind that this is only a quick Mockup, done over a year ago and does not 100% fit the context here. Have you read [#2387](https://github.com/bitfocus/companion/discussions/2387) ?

I read thru this and it makes slightly more sense, but I still conceptually have a hard time with the windows/surfaces concept as laid out there. I think I could grow to learn it, but trying to be an advocate for the most basic of user who just needs a simple setup with one or two decides. I feel like being as simple to go from install to programed surface is still important.Maybe, as you in that discussion, "once you see this in action it will be very easy." But having not seen anything like this implemented or drafted, and as such have a hard time breaking the current layout/ pages mentality. I do think getting more voices in the discussion would also be helpful.

All that said, this PR is likely not the best place for this whole discussion, so I'll defer to that discussion thread if I have more input.

Julusian:
> Do we have any telemetry on how many people actually use Companion with multiple surfaces / overlapping surface layouts?

There is some, here is a sampling of how many unique surfaces each online user has reported across the last week This will only include users who are online, an unknown proportion are offline.

Fun fact, one user of 2.2.0 has 87 surfaces!
There is a potential for data skew if users have copied around the full config folder including the machid file, as we use that file as the uniquely id of the machine.
This data doesn't include emulators either.


-------
> Maybe I’m the odd one out, but I don’t utilize surfaces in that way and would consider it more of an advanced use case.

> Overlapping surfaces have been the standard ever since. When we started this workflow it was a little awkward to many people and still is today.

I think we are talking about two related but not quite the same thing.  
I suspect that very few people have surfaces overlapping showing the same page. But many will have them overlap but on different pages.

So with what Dorian proposes in #2387, if we show every surface on every page, then yes it will be a confusing mess of overlaps.

Even though I implemented the larger grid, I haven't reconfigured my office setup to put everything on page 1 simply because I can't be bothered to reconfigure things. But if I were to add a new surface, then I would do it the new way.

-------

But getting back to the question at hand, does this proposed view serve a good purpose or is it going down the wrong road?
Is the consensus that this should be done as a popover like the zoom, or is keeping it as this tab acceptable?		

I still think it has value as a way of viewing the portion of the the grid being configured with a specific streamdeck in mind, while #2387 proposes something that will help with placing multiple within the grid. I don't think it has to be one or the other.
And maybe later on it could be turned into a new 'emulator' as well as this view or instead of.

Based on the stats above, I think the side by side view is probably not that useful, so while a nice idea probably won't be done.

And honestly, I am not how long it will be until the grid changes from #2387 get done. I so rarely find myself doing something complex enough to benefit from windows, that the times I do I would be satisfied with a better way to link/ref a button to another. 
So I do think we should try to avoid doing things that won't work in the world proposed by #2387, but I don't think we should avoid doing things just because that proposes different/better/bigger-picture solutions.

Julusian:
The discussion on this ended before the Studio was announced. But I think the Studio brings a challenge that this would solve well.

The buttons on the studio aren't square but are widescreen. At some point, we will start drawing these in the correct aspect for the Studio, and will also need to draw square for other streamdecks.
So the challenge becomes how you do easily see in the UI how the buttons will draw in different aspect ratios?

For me to get a good overview of what I have programmed, I will need to switch the grid view to be using the correct aspect ratio for the studio, and easily switch back to check that relevant ones still look correct on an XL.

I think that could either be solved by this, or a different/simpler view mode which only changes the aspect ratio of the buttons. In the future, there likely will be more aspect ratios involved.
But what about for the strip on the streamdeck+? Once we can do different aspect ratios, we ought to use it for that too. And for the vertical strips on the loupedecks. There should be a nice way to preview those.

I could also throw in that xencelabs above; the LCD on it is 8 characters per button with no colour control. Being able to verify that my labels for those buttons are correct and will fit would be nice (but not essential, this one is simpler to interpret)

fbosman:
My 5cts (and perhaps overkill)
I think we should first detach buttons from action procedures.
When we have a layer of abstraction we can define a lot of surface layouts and attach buttons to procedures. 
This does not only apply to action, but also to feedbacks.
When this separation is not done we only add a lot of complexity to layout and actions.

Julusian:
With the graphics overhaul making it so that we can now draw non-square, I think that this has become more important. The lack of a good way to see how a SD+XL will look from within the UI is an annoyance I have been hitting.

It is no longer possible to get an easy overview of how a SD+/+XL will look because of the 2:1 aspect ratio of the lcd strip. It is possible to see by individually opening the editor of each button to see the browser preview, but it would be much better to be able to see in the grid how everything will look. (and assumes there are no browser preview rendering differences. there shouldnt be, but they do happen)

And with how the SD+XL has 6 encoders below 9 columns of buttons, where the encoders sit is non-obvious, which this would also explain.

So I might look into reviving this for 5.1/5.2.

I havent re-read the conversation here, but the only thing that I am currently not really sure on is whether this should be in the button editor or whether it should be a new emulator view.  
For me, inside the button editor is best, having it as a separate view will mean a lot of switching back and forward while refining the looks.

krocheck:
I just gave this a look and for the most part I like how the interface provides the preview option.

But, the stretch goal long-term to display a preview overlayed on the device rather than the raw grid is ideal.  As an example I have a Contour Shuttle Pro on my desk (not sure why, since I never use it) and I plugged it in.  The grid that was provided in the preview did not intuitively align to the device, plus which button maps to the jog and shuttle is unclear.  Now, could I look up the help to figure it out?  Yes and probably should.  But we may want to pull the module help through to the edit surface interface rather than having to dig into Settings/Surfaces to get that documentation, and possibly pull it through to here because being able to pop that up while looking at the surface preview would make a big difference.

I like the ability to view as "A model of surface ...", but I feel like it should populate the surface types for all modules installed.  e.g. I can't view as an Elgato anything until I plug one of that type in.  But if I'm trying to make a config in advance and don't have the particular surface I'm SOL.

I'm not a fan of the older mock-up where you see the surfaces overlayed on a master grid for this.  That's too busy for me.  I could see something like that on the Surfaces page as a replacement for manually defining the grid offsets and to help with groups instead.

Julusian:
Playing around with making the UI reflect the actual model specs, and a naive representation of each model is looking messy but sane.   
In a new branch, don't have anything worthy of a new PR yet.

The intention is to ask surface modules for more detail on how to accurately draw them, but we can't guarantee they will do that (backwards compatibility), so the UI is doing a crude draw from just the layout definition we already use.  
So the final version will look better than this in most cases, but sometimes we may need this gradual degredation


For a SD+:
image

Studio:
image

Neo:
image

Julusian:
Closing this in favor of #4469

A clean PR without stale discussion from 2 years ago will make it easier to track what is still relevant

PR #2947, opened 2024-07-11
feat:  Add Feedback Logic #372

### LINKED ISSUE #372: Add Feedback Logic
**Describe the feature**
Give the ability to have AND conditions to create additional states.  As a result, need to be able to set priority order of feedbacks.

**Is this platform dependent (windows, mac, ..)?**
No

**Usecases**
Was on a show where I was controlling a Videohub upstream of an ATEM.  I wanted to be able to have my source buttons be a lighter shade of their background when they were routed into their respective feed and display standard pgm/pvw style feedback when appropriate.  I would have to build 3 'sets' of feedback (lower the number, higher the priority):

1. The in/out route AND source PGM (red BG)
2. The in/out route AND source PVW (green BG)
3. The in/out route (light purple BG)
- No feedback (purple BG)

Of course this is a nice to have, but I ended up deciding which buttons were more important to show the route feedback vs. pgm/pvw feedback.  Worked, but not as clean as it could have been.

Thanks,
Keith

Julusian:
Closes #372

This implements some special feedbacks that can provide logic operations.
Currently this is two feedbacks a `Logic: AND` and `Logic: OR`.

The idea is simple, these feedbacks can contain other feedbacks, and will perform the specified logic operation on their values. There is no defined limit on how deep/complex these structures can go, it is possible to nest them inside each other without limit. (although the ui will probably go a bit weird after a few levels)

image

This change turned out larger than I was hoping, as it involved some heavy refactoring of `FragmentFeedbacks`. That class has been broken up a bit and instantiates a class for each feedback it contains, to make it easier to handle the recursive nature of various operations.
To keep things simple in the frontend, the socket-io api hasn't been changed much, instead we rely on 'searching' for the feedback amongst the hierarchy when making a change to one.

It is possible to drag and drop feedbacks around within the list, including in and out levels

TODO:
- [x] Add back the 'invert' checkbox (or implement as an option?)
- [x] Test import/export
- [x] Test subscribe/unsubscribe flow is executed correctly
- [ ] Unit tests?


phillipivan:
@Julusian two questions / requests regarding this:

[1] Will these feedbacks have an invert option like regular boolean feedbacks?
[2] Can you please add an `Logic: XOR` logic option?

Between the two of these we should then have a complete palette of boolean operators to work with for complicated setups.

Julusian:
1) I disabled the invert option because it messed with the layout and I wasn't sure if it was necessary, but they can be added back
2) I almost did, then wasn't sure what an XOR should do considering there can be any number of input values to combine. If you could define how it should work, then it can be easily added. The implementation of the actual feedbacks are pretty simple https://github.com/bitfocus/companion/blob/feat/feedback-logic/companion/lib/Internal/BuildingBlocks.js

phillipivan:
1. They could also be defined as `Logic: NAND` and `Logic: NOR` if that is cleaner. The logical output should be the same.

2. I've only ever worked with 2 input XORs, for which the truth table is well understood. Intuitively for x-Input XORs I would define them as true when exactly one input is true. However the quote unquote intuitive definition of an x-input XNOR (ie with the invert enabled) Id probably offer as true when all inputs are the same; which is not a logical complement of the former.

phillipivan:
I did some more reading on `n-input XORs`. There seems to be two completing definitions, the alternate, which mimics a tree of XOR gates, is true when an ODD input of inputs is true. To my way of thinking this is far more esoteric in its potential uses than a true with only 1 definition; and could always be recreated with the tree of 2-input XORs if it was required.

PR #2962, opened 2024-07-25
fix: kebab menu not opening in Safari

bryce-seifert:


willosof:
thanks. weird that it worked on my safari 💭

PR #2969, opened 2024-07-29
feat: support node22 from modules

Julusian:
This is groundwork for supporting modules to run on node22.
This should land in 3.5 beta once 3.4 is released

Before this can be merged we need:

- [ ] Update [bundled modules workflow](https://github.com/bitfocus/companion-bundled-modules/blob/31878343c8ed5d7b4b7911d800ac7c999e315321/.github/workflows/update-module.yml#L37-L41)
- [ ] Update module-base [schema](https://github.com/bitfocus/companion-module-base/blob/main/assets/manifest.schema.json#L76) [engines](https://github.com/bitfocus/companion-module-base/blob/main/package.json#L21) and maybe more
- [x] Any more hardcoded checks for node18?
- [ ] Update templates
- [ ] Check for dependencies which can be updated
- [ ] Test on macos
- [ ] Test on windows
- [x] Test docker images
- [x] Test on pi


Julusian:
There are still bits of this to do, but some portions needs a little coordination with buttons (before we can accept any node22 based modules), and further testing is easier once this is in develop

PR #3019, opened 2024-09-06
feat: collapsible variable table values

phuvf:
- Added separate copy actions for variable names and variable values
- Added `More` and `Less` actions to variable values where value length > 100 chars

Note: My first time working in React - may have done things in a particularly clunky way.

### DIFF
--- a/webui/src/Components/VariablesTable.tsx
+++ b/webui/src/Components/VariablesTable.tsx
@@ -202,8 +202,6 @@ const VariablesTableRow = observer(function VariablesTableRow({
 						'(empty)'
 					) : (
 						<code
-							className="compactValue"
-							data-name={`${label}:${variable.name}`}
 							style={{
 								backgroundColor: 'rgba(0,0,200,0.1)',
 								color: 'rgba(0,0,200,1)',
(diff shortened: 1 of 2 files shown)


Julusian:
Thanks for giving this a try.

This direct dom manipulation should be avoided in react, these kinds of state changes should be done with react state instead. I would suggest looking at https://github.com/bitfocus/companion/blob/main/webui/src/Controls/ActionSetEditor.tsx#L61, and how we use the `panelCollapseHelper` to store and update the state of what should be collapsed or not. 
The same helper type could probably be used here too, although it might need a small change to be able to tell it to default to collapsed, I think it defaults to expanded currently.

Because we are so close to the release of 3.4, and I think this would be good to include there, I hope you don't mind if I make these changes for you tomorrow evening. Let me know before then if you would like to make these changes yourself and this can be squeezed into 3.4.1 in a few week instead

phuvf:
Julian,

Thanks for the notes. Very happy for you to take it from here - I'll
compare what I wrote with what we end up with as a lesson on how to do this
correctly.

Nick

PR #3028, opened 2024-09-11
feat: move legacy modules into bundled-modules repository

Julusian:
This makes development/building a bit smoother, as these modules are now pulled in precompiled


jwetzell:
Is there any concern with the fact that Buttons is pointing at the bundled-modules repo when it downloads modules?

Julusian:
No, I think that for now it will have no impact, but does mean that buttons can start loading those in too, once they have a chance to implement that

jwetzell:
That makes I was going through the list trying to compare and noticed some modules that exist both in `module-legacy` package.json and the `companion-bundled-modules`. Is this intentional or could the `module-legacy` package.json be further cleaned up?

Julusian:
@jwetzell There shouldn't be any duplicates, but there could be. There is a script which checks for duplicates https://github.com/bitfocus/companion/blob/main/tools/check_for_unused_legacy_modules.mjs, that I run every now and then.
The one exception is `bmd-videohub`, which has to remain, as a couple of other modules borrow code from it.
But it should now be removed from the new compiled directory of legacy modules, it doesn't need to be in there.

jwetzell:
Ha, That's the one I noticed didn't look into further, figured there might be a reason.

Julusian:
also going forward, it should cleanup by itself https://github.com/bitfocus/companion-bundled-modules/commit/a020fa6234a2476dc69353b60f71935896b3a240
instead of the manual process today

Julusian:
This needs a change to the website before it can land in beta, but it can go to develop now

PR #3045, opened 2024-09-20
feat: Migrate to SQLite

krocheck:
This implements SQLite to replace the JSON file DB currently used.

Existing DB implementations were moved to `/Data/Legacy` with write functionality removed in order to facilitate migration to the new format for existing installations.  The tables used are simple `(id, value)` string rows with `id` flagged as `UNIQUE`.  

This does include a DB version increment from v4 to v5, and this was done to simplify conditions during the migration and ensure import functionality remained untouched.  Part of the `v4tov5` upgrade is to move `controls` from the `main` table into a dedicated `controls` table.  Also, it migrates the separate, legacy `clouddb` into a `cloud` table.

Upon startup, the driver ...
1. Attempts to start without create ability.  If that fails ... 
2. It checks the filesystem for a legacy DB file
- If found, the Legacy driver is loaded, a DB is created, and all data is "as-is" moved to the `main` SQLite table 
- If not found ...
3. It creates a new DB and loads the defaults

TODOs
- [x] Bolster error handling, including startup (attempt to recover backup)
- [x] Deal with types better
- [x] Deal with types even better
- [x] Fix tests
- [x] Test, test, test

Initial testing showed successful OOB startup and migration from a v4 config of relatively small size/scope.


companion/lib/Cloud/Controller.ts
```diff
@@ -80,31 +90,31 @@ export class CloudController extends CoreBase {
...
 
-	constructor(registry: Registry, clouddb: CloudDatabase, cache: DataCache) {
+	constructor(registry: Registry, cache: DataCache) {
 		super(registry, 'Cloud/Controller')
 
-		this.clouddb = clouddb
 		this.cache = cache
 
-		this.data = this.clouddb.getKey('auth', {
+		this.data = this.db.getTableKey('cloud', 'auth', {
```
Julusian: it would be nice for these `'cloud'` strings to be using a constant

companion/lib/Cloud/Controller.ts
```diff
@@ -80,31 +90,31 @@ export class CloudController extends CoreBase {
...
-		const uuid = this.clouddb.getKey('uuid', undefined)
+		const uuid = this.db.getTableKey('cloud', 'uuid', generateMachineId())
 		this.#setState({ uuid })
 
-		const regions = this.cache.getKey('cloud_servers', undefined)
+		const regions = this.cache.getKey('cloud_servers', {})
 
 		if (regions !== undefined) {
-			for (const region of regions) {
+			for (const region of Object.values(regions)) {
```
Julusian:
```suggestion
			for (const region of Object.values<any>(regions)) {
```
avoids the need for the `@ts-ignore`s below (probably)

companion/lib/Cloud/Controller.ts
Julusian, line 1: I'm still wondering about whether it is ok to be including this cloud auth stuff in the full-export or support bundles.
But if we are doing that then no other comments on this file

krocheck:
@Julusian @haakonnessjoen  I've been testing for a lot of edge cases especially comparing against the previous DB behavior.  We didn't guard against the DB not being able to written to disk and therefore the software would work and simply log an error when the writes failed because everything was in RAM.  This is different.  If for some reason the SQLite doesn't fire up it will start returning 'undefined' to get 'get' calls and the software goes sideways from there.  So we need to find a way to guard against this.  I think we can either
- Fatal error and quit
- Spin up a RAM SQLite and still attempt backups

Lemme know your thoughts.

Julusian:
>  If for some reason the SQLite doesn't fire up it will start returning 'undefined' to 'get' calls and the software goes sideways from there. So we need to find a way to guard against this.

Thats a good point that I wouldn't have considered.  
I want to say that it makes sense to refuse to run because of this, but I wonder if that will break workflows for some people? Because particularly in shared environments, being able to load config from a custom dir has been a questioned asked a bunch of times. And if I were doing that, I would make that config readonly, to ensure that one person can't cause problems for the others.

So I feel like we need to support continuing to run normally, just with no persistence to match today. Meaning your second option.

Perhaps we should consider marking the system as being in readonly mode, and propogating that into the ui to disable everything? That would be quite a large task by itself, and I'm not sure justifies the effort. Or it could be a banner at the top of the page, and let everything be edited as normal?  
Just throwing ideas out on some better UX than today, I don't mind if we don't do these.

companion/lib/main.ts
```diff
@@ -143,12 +143,20 @@ program.command('start', { isDefault: true, hidden: true }).action(() => {
 	}
 
 	// copy an older db if needed
-	if (configDir !== rootConfigDir && !fs.existsSync(path.join(configDir, 'db'))) {
+	if (
+		configDir !== rootConfigDir &&
+		!fs.existsSync(path.join(configDir, 'db')) &&
```
Julusian, line 148: does this need to check for 'db' or should it just be 'db.sqlite'?

companion/lib/Data/StoreBase.ts
```diff
@@ -39,334 +48,244 @@ export class DataStoreBase {
...
 	private readonly cfgFile: string = ''
 	/**
-	 * The full temporary file path
+	 * The full main legacy file path
 	 */
-	private readonly cfgTmpFile: string = ''
+	private readonly cfgLegacyFile: string = ''
 	/**
-	 * The stored defaults for a new store
+	 * The default table to dumb keys when one isn't specified
```
Julusian:
```suggestion
	 * The default table to dump keys when one isn't specified
```

companion/lib/Data/StoreBase.ts
```diff
@@ -39,334 +48,244 @@ export class DataStoreBase {
...
-		let out
-		this.logger.silly(`${this.name}_get(${key})`)
+	public getTable(table: string): any {
+		let out = {}
 
-		if (this.store[key] === undefined && defaultValue !== undefined) {
-			this.store[key] = defaultValue
-			this.setDirty()
-		}
+		if (table.length > 0 && this.store) {
```
Julusian: should these checks cause the methods to be no-ops, or perhaps it would make more sense for them to throw an error instead? They shouldn't ever be hit anyway

companion/lib/Data/StoreBase.ts
```diff
@@ -39,334 +48,244 @@ export class DataStoreBase {
...
-		} else if (fs.existsSync(this.cfgBakFile)) {
-			this.logger.warn(`${this.name} is missing.  Attempting to recover the configuration.`)
-			this.loadBackupSync()
-		} else {
-			this.logger.silly(this.cfgFile, `doesn't exist. loading defaults`, this.defaults)
-			this.loadDefaults()
-		}
 
-		this.#setSaveCycle()
+			this.setDirty()
```
Julusian: setdirty after a read?

companion/lib/Data/StoreBase.ts
```diff
@@ -377,67 +296,134 @@ export class DataStoreBase {
...
-						if (!dbObj[k]) dbObj[k] = {}
-						dbObj = dbObj[k]
+	 */
+	public setTableKey(table: string, key: string, value: any): void {
+		if (table.length > 0 && key.length > 0 && value) {
+			if (typeof value === 'object') {
+				value = JSON.stringify(value)
+			}
+
+			const query = this.store?.prepare(
```
Julusian: these prepared statements could be cached and reused, which might be good for performance. It also might be a bit premature to be thinking about that, so feel free to ignore

companion/lib/main.ts
krocheck: It needs to include `db` still if they leapfrog major updates.

companion/lib/Data/StoreBase.ts
krocheck: I mean, it could be poetic in the right context :)

companion/lib/Data/StoreBase.ts
krocheck: 🤦

companion/lib/Data/StoreBase.ts
krocheck: Yeah I could.  I want to try subbing the hard table with an `@table` param (but a different keyword) and if that works that makes caching these easier.

companion/lib/Data/StoreBase.ts
krocheck: So if we failover to RAM this theoretically would never happen.  But yeah, I've considered logging errors instead of throwing them because we'd have to go out into the software and try-catch our calls then.  This is why if we don't get a successful `store` object (disk or RAM) we need to just error out at that point.  I don't know how or why `new Database(':memory:')` could possibly throw an error, but the spec says it could.

companion/lib/Cloud/Controller.ts
krocheck: Moved to `CLOUD_TABLE`

companion/lib/Cloud/Controller.ts
krocheck: done

companion/lib/Data/StoreBase.ts
krocheck: done

companion/lib/Cloud/Controller.ts
krocheck: Support bundles is one thing, as we do a popup warning about that.  I checked full export though and that is not a dump of the database, instead it requests all the data from the controllers, then bundles and sends.  So at this time, `cloud` data is still not included in a full export.

haakonnessjoen: sounds great!

companion/lib/Data/StoreBase.ts
krocheck: They get cached now.  Couldn't make table a var.

companion/lib/Data/StoreBase.ts
krocheck: So there's a final failover to RAM and then a fatal error if it still can't do it.  Still testing through the edge cases and how the error-exit works.

krocheck:
@Julusian I think this is ready for another scrub

Julusian:
I'm happy once the tests are passing :)

dnmeid:
I'm not a SQL expert, please forgive me the question if it's stupid. I thought SQL has quite a strict typing of the value fields. We have in the JSON database some properties that have been repurposed over time and can possibly be of different types. E.g. the text size has been originally a number and now is a string, that can have a number or auto. Also a lot of modules have changed types of their values over time without renaming them or doing an upgrade script. How does SQL handle that loose typing?

Julusian:
@dnmeid this isn't using the sql db for any 'proper' sql queries, we are pretty much still using it as a simple key value store.
we don't need the full power of sql queries, we really just want a key value store that is more efficient than reading and writing a large blob of json to disk, and that avoids the max string length limitation of nodejs.

but what this does now allow us to do, is to start splitting busy/large areas into their own key-value stores. I think this pr does this for controls, so that whenever the json for a control changes, we can update just that one in the db. Instead of today where we are telling the db that the json blob of all the controls has changed, and where it writes the whole config file to disk again.

the tables in this database currently have two columns, id and value. the value being a stringified blob of json.

of course, we could embrace the sql more in future, but as we don't have anything that we don't want to keep loaded in memory that we want to be able to search, there isn't any value in doing so currently.

krocheck:
> I'm not a SQL expert, please forgive me the question if it's stupid. I thought SQL has quite a strict typing of the value fields. We have in the JSON database some properties that have been repurposed over time and can possibly be of different types. E.g. the text size has been originally a number and now is a string, that can have a number or auto. Also a lot of modules have changed types of their values over time without renaming them or doing an upgrade script. How does SQL handle that loose typing?

TL;DR there's no issues with typing because this implementation is very basic for starts.

In this case we're not doing a strict schema for all types with tons of tables.  While migrations like that could be done in the future, the first step here was to take the existing key/value pairs in the database and transpose them into key/value (`id`/`value`) rows in a `main` table.  The values still live as JSON and are parsed on query.  `control` is the only key/value pair that was moved to its own table of the same structure, but those `value`'s are stored as JSON as well.

dnmeid:
Thanks for clarification. So all the values are technically stored as strings and we do a JSON.stringify on write and JSON.parse on read. Got that.

krocheck:
> Thanks for clarification. So all the values are technically stored as strings and we do a JSON.stringify on write and JSON.parse on read. Got that.

Yep, here's a quick link to that block for your reference: https://github.com/bitfocus/companion/blob/c35c52efb637d104a4df6fb2a9e216bffeaf5c25/companion/lib/Data/StoreBase.ts#L248

Issue #1648, opened 2021-07-01
Remove yellow border around active button


Brennan-Dan:
**Describe the feature**
A clear and concise description of what the feature does
Remove Yellow border around active buttons, Like the ability to remove the top bar of the buttons.

**Usecases**
Add a couple of usecases for why this should be implemented and when/why to use it.

When using feedback control it isn't necessary to know if a button is active/latched when using feedback control. Would be a nice feature to be able to turn on/off as per the topbar setting.

Mozoloa:
Please god yes, how many hours have I spent trying to work around the absence of this feature, my brain is a bowl of spaghettis

Brennan-Dan:
In addition to disabling the yellow border like the top bar, the ability to turn off the play symbol that displays when a series is actions are playing out on the button. This is because the feedback state of the button is already showing me what is going on.

Schrootpoot:
Would also love to see this. My usecase the border is completely unnecessary and makes it look a lot less clean than it could be.

nqeldridge:
What consequence does this have on the button state itself?  The most annoying part of the yellow border in my experience is that when it is active, I have to press the button once to acknowledge the border and a second time to execute the actions associated with the button.  Is that functionality being resolved as part of this or does that need to be a separate feature request?

Julusian:
> The most annoying part of the yellow border in my experience is that when it is active, I have to press the button once to acknowledge the border and a second time to execute the actions associated with the button.

No, this is not to change any behaviour, just how the buttons look visually.

So for you it will likely make that more confusing as there will be no visual indication that it needs to be pressed twice.

> I should add, this specifically is the case when using button presses as an action of another button.

That sounds like you have some quirks in your setup causing these buttons to be pushed but never released.

Julusian:
This is possible in the betas now

<img width="812" height="857" alt="Image" src="https://github.com/user-attachments/assets/36b165c4-7d79-4b36-b445-1fc3fac413d4" />

Issue #617, opened 2019-04-17
Move button labels


PhillPower:
**Describe the feature**
An option to move the button labels (the black 1.1, 1.2, 1.3 etc.) to the bottom rather than the top.

**Is this platform dependent (windows, mac, ..)?**
No.

**Usecases**
When a Stream Deck is mounted at an angle, or viewed from a sharper angle than normal, to help with seeing the buttons assigned function.

stuartbaulch:
This would be great for when I have it sitting on my lighting console.

I came here to make this request so I'll boost yours instead.

Julusian:
Is this still wanted? 
It has been possible for a few releases to hide the top bar entirely, is there still value in being able to move it to the bottom?

PhillPower:
> Is this still wanted? It has been possible for a few releases to hide the top bar entirely, is there still value in being able to move it to the bottom?

I personally always hide the labels anyway these days so that's a no from me, but other people may still want this.

PR #3065, opened 2024-10-02
feat: remove yellow border

leonreucher:
Implemented https://github.com/bitfocus/companion/issues/1648
Added a new config option for disabling the yellow border around a pressed button, if the top bar is disabled.

### DIFF
--- a/companion/lib/Data/UserConfig.ts
+++ b/companion/lib/Data/UserConfig.ts
@@ -30,91 +30,91 @@ export class DataUserConfig extends CoreBase {
 	/**
 	 * The defaults for the user config fields
 	 */
-        static Defaults: UserConfigModel = {
-          setup_wizard: 0,
+    static Defaults: UserConfigModel = {
+		setup_wizard: 0,
 
-          page_direction_flipped: false,
-          page_plusminus: false,
-          remove_topbar: false,
-          show_pressed_border: true,
+		page_direction_flipped: false,
+		page_plusminus: false,
+		remove_topbar: false,
+		show_pressed_border: false,
 
-          xkeys_enable: true,
-          xkeys_legacy_layout: false,
-          elgato_plugin_enable: false,  // Also disables local streamdeck
-          usb_hotplug: true,
-          loupedeck_enable: false,
-          contour_shuttle_enable: false,
-          vec_footpedal_enable: false,
-          blackmagic_controller_enable: false,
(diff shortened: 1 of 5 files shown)


dnmeid:
I see that this is a user request and I also can imagine situations where you want to style your buttons yourself and don't want to have that border. But may I suggest a more sophisticated implementation.
I don't like to have yet another configuration checkbox, especially if then there are two checkboxes which influence the visual outcome. The "hide border" is only doing something if "show topbar" is off.
I'd prefer to have this combined in one dropdown, e.g. named `Indicators`. There could be `Topbar`, `Border` and `none` options.
The setting on the individual button can reflect this and and additionally offer `use default` like we have today.
Upgrading this in existing database from a boolean value to a dropdown is not that big problem.

Side bonus: we can offer bottom bar (#617) with same elegant ui and don't need another checkbox for that again.

Julusian:
I agree with Dorian.
I did start and not quite finish that task for supporting moving the top bar to the bottom, which might be a useful reference on what is needed to support more options in thag dropdown. I didn't finish it though, probably not doing all of the upgrade/import bits (or maybe doing it but not testing it fully)
https://github.com/bitfocus/companion/compare/main...feat/move-topbar-to-bottom

leonreucher:
Thanks for the feedback!
I adjusted the code and now you can choose between Top, Bottom, Border and None. It's working fine, I am only having trouble coding the upgrades from previous versions to the new layout.

dnmeid:
You could do an upgrade function to upgrade the database. Look in companion/lib/Data/Upgrades/ and companion/lib/Data/Upgrade.ts
Other option is a rolling upgrade, that means the logic you are using would need to be able to cope with the old boolean value and the new enum (string?) value as well. Then there is no need to rewrite existing database values, they are just interpreted correct. In a new database or upon change of the setting, the new value is set.

Drawback of the upgrade script is that the users can't go back, i.e. the database will not be compatible with older Companion versions. Drawback of the rolling upgrade is that the logic of the upgrade is offloaded to different locations, including a dropdown, which actually should not do anything besides dropdown stuff.

Julusian:
@leonreucher did you mean to push some more commits to this PR/branch?

leonreucher:
@Julusian tbh. I was still not able to get the upgrade part ready and working. I'll close this PR and open a new one when I got it to working.

Issue #3053, opened 2024-09-25
[BUG] Feedback Logic AND not correctly exported or imported 


LikeThisReview:
### Is this a bug in companion itself or a module?

- [X] I believe this to be a bug in companion

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Imported a full export which contains a button with Feedback Logic AND with condition 1 “atem: transition: selection {ME1, contains, Key1} and condition 2 “boolean expression “. After the import the Feedback the first condition shows “atem: transition” but not not {ME1, contains, Key1) and the logic AND does not return the desired result.

### Steps To Reproduce

1. create a button with Logic End; 
3. Export; 
4. Import; 
5. Check config button under Feedback

### Expected Behavior

Feedback as defined in original configuration 

### Environment (please complete the following information)

```markdown
- OS:macos
- Browser:safari
- Companion Version: 3.4.0
```

Julusian:
This is fixed in the v3.4.2 release

LikeThisReview:
Many th!

PR #3087, opened 2024-10-13
fix: import children for conditions in triggers

mrlinden:
Import children for feedback was fixed in #3053. This PR makes the corresponding changes for the condition in triggers.


Julusian:
Thanks!
I'm annoyed that I forgot to test this when doing the other fix...

PR #3089, opened 2024-10-16
feat: move custom variables to $(custom:...) with backward compatibility

krocheck:
Responds, in part to #2812 by moving custom variables from `$(internal:custom_...)` to `$(custom:...)`.  Checks are added to maintain backwards compatibility with legacy names.

My main reason for attempting this is that with the addition of `$(this:...)`, it seems logical to me to make this change.  The fact that `custom` isn't a reserved word in v2.x is concerning, but I think is quite the edge case for someone to completely rename an instance to `custom`.  But that can be investigated for upgrade mitigation ... though scrubbing the variables might become difficult.

TODO:

- [x] Validate basic functionality and backwards compatibility
- [x] Validate custom variable internal actions
- [x] Validate triggers
- [x] Update documentation
- [x] Update any labels, like indicated in #2812
- ~~Investigate `custom` instance name mitigation in v2-v3 upgrade~~


krocheck:
@Julusian the only thing I think is broken is backwards compatibility in expression parsing.  I haven't been able to trace through yet where to fix it, so you might find it faster, but I'll keep at it.

docs/4_secondary_admin_controls/custom_variables.md
```diff
@@ -6,18 +6,18 @@ The key difference is that their values are set by you. This can be done either
...
+                    v      v
+               $(custom:counter)
 ```
 
 For each Custom variable, you can see and set:
 - **Current value** The current value of the variable
 - **Startup value** The value to use for the variable upon restarting Companion
 - **Persist value** Whether to persist the current value to be used upon startup. This will increase disk IO.
+
+_For backwards compatibility, all custom variables will also parse under the legacy `$(internal:custom_counter)` scheme.  This use is depcreated and expected to be removed in a future Companion release._
```
Julusian:
```suggestion
_For backwards compatibility, all custom variables will also parse under the legacy `$(internal:custom_counter)` scheme.  This use is deprecated and expected to be removed in a future Companion release._
```

Julusian (approved):

Issue #3075, opened 2024-10-06
Request: Text input in export options for custom config filenames


JaaHann:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

**PROBLEM:**
I work on lots of different events with individual companion configs. I always need to rename exported config files according to the event for future use, because the filename defaults to something like `${HOST}_custom-config_${DATE}-${TIME}.companionconfig`. Not a dealbreaker, but it annoys me from time to time.

**SOLUTION:**
A text input to enter a custom filename as an export option would be highly welcomed. Event more if it would support variables (maybe only "internal" and "custom variables") or another form of placeholders for time, date etc.

Using variables e.g., the name would still default to `$(internal:hostname)_custom-config_$(internal:date_y)$(internal:date_m)$(internal:date_d)-$(internal:time_h)$(internal:time_m)` but can be customized to something like `$(internal:custom_configname)_$(internal:date_y)$(internal:date_m)$(internal:date_d)$(internal:time_h)$(internal:time_m)$(internal:time_s)` (typical scheme I like to use).

I'm not sure what should happen if companion gets reset, a config gets imported or whether to export the current naming scheme to a config file. Maybe it would be best to have a little checkbox in all 3 scenarios to have the option to keep the current scheme after reset, overwrite a scheme from an imported config file (if available) or export current scheme to config file.

In addition some UI indicators would be a nice to have, to make this feature more convenient.
- If the naming scheme includes custom variables and the config gets exported including the current naming scheme but not with "custom variables" selected, something should inform the user about that and the possible impact.
- The same situation as described above but for importing a config. If a naming scheme includes custom variables but the config doesn't have any or available custom variables don't get imported, the user should be informed about that and possible effects.
- Some visual reminder to add custom variables used in the naming scheme after reset when still keeping the naming scheme. Or even better, keep any used custom variables from the naming scheme to ensure functionality and inform the user that those variables can't be reset because they are currently in use.

### Usecases

I for myself would manifest a workflow of adding a custom variable "configname" to dynamically change the config filename for each specific event/project and keep the naming scheme no matter if companion gets reset or a config gets imported. This way I only ever need to set this variable to a suitable name once and all exports (daily backups or whatever) will include this name.

[...]

PR #3090, opened 2024-10-16
feat: allow custom export file name

bryce-seifert:
This adds the a config field to customize export file names to aid with organization of config exports.
It partially addresses https://github.com/bitfocus/companion/issues/3075

Screenshot 2024-10-15 at 8 55 15 PM

### DIFF
--- a/companion/lib/Data/ImportExport.ts
+++ b/companion/lib/Data/ImportExport.ts
@@ -351,8 +351,10 @@ export class DataImportExport extends CoreBase {
 			// @ts-expect-error
 			const exp = generateCustomExport(req.query)
 
-			const filename = encodeURI(`${os.hostname()}_custom-config_${getTimestamp()}.companionconfig`)
-
+			const filename = req.query.filename
+				? encodeURI(`${req.query.filename}.companionconfig`)
+				: encodeURI(`${os.hostname()}_custom-config_${getTimestamp()}.companionconfig`)
+		
 			downloadBlob(this.logger, res, next, exp, filename, parseDownloadFormat(req.query.format))
 		})
 
--- a/companion/lib/Data/ImportExport.ts
+++ b/companion/lib/Data/ImportExport.ts
@@ -350,9 +350,9 @@ export class DataImportExport extends CoreBase {
 		this.registry.api_router.get('/export/custom', (req, res, next) => {
 			// @ts-expect-error
 			const exp = generateCustomExport(req.query)
-
-			const filename = req.query.filename
-				? encodeURI(`${req.query.filename}.companionconfig`)
+			let parsedName = this.variablesController.values.parseVariables(`${req.query.filename}`, null).text
+			const filename = parsedName
+				? encodeURI(`${parsedName}.companionconfig`)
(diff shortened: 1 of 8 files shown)


krocheck:
@bryce-seifert did you try to investigate variable parsing as part of adding this field?

bryce-seifert:
I did look at parsing, but was not sure how to get that functional. If you have any pointers or ideas, happy to investigate more.
I'm curious how much variables would be used, too, since this modal is ephemeral, so you'd have to enter variables every time. Perhaps that issues is solved outside of this modal, but not sure where best to do it

krocheck:
@bryce-seifert I hear you on attempting to figure out how to handle and being confused.  That's why I asked and got an appropriate response.  To properly implement to the issue referenced, we probably need a `default_export_filename` `userconfig` value with a default just as that user prescribes.

I like what you're doing here, as its attempting to respond to #3075.  I get its not ephemeral, but I'd challenge you to reach beyond that as I've described and reach out to me or Julian in Slack if you need help or inspiration.

bryce-seifert:
@krocheck I have worked on adding the default_export_filename to the user config that allows for variables, as well as added the ability to override the default and use variables in all of the export modals. Let me know what suggestions you have on the implementation 

Here's the new user config, which defaults to a similar format as the current export defaults
Screenshot 2024-10-21 at 8 48 55 PM

Here's the file name parameter in the custom export modal
Screenshot 2024-10-21 at 8 49 02 PM

And the file name parameter in the other export modal used for pages and triggers
Screenshot 2024-10-21 at 8 49 26 PM

companion/lib/Data/ImportExport.ts
```diff
@@ -174,6 +174,17 @@ export class DataImportExport extends CoreBase {
 			return instancesExport
 		}
 
+		//Parse variables and generate filename based on export type
+		const generateFilename = (filename: string, exportType: string, fileExt: string): string => {
+			//If the user isn't using their default file name, don't append any extra info in file name since it was a manual choice
+			const useDefault = filename == this.userconfig.getKey('default_export_filename') ? true : false
```
Julusian:
```suggestion
			const useDefault = filename == this.userconfig.getKey('default_export_filename')
```

webui/src/Components/ConfirmExportModal.tsx
```diff
@@ -14,9 +16,12 @@ interface ConfirmExportModalProps {
 
 export const ConfirmExportModal = forwardRef<ConfirmExportModalRef, ConfirmExportModalProps>(
 	function ConfirmExportModal(props, ref) {
+		const { userConfig } = useContext(RootAppStoreContext)
+
 		const [data, setData] = useState<string | null>(null)
 		const [show, setShow] = useState(false)
 		const [format, setFormat] = useState(ExportFormatDefault)
+		const [filename, setFilename] = useState(String(userConfig.properties?.default_export_filename))
```
Julusian: I think this should be set in the `show` function (line 63), so that it gets reset to the default each time this export modal gets shown

bryce-seifert: I think my latest commit should fix this, let me know if you had a different implementation idea

PR #3110, opened 2024-10-27
feat: make connection of action changeable, closes #1498

### LINKED ISSUE #1498: Separate Action and Instance on the buttons
Describe the feature
Ability to copy buttons and change which instance they affect without changing or reselecting the action. Sub-request, add ability to copy an action within a button (and then select the instance separate from the action.)

Usecases
I find I do a lot of work where I am copying buttons that all have the same action and/or action parameters but want to control different devices (instances) or I'm copying actions on a button to control the same function on multiple devices (instances.) As examples, copying a button which starts a cue playing to another button to start that same cue on a backup device, or in the multiple actions on a single button front starting a record across multiple record decks with a single button. If done as I'm envisioning the implementation, the action list could/would additionally be limited by the selected instance giving an added bonus to usability. With even a modest number of devices the current method of having to re-add all the actions is not only tedious but the list of actions one needs to search through can get quite large.

dnmeid:
This adds the possibility to change the connection of an action after creation.
A dropdown with all connections of the matching type is shown only if there are multiple connections of that type.
Upon change of the connection unsubscribe is called for the old instance and subscribe is called for the new instance.

image

Todo: improve positioning

### DIFF
--- a/webui/src/Services/Controls/ControlActionsService.ts
+++ b/webui/src/Services/Controls/ControlActionsService.ts
@@ -9,7 +9,7 @@ export interface IActionEditorService {
 	setValue: (actionId: string, action: ActionInstance | undefined, key: string, val: any) => void
 	performDelete: (actionId: string) => void
 	performDuplicate: (actionId: string) => void
-	setConnection: (actionId: string, connectionId: string | number) => void | undefined
+	setConnection: (actionId: string, connectionId: string | number) => void
 	setDelay: (actionId: string, delay: number) => void
 	moveCard: (dragStepId: string, dragSetId: string | number, dragIndex: number, dropIndex: number) => void
 	performLearn: ((actionId: string) => void) | undefined
@@ -21,7 +21,7 @@ export interface IActionEditorActionService {
 	setValue: (key: string, val: any) => void
 	performDelete: () => void
 	performDuplicate: () => void
-	setConnection: (connectionId: string | number) => void | undefined
+	setConnection: (connectionId: string | number) => void
 	setDelay: (delay: number) => void
 	performLearn: (() => void) | undefined
 	setEnabled: ((enabled: boolean) => void) | undefined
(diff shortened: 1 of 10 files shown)


webui/src/Services/Controls/ControlActionsService.ts
```diff
@@ -9,6 +9,7 @@ export interface IActionEditorService {
 	setValue: (actionId: string, action: ActionInstance | undefined, key: string, val: any) => void
 	performDelete: (actionId: string) => void
 	performDuplicate: (actionId: string) => void
+	setConnection: (actionId: string, connectionId: string | number) => void | undefined
```
Julusian:
```suggestion
	setConnection: ((actionId: string, connectionId: string | number) => void) | undefined
```
The `| undefined` is being considered as part of the return value currently, rather than the function being possibly undefined

webui/src/Services/Controls/ControlActionsService.ts
```diff
@@ -20,6 +21,7 @@ export interface IActionEditorActionService {
 	setValue: (key: string, val: any) => void
 	performDelete: () => void
 	performDuplicate: () => void
+	setConnection: (connectionId: string | number) => void | undefined
```
Julusian:
```suggestion
	setConnection: ((connectionId: string | number) => void) | undefined
```

Julusian (approved):
code looks good

phillipivan:
In the case of connections which define available actions based on the model of the connected device, or similar, what happens when a connection is selected which doesn't support the specific action? Does it just become undefined similar to a disabled module?

thedist:
Is there are a reason this couldn't be expanded beyond just a dropdown of connections, and actions? If feedbacks could have a connection option too, and if this option supported variables, it would mean that an entire set of buttons could be changed from one connection to another with just 1 button that cycles between instance labels in a variable.

dnmeid:
> In the case of connections which define available actions based on the model of the connected device, or similar, what happens when a connection is selected which doesn't support the specific action? Does it just become undefined similar to a disabled module?

Yes, also when the connection is disabled. It is possible to change the action to a currently disabled connection.
If there are different option choices on the different connections and the one you've chosen is not available at the new connection, you're also out of luck.

> Is there are a reason this couldn't be expanded beyond just a dropdown of connections, and limited to actions?

There is no reason, it is just one step. I think I'll do the same for feedbacks soon. According the variable, there is a different feature request about that and I think this still needs some more thoughts but maybe in the future...

PR #3119, opened 2024-10-29
feat: add HTTP API GET for module variables

### LINKED ISSUE #2830: API: GET-Requests for all variables

### Describe the feature

As already discussed in this thread (https://github.com/bitfocus/companion/issues/2575#issuecomment-1909055556) a feature to access all variables inside of companion via the API would be great.
But only GET is needed because the values are ste by the devices and only represent a live state.

Example path: `ip:port/api/<connection name>/<variable name>/value`
Example for ATEM: `ip:port/api/atem/pgm1_input/value` returning the Lable of the input routed to the ATEM's PGM out
Example for Hyperdeck: `ip:port/api/hyperdeck/countdownTimecodeHMS/value` returning the remaing playout time

If I'm not wrong at the moment this is only possible by setting a custom variable via a trigger to the value of an other variable an getting the custom variable.

### Usecases

Getting (and posting) the custom variables is already a big thing.
But for external tools (e.g. a dashboard) using the already existing numerous connections within companion would simplify programming (and reducing network load on devices because only one instance (companion) has to communicate).
Dashboards or other external tools would just communicate with the companion API to get the desired variables.

Example:
Getting states of video matrix, vision mixer, playout, recorders, etc. out of companion via API, displaying it e.g. in a dashboard or another software.
Optional via already existing API: giving values to custom variables back to companion and trigger a button to execute the desired command.

GuideOTA:
resolves #2830 

added an API GET command to return value of variables from modules rather than just custom variables

This is my first experience with GIT/TypeScript and honestly I have no Idea if what I added works or how to go about testing. any guidance would be appreciated, I am enthusiastic but ignorant.


GuideOTA:
Tested locally and the GET returns the value of module variables correctly

thedist:
For better User Experience (and any developers who have to work on your code later), I would suggest changing the naming you're using in the API and docs for `module`, as that could lead to potential confusion as you're not getting the variable values of a module your getting the variable values of a specific instance of a module based on its label.

It's very possible that a user could have multiple connections to the same module, each with different values for the same variable, so why not make things clearer that the API requires the connection label.

For consistency it would be best if you adhere to the naming scheme already used within Companion.

Julusian:
Looking good!
Could you try adding some unit tests in https://github.com/bitfocus/companion/blob/main/companion/test/Service/HttpApi.test.ts for this new endpoint? The tests for the custom-variable get will be a good reference of what cases to test and how it should behave

GuideOTA:
Thanks for the guidance! working on this issue has been a big confidence booster for me. Hopefully the first of many contributions.

Issue #3064, opened 2024-10-02
Request: Show custom variable value in collapsed view


Coupe70:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

The collapsed view of a variable is one line high and only shows the name of the variable.
There would be enough space to show the current value of collapsed variables without the need to expand them.
![grafik](https://github.com/user-attachments/assets/87929e12-a01c-4299-88eb-8271574c7ee2)
I wouldn't be against the possibilty to also edit the value in collapsed view, but my request is mainly about watching the current value. 

### Usecases

Having more variables on the screen at once and still being able to watch their current values without expanding them and having to scroll around.

PR #3127, opened 2024-11-03
feat: show custom variable in collapsed view, ui tweaks

bryce-seifert:
This PR resolves https://github.com/bitfocus/companion/issues/3064 and now displays the current custom variable value in the collapsed state.

It also cleans up the custom variable page UI to better match the other variable tabs, including:

- Adding copy button next to variable name for clearer interaction / consistency with other variable pages
- Change the copy button to the right of the value to copy the current value (consistency with other variable pages)
- Rearrange custom fields so that persist follows "current value", so it's more clear what is being persisted
- Added icon to help discoverability of tooltip for persist value

In these screenshots, old page is on the left and new is on the right 

Screenshot 2024-11-03 at 5 24 44 PM

Screenshot 2024-11-03 at 5 25 17 PM

### DIFF
--- a/webui/src/Components/CheckboxInputField.tsx
+++ b/webui/src/Components/CheckboxInputField.tsx
@@ -1,6 +1,8 @@
 import React, { useEffect, useCallback } from 'react'
 import { CFormCheck, CFormLabel } from '@coreui/react'
 import { InlineHelp } from './InlineHelp.js'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faCircleQuestion } from '@fortawesome/free-solid-svg-icons'
 
 interface CheckboxInputFieldProps {
 	tooltip?: string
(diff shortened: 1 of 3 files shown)


dnmeid:
I'd go with the value in the collapsed view but not 100% with the changes in the expanded view.
- We should keep the content of the cells top aligned and keep the layout with one header line at the top
- I like the button for copy of the variable name, but feel the copy variable value button odd. I know it was not consistent before but if we change it, we should improve and not replace one inconsistency with another one. The copy icons in the top right are usually used for something addressing the whole item, the copy icon there is usually a "duplicate...". I think we should either place the copy value icons directly behind the tetinput or omit them completely because it is quite easy to copy text from a textinput without a button. (A "duplicate variable" button btw would be a good idea but that's a different goal)
- The text of the persist checkbox should stay "Persist value". "Persist current value" for me makes it more likely to be confused with a constant.
- I'm unsure about the position of the "Persist value" checkbox. On one hand I absolutely think that it should be treated like an option and should be in the same list with Value. On the other hand do we have a lot of places where we put general options at the left side, like delay of actions, invert for feedbacks. Generally I'd like to have a more uniform and less cluttered UI, so in the end I'd prefer if all of our editors would have only one column. The question is: could this change be the start of a development towards this or should we keep it consistent for now and do the change everywhere in one go?

Issue #2816, opened 2024-03-24
[Request] Make entries in connection list fixed height


lfrisk:
### Is this a bug in companion itself or a module?

- [X] I believe this to be a bug in companion

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

On the connections page, when one or more connections are either disconnected, error or warning status, the message displayed often resizes the page as it cycles through the text that it is displaying and makes it challenging to click on other instances to edit or make changes.

### Steps To Reproduce

Disconnect a device

### Expected Behavior

Error messages should display in their own line and not push the remaining connections down on the screen, then move them back up. I would suggest adding additional text lines per each instance that can be expanded or contracted manually 

### Environment (please complete the following information)

```markdown
- OS:ALL
- Browser:All
- Companion Version: v3+ (without doing research I am guessing that v3 is when the status previews were added to the connections page (formerly instances), on macOS, Windows and CompanionPi
```

mark-epstein:
FWIW I agree that the constant moving of the page is very annoying. This is true on the horizontal axis as well.

My suggestion would be to populate the status column with only
- green check (connection OK)
- yellow ! triangle (warning)
- red X (connection disabled or status is changing rapidly)

and have the corresponding icon be a hyperlink to more information, maybe a log history only pertaining to that module

PR #3132, opened 2024-11-06
feat: standardize status to icons

bryce-seifert:
This looks to resolve https://github.com/bitfocus/companion/issues/2816 by implementing the strategy outlined in the first [comment ](https://github.com/bitfocus/companion/issues/2816#issuecomment-2019266064).

Basically, this moves all statuses to icons (previously only OK was an icon, and the rest could be any amount of text determined by the module). This uses icons to consistently and cleanly get across the status state. More info can be found by hovering over the icons, which prints out the status level and additional messages from the module.

Screenshot 2024-11-06 at 5 27 22 PM

Screenshot 2024-11-06 at 5 27 33 PM

### DIFF
--- a/webui/src/Connections/ConnectionList.tsx
+++ b/webui/src/Connections/ConnectionList.tsx
@@ -1,5 +1,5 @@
 import React, { RefObject, useCallback, useContext, useRef } from 'react'
-import { CButton, CButtonGroup, CFormSwitch, CPopover } from '@coreui/react'
+import { CButton, CButtonGroup, CFormSwitch, CPopover, CSpinner } from '@coreui/react'
 import { socketEmitPromise } from '../util.js'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import {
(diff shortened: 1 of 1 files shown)


dnmeid:
Well, maybe the icons should be in line with the icons we show in the topbar.
Also I want to throw in, that icons should not only differ in color but also in shape.

bryce-seifert:
> Well, maybe the icons should be in line with the icons we show in the topbar.

Not quite sure what you mean here, do you mean the alignment of the icons in the connection list / table header? Or the actual topbar of buttons in companion?

> Also I want to throw in, that icons should not only differ in color but also in shape.

Happy to take suggestions if you think there is a more representative icons in the library for a specific state

dnmeid:
> Not quite sure what you mean here, do you mean the alignment of the icons in the connection list / table header? Or the actual topbar of buttons in companion?

Yes, the icons on the topbar of the button. I don't say that the icons in the connection list should look like the ones we are using currently in the topbar, only that they maybe should match because they actually represent exactly the same. I'm not oposed to change the icons on the topbar. But the icon needs to work on a low dpi screen.

PR #3142, opened 2024-11-14
feat: sds firmware update

Julusian:
Inform users about available firmware updates for their connected stream deck studios

image

The notification number in the tabs bar is not something we do currently, but is important as some of the firmware updates are important so should be pushed hard on them.

In this first iteration, we are relying on an external proprietry tool by Bitfocus to perform the updates. In a later iteration, we can look into this being a command line tool that is invoked automatically.

This isn't differentiating between important and optional updates, that would add quite a lot of complexity and without public release notes for the firmware won't make much sense


companion/lib/Surface/FirmwareUpdateCheck.ts
```diff
@@ -0,0 +1,170 @@
...
+				// Inform the ui, even though there may be no changes
+				this.#updateDevicesList()
+			})
+			.catch((e) => {
+				this.#logger.warn(`Failed to check for firmware updates: ${e}`)
+			})
+	}
+
+	/**
+	 * Fetch the pauload for a specific url, either from cache or from the server
```
peternewman:
```suggestion
	 * Fetch the payload for a specific url, either from cache or from the server
```

companion/lib/Surface/USB/ElgatoStreamDeck.ts
```diff
@@ -63,6 +64,17 @@ function getConfigFields(streamDeck: StreamDeck): CompanionSurfaceConfigField[]
...
+/**
+ * The latest firmware versions for the SDS at the time this was last updated
+ */
+const LATEST_SDS_FIRMWARE_VERSIONS: Record<string, string> = {
+	AP2: '1.05.009',
+	ENCODER_AP2: '1.01.012',
+	ENCODER_LD: '1.01.006',
+}
+const SDS_UPDATE_TOOL_URL = 'https://bitfocus.io/?elgato-sds-firmware-updater'
+const SDS_UPDATE_VERSIONS_URL = 'https://builds.julusian.dev/builds/sds-test.json'
```
peternewman: Would it make more sense to serve this from GitHub raw like so?:
https://raw.githubusercontent.com/bitfocus/companion/refs/heads/main/nodejs-versions.json

Julusian: Both of these urls are placeholders currently, waiting on the update tool to be available

Julusian:
For now this has no data being fed into the update checks so will never find anything. But once the api starts reporting some real values it will begin prompting for updates.

PR #3163, opened 2024-11-29
feat: Modernize way of action advancement #2882

### LINKED ISSUE #2882: [Feat] Modernize way of action advancement

### Describe the feature

Today although the actions of a set are shown in a list, they are all executed at the same time or with fixed delays.
I think it would be a good feature if we would have the possibility of a better control of the actions execution.
Im thinking of something like the powerpoint way of advancing to the next animation.  
The user would have a choice if:
- the action starts together with the previous action
- the action starts after the previous action has finished
- the action starts after a manual trigger (i.e. the action set pauses until this happens). This trigger could be something like a button press or release or even the same logic used for triggering a trigger.

Actions can still have an optional delay, but this delay would always be relative.

Biggest advantage would be the part "after previous action has finished" making it possible to program sequential workflows without unnecessary delay times. E.g. time until a variable is set or time until http get has a result. Biggest concern here is that modules now should actively handle return values for the action callbacks, at least for actions which may run for a long time or which may fail. And we would have to deal with what happens if the action fails. Shoud it throw an error or return a special value or reject a promise. And do we just log it and proceed in the action set or give users a way to decide themselfes how to handle failed actions.

### Usecases

_No response_

Julusian:
Closes #2882, based on the work performed for feedbacks as part of https://github.com/bitfocus/companion/pull/2947

Currently implements 'internal: Action Group' and 'internal: Wait'.  
Absolute delays mode is removed, making relative the default with it possible to do workarounds to make absolute behaviour

Relative delays:
image

Absolute delays:
image

Execution order:
image


Julusian:
I think this is looking almost done.

It needs some further testing, but is functionally complete.

I almost didn't implement sequential execution as part of this PR, but it turned out to be as trivial to add as I hoped it would become (https://github.com/bitfocus/companion/pull/3163/commits/2f49dccfb9e0ce06dbae7e28d2b747ec3b307dc2)

Julusian:
@dnmeid if you could give this a quick look over the user experience of this, and confirm that it fulfills your requirements from #2882 that would be appreciated.

I am not 100% happy with how configs which are using relative delays become, but it does make sense. It becomes a set of nested action groups, one for each delay point:
image

jswalden:
Am I seeing this correctly?  If I had fifty actions of form `ACTION`, each executing 50ms after initiation of the previous one, the step would look like

* Action group, parallel, 0ms delay
    * `ACTION`
    * Action group, parallel, 50ms delay
        * `ACTION`
        * Action group, parallel, 50ms delay
            * `ACTION`
            * Action group, parallel, 50ms delay
                * `ACTION`
                * Action group, parallel, 50ms delay
                    * `ACTION`
                    * ...

and so on to a total depth of 50?  I'm kibitzing from the cheap seats, but that seems absolutely wild to me.

Julusian:
> and so on to a total depth of 50?

Correct.
I'm open to ideas on how to do this in a better way, but it needs to in some way be composed of these actions instead of the old toggle.

I am wondering how often people will have these long chains of relative delays.

dnmeid:
Sorry for the late reply, I've just seen this conversation.
I think the kind of a "Action Group" container action is a good solution, also with the parallel and sequential option.

But I'm not convinced about the delay handling.
I did already write in #3163:

>>    maybe the delays we currently have should be moved to be composite actions

> Im also not super happy with the current implementation of delays. If we give actions the choice wether to run at start or at end of the previous action, I'm wondering wether it would be better to have the delays as a separate wait action. This wait can then just be inserted where you need it and it could also easily take variables or an expression as a parameter and you can temporarily enable/disable the delay 

So still I don't see the delay as part of the action container. This would make the container the only way of adding a delay and that are two completely unrelated functionalities in one thing.
In fact actually we wouldn't need to change anything with this PR, we could just stick with the delays as they are.
If we would stick with the current delays, the action group would have its own delay value just like any other action. The delays inside of the group would refer to the start of the group. Absolute/Relative toggle could stay untouched, it's just a math thing when to fire an action.  

As I wrote in the issue (but not as a part of my original request), I wouldn't mind changing the delay handling.
I can see this as two parts:
1. getting rid of the absolute/relative toggle. I don't know how strong the feelings of the majority are about one of the two options, but in my mind I think relative is more useful and overall getting rid of the absolute option is only a small loss in functionality but a bigger win in ease of use and clarity.  
I think it is easy to remove the toggle and recaluculate the delays on the actions and reorder them where necessary.
2. getting rid of the delay field in every action. The ability to have a wait is quite important, but in my experience with many Companion users most of them are using 20% of their delays for what they are meant and 80% of their delays with only a few ms just to give the previous action enough time to finish. So I think having a delay field on every action is just using too much screen estate for only a few real delays in a configuration.  
I'd solve this with a separate "Wait" action. The wait action would always add a wait time between the previous and the following action. In a parallel execution group or top level this is like today a delay between the start of the actions (start to start). In a sequential group this would be a wait from end to start.  
The upgrade path here would just be to insert wait actions before actions that currently have a non zero delay (assuming that absolute delays have been already transformed)

companion/lib/Internal/BuildingBlocks.ts
```diff
@@ -99,6 +99,18 @@ export class InternalBuildingBlocks implements InternalModuleFragment {
...
+						type: 'dropdown',
+						label: 'Execution mode',
+						id: 'execution_mode',
+						default: 'parallel',
+						choices: [
+							{ id: 'parallel', label: 'Parallel' },
+							{ id: 'sequential', label: 'Sequential' },
+						],
+						tooltip:
+							'Using "Sequential" will run the actions one after the other, waiting for each to complete before starting the next. This doesn\'t work for all modules.',
```
dnmeid: Instead of "This doesn't work for all modules" I suggest something like: "If the module doesn't support it for a particular action, the following action will start immediately."

Julusian:
I almost did do this as a separate wait action, and this action-group action can be used in that way if desired. But I think that doing it that way makes for some really verbose usage.

For example, replicating some absolute delays today
```
action-group (parallel):
  action-group (sequential):
    delay 100ms
    some action
  action-group (sequential):
    delay 1000ms
    action-group (parallel):
      some action
      other action
```

Or some relative delays where we don't want to wait for the action being delayed to complete
```
action-group (sequential):
  delay 50ms
  action-group (parallel):
    some action
    action-group (sequential):
      delay 50ms
      action-group (parallel):
        some action
        action-group (sequential):
          delay 50ms
          some action
```
This scenario is one where I am being a bit cautious. I don't want to block on 'some action', because behaviour in 3.4 is to not do that, and adding that extra delay will result in these delays being longer than expected.  

In both these cases, I had to add extra action-groups to be able to wait on a delay, because a delay inside of a 'parallel' action-group doesn't do anything, but I want to be in parallel mode except when using a delay.  
So my logical extension of this was to make it so that the delay action could have child actions, which then became the action-group.


I am wondering if a better way to handle this is a 'non-blocking' checkbox on the action-group. That would allow this to be written as
```
action-group (sequential):
  delay 50ms
  action-group (non-blocking):
    some action
  delay 50ms
  action-group (non-blocking):
    some action
  delay 50ms
  action-group (non-blocking):
    some action
```
Only slightly less wordy, but much less nesting. Not sure if it makes sense for an action-group with a delay though; should it disable blocking only for the actions or for the delay portion too? I would say all of it, but that keeps this sequence of short delays wordy.  
Perhaps this should be its own action, with the execution mode property?  
Maybe something along these lines is what is needed so that chained relative delays can avoid nesting?


> I'd solve this with a separate "Wait" action. The wait action would always add a wait time between the previous and the following action. In a parallel execution group or top level this is like today a delay between the start of the actions (start to start). In a sequential group this would be a wait from end to start.

I don't like this. It makes the 'parallel' execution mode be parallel except when a 'wait' is found where it will then do something else. This makes the name 'parallel' be misleading, so at the very least this should be given a better name.  

I think that using these waits during parallel is quite different behaviour to how things behave today. It effectively makes the default behaviour be relative delays (absolute is the current default and makes setting up an absolute delay feel like a workaround. For users who aren't reading changelogs, will they notice the switch from absolute to relative?  

These cases are all using absolute delays.
Perhaps I have:
```
some action (delay 1000)
some action2 (delay 1000)
other action (delay 0)
final action (delay 100)
```
if we want to avoid changing the order of user actions, then one way to express this in your proposal would be (assuming root is parallel execution)
```
action group (parallel)
  wait 1000
  some action
  some action2
other action
action group (parallel)
  wait 100
  final action
```
yes that second action-group is unnecessary, but it will keep the upgrade logic simpler. We need to do the same fixup for data coming from presets, so this is a mess we will keep on making for some users.

dnmeid:
> It makes the 'parallel' execution mode be parallel except when a 'wait' is found where it will then do something else. This makes the name 'parallel' be misleading, so at the very least this should be given a better name.

You're right. Maybe 'non-blocking' is the better word. As I said the difference would be start-to-start vs. end-to-start from action to action.
Maybe 'parallel' is misleading anyway, even today. The module is run in one thread so the actions can't be started really in parallel. They are started one after another and usually they are not using async stuff a lot so technically often one action is also finished before the next one starts. So the perceived functionality of start them all together isn't there.

> For example, replicating some absolute delays today
```
action-group (parallel):
  action-group (sequential):
    delay 100ms
    some action
  action-group (sequential):
    delay 1000ms
    action-group (parallel):
      some action
      other action
```

No, I'd expect it to be different.
Today we don't have a sequential mode and parallel is how it works, so the upgrade would never need to introduce any action-group at all assuming that the root stays parallel. Some delayed actions would simply transfer to:
```
 some formerly undelayed action
 wait 100ms
 some formerly delayed 100ms action
 wait 400ms
 some  formerly delayed 500ms action
 other  formerly delayed 500ms action
 wait 50ms
 some formerly delayed 550ms action
```

Definitely that could lead to the need of resorting actions if someone is using absolute delays and has actions with a higher delay sorted more to the top than actions with a lower delay. I can think of only very rare cases where you would do that intentionally. Actually these cases would be simplified with the introduction of action-groups and your last example shows exactly how I envision how, but I wouldn't go as fancy and decide wether such an arrangement is intentional or just bad and so I would really just go with the resorting if needed. Additionally I guess most of these special arrangements would even more profit from the sequential flow and could avoid some of the delays then, but again, that should be rewritten by the user and not be an automatic upgrade. I think most users would profit and the few who did the strange order on purpose end up still with their original functionality and can rebuild the structure of your example themself.

> I think that using these waits during parallel is quite different behaviour to how things behave today. It effectively makes the default behaviour be relative delays (absolute is the current default and makes setting up an absolute delay feel like a workaround. 

Yep, that would make relative the default and basically takes away absolute at all. When we introduced the delays it was another one of that fast and not so well thought out solutions. It was just a UI to a `setTimeout()` per action, so of course it was absolute. Also by that time actions were not resortable. But soon users demanded the relative workflow.  
So today I think especially absolute delays and the resortability can lead to these situations where actions are not in order of the execution. In maybe 90% of the cases this not intended and a source of errors. With taking away the absolute delays we would avoid that users can ever maneuver in such situations and I think this is an enhancement.

> For users who aren't reading changelogs, will they notice the switch from absolute to relative?

Yes, especially if it is called "wait" instead of "delay". New users will automatically get that it is relative because they expect everything to be sequential anyway. Old users will immediately see that the delay option per action is missing. They are forced to find out how the new solution is working and that they have to add a separate wait action. If they had some delays in their configs, they will find the new waits and I think that should be enough for them to find out how it's done.  
It's not a bigger change than from the old toggle button checkbox to steps. There will be some people who don't get it, as usual.

dnmeid:
And here is another argument for dropping absolute delays:

When we introduce action-groups, the delay of stuff inside of a group would always be relative to the start time of the group itself and be only absolute inside of the group. I think this will cause a lot of confusion. Users will not be able to get any more when something is happening.

Julusian:
> You're right. Maybe 'non-blocking' is the better word.

By non-blocking I mean to get the start-to-start behaviour you mention. I think of this as being a checkbox on the action group, to change how the parent group treats this one.
In effect, it turns the action-group into a setTimeout that is then forgotten about.

This matters because a 'sequential' group with just a 1000ms wait inside will take 1000ms to execute. The same is true with a parallel group.
And I think there are cases where you want to be running a sequence, and to spawn off this other sequence of actions but you don't want it to block the execution where it was spawned from.

> Maybe 'parallel' is misleading anyway, even today. The module is run in one thread so the actions can't be started really in parallel. They are started one after another and usually they are not using async stuff a lot so technically often one action is also finished before the next one starts. So the perceived functionality of start them all together isn't there.

That is kind of true, but not exactly.  
For any module which is just fire and forget, then yeah it wont really do anything in parallel. But as they are in different threads, depending on random delay/latency the order that the actions execute in can be unpredictable.
Some modules do wait for the device to ack the action during which time the module can execute more actions, and some of those do report that execution time back to companion by propagating a promise.
The sequential mode waiting for the action promise to resolve but parallel will allow for these modules which do have this async behaviour to run multiple at once.

bmd-atem is a good example of this, it propagates the promise for each command it sends, which will only resolve once it receives the ack packet from the atem. So parallel and sequential will at times behave noticeably differently with this module.
I expect that more and more modules will want to start doing this waiting too, so that the sequential mode will become better over time. 

> No, I'd expect it to be different.
Today we don't have a sequential mode and parallel is how it works, so the upgrade would never need to introduce any action-group at all assuming that the root stays parallel.

Only if we are ok with changing the numbers that a user has written as absolute to be relative. I'm not entirely against that, but am not entirely convinced by it.

> Definitely that could lead to the need of resorting actions if someone is using absolute delays and has actions with a higher delay sorted more to the top than actions with a lower delay. I can think of only very rare cases where you would do that intentionally. 

I'm not ok with reodering users actions. I can think of some scenarios where they would have chosen this out of order intentionally.
Perhaps the following, with user defined summaries:
```
'Start A' (delay 0)
'1' (delay 100)
'2' (delay 500)
'Reset' (delay 1000)
'Start B' (delay 0)
'1' (delay 100)
'2' (delay 500)
'Reset' (delay 1000)
```
If we reorder those, then the title/headline they gave each action loses its meaning. Would they have used absolute instead of relative delays for this? I don't know, but it could be either.

> With taking away the absolute delays we would avoid that users can ever maneuver in such situations and I think this is an enhancement.

fyi we won't avoid this situation entirely, even with your proposal the user could still do:
```
group
  if X
    wait 1000
    action (which will execute at 1000)
  else
    wait 100
    action (which will execute at 100)
  if Y
    wait 100
    action (which will execute at 1000)
  else
    wait 1000
    action (which will execute at 100)
``
and I think we need to allow this, otherwise the system will have some annoying limitations.


> When we introduce action-groups, the delay of stuff inside of a group would always be relative to the start time of the group itself and be only absolute inside of the group. I think this will cause a lot of confusion

In a similar fashion, I think this confusion is unavoidable. Even with your proposed wait action, it is still relative to the previous action/wait in the group. But there can be other groups running in parallel to the current one, with their own sequence of waits.


> It's not a bigger change than from the old toggle button checkbox to steps. There will be some people who don't get it, as usual.

True

> Yes, especially if it is called "wait" instead of "delay".

I'm ok with calling it wait to help differentiate it to before.

But I still don't think I am fully convinced by this hybrid parallel but not fully parallel behaviour. While I can see the appeal for users, I am worried that some users will find it confusing, because it is an opinionated 'simplification'.

It definitely needs a name other than parallel, maybe simply calling it 'default' would be sufficient, or this could be a checkbox called something like `wait for previous action to complete`? I do prefer a dropdown though, it makes it simpler to add more modes which I wouldnt be surprised if we do later on.


90% of the time, the defaults will probably be used. But I think that these options for controlling execution and branching will be really useful especially for triggers and once we have if statements.

Honestly that pitch someone did for wanting a flow diagram of action execution is now starting to make some sense. It didn't really before, but it is now.

**Comment by dnmeid** [comment]
> This matters because a 'sequential' group with just a 1000ms wait inside will take 1000ms to execute. The same is true with a parallel group.
And I think there are cases where you want to be running a sequence, and to spawn off this other sequence of actions but you don't want it to block the execution where it was spawned from.

Now I get it. The execution mode is how to treat the actions inside and there is another option with no title yet about when the container itself returns and advances to the next action.  

My assumption has been that the outer scope declares how the children behave. If an action is used in the parallel/non-blocking flow, it is just triggered and then the next action is triggered and so on. If it is triggered in the sequential flow, it is triggered and then the next one is triggered after it has finished. Likewise if a group is used inside a parallel flow it is triggered and then whatever happens inside of the group is not been waited for.  
Example:
```
Root
  group-sequential
    long running action 1
    long running action 2
    wait 1000ms
    action 3
  group-sequential
    long running action 4
    long running action 5
  action 6
  wait 500ms
  action 7
```

As root is parallel it will start the two groups and action 6 as simultaneously as possible and action 7 after 500ms.
Both sequences in the groups will run independent. E.g. we don't know when action 3 will be triggered because it depends on running time of action 1 and 2. Action 5 could be triggered before action 7 if action 4 if finished <500ms and after action 7 if action 4 needs >500ms.

```
Root
  group-sequential
    group-parallel
      long running action 1
      long running action 2
      wait 500ms
      action 3
    action 4
  wait 1000ms
  action 5
```

Here the sequential group is started at 0 but the root context don't care about its running time because itself is parallel. So action 5 is started at 1000ms absolute. Inside the sequential group there is a parallel group that starts action 1 and action 2 simulataneously at absolute 0ms and action 3 at absolute 1000ms. Action 4 will be started after the parallel group has finished which is at whatever happens later, either the 500ms wait + action 3 or the finish of action 1 or the finish of action 2.

> It definitely needs a name other than parallel

Do you think 'asynchonous' is too technical? Or maybe really use the more descriptive 'start-to-start' and 'end-to start'.

> fyi we won't avoid this situation entirely

Absolutely, I didn't want to say that the order of appearence always needs to be the order of execution. It is inevitable that there will be situations where this is not the case. It may not always be intuitive but it is logical and the only possible solution. My intention was to avoid situations where you can accidentially mess up your flow. If we have wait as a dedicated action, it will matter where it is positioned.

> I'm not ok with reodering users actions.  
...  
If we reorder those, then the title/headline they gave each action loses its meaning.

Yeah, exactly this is the only example I can think of where that strange order makes sense.  
Two points here:
1. I'm not aware of how many users are using that pattern.
2. This pattern is the current workaround for what we are trying to improve with multiple sequential groups inside of a parallel context
The question is how to solve this dilemma. I'm sure there are some users utilizing this pattern and I'm sure there are some users with just weird messed delay times and orders.
The only two possibilities I see at the moment is to either reorder the actions and avoid introducing groups automatically and disturb the users with the intentional pattern or to keep the order of actions and insert a group whenever we find a order/delay discontinuity and disturb the already lost messed-up users even more.
A third option would be to give a choice during upgrade but we didn't do that so far and I'm sure there will be users that don't get it and click on the wrong option and then complain later and bla bla bla.

I'm fine with any of these options, but as I said, I don't know how many users re there with such configs and I'd prefer to disturb the smaller amount of users.

Julusian:
> Now I get it. The execution mode is how to treat the actions inside and there is another option with no title yet about when the container itself returns and advances to the next action.

Yes, exactly that. I think it is important for us to keep the behaviour this way, to avoid complexity spiralling when more types of actions that can influence behaviour get added.  
This way, the behaviour is entirely self contained, its just another action that runs and runs other actions without needing to care what they are or do.

Ill admit, I am doing this from a very programmer point of view, so to me this composition/nesting of functions makes perfect sense, and makes for a nice implementation.

Potential name for that checkbox: 'Wait for completion' (default true)?

> The examples

I agree with those examples, assuming we decide to go with your proposal for how waits behave.

> Do you think 'asynchonous' is too technical? Or maybe really use the more descriptive 'start-to-start' and 'end-to start'.

Yeah I think 'asynchonous' is too technical, but 'start-to-start' and 'end-to start' don't feel clear enough to me. They could be phrased as `Start each action: 'with end of previous'/'with start of previous'`, but it would still need a disclaimer that waits break this rule.  

Perhaps just simply 'Parallel with waits' is good enough? Or we could not name it, and have a checkbox 'Run actions sequentially' to enable the sequential mode.  
I do keep coming back to keeping it vague and just calling it 'default', (more could be shown on the tooltip), if we think that this is how users would expect it to behave.  

> I'm fine with any of these options, but as I said, I don't know how many users re there with such configs and I'd prefer to disturb the smaller amount of users.

I don't think a choice during upgrade will be nice. These sequences could appear in presets too, where a choice would be annoying so we would need to make a decision for that scenario anyway.

But I still think that 2 is best. I worry that reordering would result in bug reports from users who don't realise that this reorder is intentional, and that could go on for years until we break the module api and change it there too. So while adding these groups isn't the prettiest, it is very clear that they are fighting the system a bit, and they either need to live with the mess or clean it up.  

With the 'blocking waits', the amount of these should hopefully be minimised.  
And it could probably be more intelligent on reinterpreting delays; I am not entirely convinced about doing it for user buttons, but changing the numbers in presets to be relative I am very tempted by.

dnmeid:
Hm, even if I now see what your intention is with that checkbox (or maybe dropdown), I still rather would say that it is not needed.
In my examples I wrote what I'd expect to happen by default even without the need of that differentiation.
The mode of the container decides how its content is treated. All actions in a parallel container are just started and we don't care how long they need to finish. The same should be true if there is another container in the container, it is just started and we don't care how long it takes to finish. 
All actions in a sequential container are started and waited for finish sequentially. If one of these steps happens to be another container it should be the same.
If I get the idea of the checkbox right, you could have a parallel container and inside some actions and another container but the inside container is switched to "blocking" and then would essentially work like a wait.
I think this would lead to a complete mess if we say the mode of the container decides how the content it treated but if it is another container, you may override it.

So my go for the topics in question is now:

- There is one type of action-group with one option: `parallel` or `sequential` (wording still tbd). Like with the actions itself its own duration is only taken into account in a sequential surrounding
- Delays are replaced by a `wait` action. In a parallel surrounding the wait adds time between the start times of the adjectand actions (start-to-start). In a sequential surrounding the wait adds time between the end of the previous and start of the following action (end-to-start). So the `wait` would in the parallel surrounding brake the rule that we don't wait for its completion, but that seems to be completely logical for me.
- Existing configurations are upgraded in a way that whenever an action further down has a smaller absolute amount of delay than the previous, a new action group will be added so actions are not reordered. Preset definitions with delays are treated the same.

Julusian:
Lets drop the non-blocking for now. I think its useful, but its not essential to this being merged. When/if someone asks for it, it can be re-discussed.

How about 'Burst' and 'In order' for the naming?

Julusian:
the output of these upgrades is looking a bit more sane now, going at most one level deep.  
This was a button using absolute offsets, with a drop to 0 partway through
image

dnmeid:
For the syncronous mode I think 'Sequential' is better than 'In order' because it makes it more clear that one action has to finish before the next can start.
I'm not a native speaker but I think 'Burst' does fit better than parallel. But actually to be precise the wait is breaking the rule and not everything is in parallel or burst. Actually what we want to say is, that things are triggered and we don't wait for their finish, but we wait for `wait`. So what do you think about "Scheduled"? It says exactly that the action starts when it is time for it to start.

Julusian:
I'm happy with Sequential.

I think this is becoming a case of any word which describes how it might respect the waits is unclear on how the other actions are run, and any word which describes how the other actions are run ignores the special behaviour of the waits.  
To describe both, we need a short phrase.

I do think burst makes sense, as you get bursts of actions between the waits.  I feel like scheduled has implications of larger timing than within a button. (also we still call the triggers system 'triggers and schedules', so I worry about user confusion with that).
Maybe combine them to 'Scheduled bursts'?

dnmeid:
I've been discussing this with ChatGPT, for language it is quite good. The suggestions there are "Concurrent", "Parallel" and "Chunked"
> I recommend "concurrent" as the most accurate one-word adjective for your described behavior. It precisely captures the idea of multiple actions starting simultaneously, with the flexibility to include potential breaks between action groups.
If your focus is on the idea of grouping and managing actions rather than their simultaneous initiation, "chunked" is an excellent choice! While "parallel" has a similar meaning to concurrent it is slightly more general. It is often used in computing and process descriptions.

dnmeid:
OK cool. So it is `Concurrent` and `Sequential`.

And I also have an idea about a third mode: `Inherited` which actually inherits or doesn't change the mode from the outer scope/parent. This could be the default value, so users only have to touch the option when they deliberately want to change the execution mode. The upgrade script should still set the mode of needed groups to concurrent to guarantee the same flow.
I think inherited as the default option will especially be useful because it adds a new usecase to the group: just grouping actions without caring of execution mode. Also I'm thinking of making the default execution mode (the root) an user config option but that is not necessarily a part of this PR.

I'm fine with this in the beta, I think it is a very good addition to the next release.

Julusian:
> And I also have an idea about a third mode: Inherited

Maybe. I can see value in it, but need to think on it some more

> I think inherited as the default option will especially be useful because it adds a new usecase to the group: just grouping actions without caring of execution mode

This is a good argument for it. Being able to group and tidy up long action sequences.

> Also I'm thinking of making the default execution mode (the root) an user config option but that is not necessarily a part of this PR.

I am open to this, but not 100% convinced. If its a user config field, I think that will have to be a 'default' value that is then a field per button. Otherwise it will be very easy to break your config by changing that, or break pages when importing them because they were built for the other mode.
Then I worry about whether that would be adding too many fields to the buttons. Especially as this sequential execution is on the more 'advanced' side of companion usage, so I like the barrier of shielding new users from it. 
This is especially true when we end up with more config fields on the action-groups, should they also be on the button itself? if not then it could result defeat the purpose of having the root configurable.
(One field I have in mind is 'stop on error'. Maybe that won't be the one, but I expect there will be something at some point)

And if its on the root, does that mean that things like the if statement should also have it configurable? Doing so will really pile on the complexity of the implementation, and unless its configurable for each of if and else independently still wont solve everything. (I am wondering if things like the if statement should be fixed at the 'inherit' you propose. That avoids cluttering them with fields to control this, and avoids fighting a fixed behaviour)

So I am concerned that doing this will be extra complexity/confusion in both the implementation, and for all users, to serve a small portion who want this, when it could also be achieved with a couple more clicks by those users.  
Lets wait and see how users find this iteration before deciding.

dnmeid:
Happy to have this next step towards a Turing-complete product merged! @Julusian Thanks for implementing and also thank you for the good discussion.

Regarding the `Inherited`, I found another argument. Especially in regard to the future extensions. We are now for the first time introducing container actions. Future things like if/then/else, try/catch or loops will also be containers. I would prefer to have on all of them only the minimum needed options to fullfill the purpose of the container to not mix functionality of two separate tasks in one item.  
If we now teach the users that action groups have an option to change the execution mode and that is actually non-optional, they are more likely to expect that option on all containers in the future because the name is "action group" and from that name the purpose is grouping and not changing mode. If we have `Inherited` as the default, they will learn that this functionality truely is an option. Even if I'd say changing the mode is going to be the main use of the action group, we won't introduce an expectaion of execution mode handling for future containers. I think for stuff like if/then/else the inheritance should be the default, non-changeable mode and if the user wants to use a specific mode he'd have to surround it with an action group.

Regarding the default mode configuration, I totally agree. Let's wait with it for some user feedback.

Julusian:
That is a good argument. I'll add this 'inherited' mode

krocheck:
@Julusian I really like how this turned out.  Excellent work and thanks for doing this!  I had some notes in realtime for the discussion but then was noticing my notes were getting worked out the way I'd suggest anyway.

One cleanup note is that I think users around going to be confused with the delay field going away, but especially since the replacement action is called "Wait".  I went into the action browser and searched "Delay" and was confused why something wasn't coming up.  I tried adding "delay" into the description language but it appears we don't use that for the search in the actions browser.  I feel like a little hover hint somewhere/somehow to tell users about the new wait action would be a good thing.

Also, the documentation is going to need to be updated to reflect these changes.

Julusian:
A nice effect of this which I am noticing now that I am using this with some real config and devices, is that my 'slow' actions (spotify: play/pause, sonos: load stream uri) now show the little play icon while they are executing, right up until the feedback changes to reflect the change. 
Not something I considered, but I like it. For me its a nice confidence factor of 'its thinking about it' instead of being unsure if I should wait or try again.

And this is happening because of those modules are awaiting their http/whatever request inside the action callback, and have been doing so for a long time.

-----

> One cleanup note is that I think users around going to be confused with the delay field going away, but especially since the replacement action is called "Wait"

Thats a fair point. Perhaps the action should be called wait/delay? Would that still be distinct enough from the old field while still be easily discoverable?

> I feel like a little hover hint somewhere/somehow to tell users about the new wait action would be a good thing.

An idea on where/how to do this would be good. I suppose we do have space to 'waste' on this for now, the actions are still in the split layout (except for the new action-group).
But should it be a blue 'alert', or a `Delay has moved (?)` style thing?
I guess it would stay for 3.5, and maybe 3.6 then be removed?

> Also, the documentation is going to need to be updated to reflect these changes.

Tbh it probably needs another round of general updating anyway. I don't think I updated it for the dynamic page count either..

dnmeid:
I don't have a strong oposition against a hint of the new wait but actually I'd just rather make the cut and that's it.
Of course do we need to update the documentation.  

All users with existing delays in their configs will find waits after upgrading, they will immediately know what's going on, how it's called and how to work with even without reading a release note.

All new users will need to read the getting started anyway.  

All old users that occasionally used delays but don't stumble upon a converted one are out of luck. They need to read the changelog or the documentation or they have to ask on facebook. I guess there will be maybe 20 guys in the first months after release that ask on facebook what happend to the delay. I don't think we necessarily need to make more hints that will go away later.  
I'm rather looking forward to ginving good use to the new space.

> perhaps the action should be called wait/delay?

Again I think the confusion will be moderate and calling it wait/delay would only make it worse. The point is that wait is now something different then delay was and the new name will make that more clear and will help talking about it and adressing it. With a wait/delay I think there will be even more confusion.  
The only thing I can imagine is that we also introduce keywords to actions like we have for connections. Then we could give additional search terms and wait can show up if someone searches for delay. On the other hand a lot of audio gear is using some stuff with delay and probably has some actions with delay in the name and I think overall it is better if users are searching for delay that there will be only actions with delay. Users just have to learn that it is now called wait and that is the best name.

krocheck:
In general I was nit-picking and everything's probably fine as-is.  As indicated, we just need to communicate and answer the inevitable questions.

I can just hear my dad in my ear like every time there's an update to his phone where something moves/changes and his reaction is, always: "Why do they feel the need to change crap and move things around?  It's not better.  I swear they change things just to keep people confused."  LOL. That's clearly not the case here, but I just saw it through that lens for a moment.

thedist:
This update seems to have some bugs with how it's upgrading existing setups with delayed actions.

I had some buttons where on press they would change page, but if I held then after 1 second it would flash green then back to its original colour, and go to a different page on release.

In the latest beta build, the long press works as intended where after 1s it flashes green and goes to the correct page on release. But for a short press it changes page as intended but also sets the button to green after 1s and never sets it back to the original colour (and these colour changes should never even fire as the short release has abort delayed actions, which previously worked as intended but now is broken).

kenj1:
I do not believe this solves the request I made in #3222.  

For example, how would I be able to use the predefined variable, $(mitti:currentCueTRT_s), in a Delay for an Action?  This predefined variable comes from the Companion module Imimot:Mitti and specifies the Total Run Time of the current video playing.

Would I need to create another custom variable that points to the Mitti variable??  Please explain so I understand how it satisfies my request.

Thank you.

Julusian:
> For example, how would I be able to use the predefined variable, $(mitti:currentCueTRT_s), in a Delay for an Action? This predefined variable comes from the Companion module Imimot:Mitti and specifies the Total Run Time of the current video playing.

Something like this:
image

What I have done here is weird, but it will do the wait based on whatever that module variable is.

A slightly less weird example:
image
In this, it will wait for however long the `mini` device has been streaming for

PR #3183, opened 2024-12-17
feat: add 'Conditionalise existing feedbacks' feedback

### LINKED ISSUE #2800: Conditionalise "Use another button style"

### Describe the feature

Is it possible to enhance the "Use another button style" by putting a check mon a boolean condition.

### Usecases

Use case:

I have a number of buttons that currently have 3 feedbacks on them and set the style of the button depending upon a set of conditions. If the style of one of these conditions changes then all of the buttons have to be altered. It would be a lot easier to say - if <condition> then <use abother button style>. That way only one button needs to changed and the change in style automatically ripples through.

Julusian:
Extends the work from #2947 so that it can be used in combination with 'advanced' feedbacks from modules.

Resolves #2800 

image

This is a semi hardcoded to there being 'children' and 'advancedChildren' for internal feedbacks, with other parts of the implementation being generic. I don't feel like this is worth unifying into a truely generic implementation at this time, most of the effort in doing so is the execution (which has to be specific to this feedback) and fixing up existing configs (as that would require changing the structure of the 'children' property added in #2947)

### DIFF
--- a/companion/lib/Controls/Controller.ts
+++ b/companion/lib/Controls/Controller.ts
@@ -323,7 +323,7 @@ export class ControlsController extends CoreBase {
 			}
 		})
 
-		client.onPromise('controls:feedback:add', (controlId, parentId, connectionId, feedbackId) => {
+		client.onPromise('controls:feedback:add', (controlId, ownerId, connectionId, feedbackId) => {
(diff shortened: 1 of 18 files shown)


Julusian:
Combined into #3185

PR #3257, opened 2025-02-06
feat: Improve set variable performance

thedist:
Part of a set of changes (the other being to companion-module-base) to improve the performance of modules setting variables.

Currently, modules set variable values as an Object of key/value pairs, such as `{ someVariableId: 'variableValue' }`, but the companion-module-base transforms this into an array such as `[{ variableId: 'someVariableId, value: 'variableValue' }]`, and then after being sent from the child process to Companion it has to be transformed into `{ someVariableId: 'variableValue' }` again. This iteration over the variables adds avoidable overhead.

A simple test I set up to generate an arrays of set amounts of variables took this long to transform to an object:
1,000 variables: 0.2ms
10,000 variables: 1.8344ms
100,00 variables: 23.3069ms
1,000,000 variables: 148.7271ms

Compared to if the object sent by the instance was already an object
1,000 variables: 0.0024ms
10,000 variables: 0.0006ms
100,000 variables: 0.0006ms
1,000,000 variables: 0.0014ms

This PR should be fully backwards compatible, so all existing modules would continue to send arrays of variables and need to be transformed, but any module that updates their companion module base (with a PR that will be coming shortly), it would be sending an object and bypass the transforming completely.

### DIFF
--- a/companion/lib/Instance/Wrapper.ts
+++ b/companion/lib/Instance/Wrapper.ts
@@ -698,9 +698,14 @@ export class SocketEventsHandler {
 	async #handleSetVariableValues(msg: SetVariableValuesMessage): Promise<void> {
 		if (!this.#label) throw new Error(`Got call to handleSetVariableValues before init was called`)
 
-		const variables: Record<string, CompanionVariableValue | undefined> = {}
-		for (const variable of msg.newValues) {
-			variables[variable.id] = variable.value
+		let variables: Record<string, CompanionVariableValue | undefined> = {}
+
+		if (msg.newValues instanceof Array) {
+			for (const variable of msg.newValues) {
+				variables[variable.id] = variable.value
+			}
+		} else {
+			variables = msg.newValues
 		}
 
 		this.#deps.variables.values.setVariableValues(this.#label, variables)


Julusian:
I think the reason I did this is that sending `{ abc: undefined }` over IPC will likely result in `abc` disappearing from the object entirely. Passing it as an array avoids this.

What would be the impact of leaving the object -> array conversion inside module-base, and passing it around as an array inside companion? We are simply iterating over the whole object, so it could easily remain as an array (will require some simple fixing up of other callsites of that method)

I suppose another hazard that would be nice to avoid is if a 'batch' of changes includes changing the same variable multiple times. It would be nice to optimise this out, but maybe we can reason that this should never happen, and that its a bug in module-base if it does

thedist:
One of the reasons I didn't want it to remain an array is the performance improvements I'm testing in companion module base involves using a Map rather than an object (but trying to send a Map over IPC results in an empty object `{}`, and transforming it to a Map rather than an Object on Companions side didn't show much in the way of performance gains (at least not with the current setup).

jswalden:
Letting the internal value encoding (or encoding constrained by IPC functionality) bleed into the user-facing UI seems like kind of a mistake.  If you're going to require modules to affirmatively opt in to the newer, faster UI, maybe it makes sense to add a new, parallel UI -- and then you can implement the old slow UI in terms of the new one.  (And by "new UI" maybe that also includes separating the variable-setting case from the variable-deleting case int two new functions.)

thedist:
> Letting the internal value encoding (or encoding constrained by IPC functionality) bleed into the user-facing UI seems like kind of a mistake. If you're going to require modules to affirmatively opt in to the newer, faster UI, maybe it makes sense to add a new, parallel UI -- and then you can implement the old slow UI in terms of the new one. (And by "new UI" maybe that also includes separating the variable-setting case from the variable-deleting case int two new functions.)

What user-facing UI are you talking about? This has nothing whatsoever to do with UI, or anything user-facing at all so I'm confused about your comment.

thedist:
Converted to Draft while I'm working on some changes on the companion-module-base to assist in this.

jswalden:
> What user-facing UI are you talking about?

Sorry, by "user" I meant "module developer" as opposed to "Companion implementer".  Didn't think of the obvious confusion that word would have here!  In my head modules use Companion, so they are "users", even if that's plainly not the best name to be thinking of here.

thedist:
Well in terms of integration for Module Developers, I personally don't see what benefit it would have to split this into multiple ways for developers to have their module send variables to Companion.

Currently, module developers use `instance.setVariableValues` and pass it an object.
From there companion-module-base (still within the instance of that module) iterates over it and makes an array of objects which is then sent to Companion over IPC.
Companion takes that array and turns back into an object, and processes the variables internally within Companion.

The changes I'm working on are to 1. lower the overhead between companion-module-base and Companion, this would be entirely transparent to Module Developers, you'd still set your variables exactly the same way. and 2. updating `setVariableValues` within companion-module-base to support a Map, as on the module developers side this can have performance gains when generating the variables and values before they are even sent to `setVariableValues`, but currently a Map has to first be changed to an Object before passing to `setVariableValues` which has some overhead itself.

What will this mean for Module Developers?
if you want to send an object to `setVariableValues` like all modules currently do, that'll work without changes.
If you want to send an object to `setVariableValues`, but have a version of Companion and use companion-module-base with these changes, you'll have performance benefits without having to do anything.
If you want to send a Map to `setVariableValues`, that'll work too with a version of companion-module-base that supports it.
If you want to delete variables, again, use `setVariableValues` the same way it has been already and just use `undefined` as a value, how they are deleted would all be handled behind the scenes as Module Developers don't need to get into the specifics of it.

So my plans are to be full backwards compatible, no need for separate functions to send variables using the new way, no need for a separate function to delete variables.

Julusian:
@thedist That all sounds fine with me.

I am still curious on this though:

> What would be the impact of leaving the object -> array conversion inside module-base, and passing it around as an array inside companion? We are simply iterating over the whole object, so it could easily remain as an array (will require some simple fixing up of other callsites of that method)

How much does it cost to do the object/map to array conversion there, especially compared to alternatives. It sounds like its going to have to do some kind of conversion (to handle both object and map, and to preserve setting to undefined), so it feels like keeping it transported as an array could be the best option.  
The key thing we can do is minimise the cost inside companion, as the array to object change is pretty pointless.

If you are going to change the ipc format, dont feel like you need to make it work within the existing ipc message. Adding a new one with a different parameter layout is perfectly fine. I did this already in I think 1.2 for how we send updated configs to modules

thedist:
This is some testing I did, essentially breaking down each function from generating variables within a module, to passing to the module base, stringifying, then on Companions side parsing and checking for changes, plus any conversions that need to happen for different combintations.

Essentially what it seems like is the Maps are the fastest to work with, but all performance gains are lost by either needing to transform the map into another form (such as when using ejson), or if using superjson the stringify is about 3 times longer so all gains are lost there.

I've tried several variations of how variables are generated, at which point they are transformed, different stringify libs, and haven't been able to find any changes to the current setup that would make a significant difference. The only thing I can suggest right now is modules will need to work to reduce the number of variables used (or at least provide users options to enable/disable sets of variables) to reduce the total number being processed, and secondly would be to perhaps rethink how we handle variables entirely to see if there's any alternative ways it could be done.

## Module
| Function | 1,000 | 10,000 | 100,000 | 1,000,000 |
| ----- | ----- | ----- | ----- | ----- |
| Generating Variables as Obj | 0.997ms | 4.0264ms | 53.7642ms | 565.5797ms |
| Generating Variables as Map | 0.1418ms | 1.5503ms | 14.7072ms | 259.631ms |
| Converting Map to Obj | 12.1193ms | 3.5406ms | 183.9174ms | 414.7543ms |
| setVariableValues with Obj | 0.5981ms | 4.8589ms | 53.8081ms | 1007.8314ms |
| setVariableValues with Map | 0.2899ms | 3.0237ms | 23.6834ms | 359.5598ms |
| Converting Map to Array | 0.1228ms | 0.6213ms | 5.483ms | 251.4124ms |
| ejson stringify with Array | 3.0806ms | 12.4404ms | 79.5981ms | 866.2747ms |
| Superjson stringify with Array | 7.3859ms | 27.8722ms | 202.8095ms | 2470.637ms |
| Superjson stringify with Map | 1.7032ms | 19.6336ms | 201.4564ms | 2432.1415ms |


## Companion
| Function | 1,000 | 10,000 | 100,000 | 1,000,000 |
| ----- | ----- | ----- | ----- | ----- |
| ejson parse with Array | 1.1377ms | 6.6636ms | 72.2971ms | 772.8875ms |
| Superjson parse with Array | 2.1457ms | 10.4405ms | 74.6938ms | 777.4161ms |
| Superjson parse with Map | 0.8677ms | 8.8555ms | 74.2021ms | 768.9108ms |
| Array to Obj | 0.3515ms | 1.9336ms | 13.9112ms | 265.8113ms |
| Map to Obj (not needed) | 0.0011ms | 0.0003ms | 0.0004ms | 0.0003ms |
| Check Obj for changes | 0.2663ms | 2.3594ms | 38.8668ms | 393.3443ms |
| Superjson parse with Map | 0.3413ms | 2.3992ms | 33.205ms | 351.2783ms |


## Stringify string sizes
| Type | 1,000 | 10,000 | 100,000 | 1,000,000 |
| ----- | ----- | ----- | ----- | ----- |
| ejson | 29795 | 317795 | 3377795 | 35777795 |
| superjson arr | 29804 | 317804 | 3377804 | 35777804 |
| superjson map | 29804 | 317804 | 3377804 | 35777804 |


## Totals
| Type | 1,000 | 10,000 | 100,000 | 1,000,000 |
| ----- | ----- | ----- | ----- | ----- |
| Current with ejson | 6.43 | 32.28 | 312.25 | 3871.73 |
| Current with superjson | 11.74 | 51.49 | 437.85 | 5480.62 |
| Map with superjson | 3.35 | 35.46 | 347.25 | 4171.52 |
| Map to Arr with ejson | 10.70 | 47.80 | 374.15 | 4777.81 |
| Map to Obj with ejson | 17.70 | 33.35 | 457.11 | 3980.53 |

Julusian:
I'm curious if JSON.stringify would be any faster?

I am thinking that maybe we could do something weird like: 
```
ejson.stringify({ // the general ipc layer
  type: 'setVariables'
  values: JSON.stringify(.....) 
})
```
essentially, pre-stringify the values with some cheaper algorithm/lib that supports less types.  
I would would be open to taking this to an extreme of some custom serializer if that provides enough benefit 

But from these numbers, I do see 2 smaller gains we can make, `setVariableValues with Map` would be a safe and easy change to support. And I still think we should drop the `Array to Obj` step inside Companion, all we do with it is iterate over the array in one place so it may as well stay as an array. Presumably that will make `Check Obj for changes` a bit quicker too, due to less random lookups. Between them that would be 10% less time, which is a start.

Something else that might be worth looking into, is whether it would be faster to drop ejson entirely and instead enable `serialization: 'advanced'` for the child process, to let the ipc use structuredclone. I suspect that will cover enough types that we need. I think I wanted to avoid anything too nodejs specific, as I want to keep the option open to supporting modules in other languages, but all that will mean is retaining support for ejson/json later on

Julusian:
I was curious about my questions here, so have taken a very quick look into some synthetic tests.
It looks like doing a pre-stringify would be faster than letting ejson stringify the variables map.
I am seeing a change in my tests of going from 44ms to 17ms by doing that.

Additionally, I gave `serialization: 'advanced'` a quick try too, and passing an object based structure across has a cost of 11ms compared to 44ms with ejson. Or if I have the same data as a map, then it costs 17ms.

A big problem though is that the format of this advanced serialisation it not stable across nodejs versions. So it would only be possible to use for modules in the same nodejs version as companion itself. This may be too much effort to utilise.

But an easy win is to either stringify just this object with `JSON.strinigify` before it gets to ejson, or to reevaluate where ejson is used, perhaps in all/most cases it can be avoided?

Looking at the code I am not 100% sure if we even need ejson.
Of the types that ejson encodes:
- Date
- RegExp
- Nan/Inf/-Inf
- Binary (Uint8Array?)

I believe the only thing that we use of these is Uint8Array/Buffer, and in only a few places, so maybe we could drop ejson entirely and a lighter weight wrapper around JSON.stringify and JSON.parse that only handles UInt8Array?  
Or maybe we 'cheat' and require everything that used to be a Buffer/Uint8Array to be transferred as a base64 string (which is what ejson is doing anyway)

-----

My test object was:
```
const testData = {
	name: 'Example',
	created: new Date(),
	nested: { a: 1, b: [1, 2, 3], c: { deep: true } },
	buffer: Buffer.alloc(1024),
	map: new Map(),
	err: new Error('Example error'),
}

for (let i = 0; i < 100000; i++) {
	testData.map.set(i, i)
}
```

thedist:
I've been testing on and off the last couple months with this and have found similar that we may be best off just using JSON.stringify and removing the ejson dependency.

Even with the some of the changes I've been testing to encoding I'm still concerned about large sets of variables being a potentially bottleneck so most of my focus has been on modules themselves to provide more granular control over what variables will be used which is why I haven't focused as much time on changes for Companions side as I had originally planned.

Julusian:
I haven't tested it yet, but I am committing to at least attempting to make it work without ejson when module-base hits v2.0 https://github.com/bitfocus/companion-module-base/commit/44b19a8877f04677a4cde4bd218848c1d329b340.  
The only use facing api change here is changing `imageBuffer` from advanced feedbacks to always need to be a base64 string (according to typescipt, adding a fixup for that one case is easy enough)

I am hoping that this switch means that stringifying that one property hack wont be needed (long term).  
But short term, maybe we should do that for 4.0

thedist:
One idea I've been thinking about recently, and I'm not the most experienced with worker threads or child processes, but could handling some of the variable stuff in a worker thread/child process be something worth considering?

Some of the slowest parts of handling variables is that we transform from objects to arrays and back, but also do a lot of checking such as seeing if a variables new value is undefined so it can be removed from the cache, or checking for changes and only passing on the changes, etc... these things are CPU bound so could potentially benefit from being in a separate thread/process.

This would add extra overhead, as sending data between main process and worker/child and back will make the entire process take longer, but it wont lock the thread of the Companion Module instance which may help some of the issues people experience with pressing a button and it taking a couple seconds to fire the action.

Julusian:
I think that trying to mix in a worker_thread for something here will be a bit of a challenge to do with any performance benefit. As most of the cost here is json stringifying/parsing, it would need to avoid parsing the full json in the main process/thread.
As any complex structures cant be shared (only cloned when being sent across), it would probably mean moving a chunk of logic across to the thread (making calls to that be async), which will have some amount of consequence to other areas.

Another option to consider is that maybe we should be doing more on the module side and trusting it to behave sanely?
If the module side did the diffing, then it would only need to report the variables which actually changed, which would cut down the size of the data being sent.  
The diffing inside companion doesnt look like it will be that costly actually, other than just the number of operations needed for it, but maybe the loop could be simplified/flattened a little as a consequence.

The final thought I have, is the cost of checking every action/feedback/whatever to see if it was interested in the variables that just changed. Perhaps some work should be done to give this a better architecture that avoids needing to search everywhere

jswalden:
Thinking afresh here:

In the web context, the right way to transfer fairly-structured data would be by structured cloning, accessible by [`Window.structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone) or in arguments to `Window.postMessage` or similarly in worker threads.

Node.js worker threads support `MessagePort` to invoke this structured cloning algorithm.

And if you create processes opting into [advanced serialization](https://nodejs.org/api/child_process.html#advanced-serialization), you can get access to this structured cloning algorithm with `process.send!()` already in use -- I think bidirectionally.

Is there some reason why you couldn't send hashes of property changes using structured cloning, which likely efficiently [handles](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) everything you need for this?  It definitely handles properties with value `undefined`.

Julusian:
@jswalden I hadnt reaslised that MessagePorts could be passed between threads, so I suppose we could have a thread handling this per module. But I am still not keen on doing so as it will be quite a lot of complexity, and while it would reduce cpu pressure on the main thread it would increase overall cpu usage.

> And if you create processes opting into [advanced serialization](https://nodejs.org/api/child_process.html#advanced-serialization),

I said somewhere above that this isn't a viable option, as it is only safe to do between nodejs processes of the same version. And as we let modules choose one of a few versions, we would then only be able to use this for some modules and not others, making a mess for ourselves and inconsistent behaviour.

Also remember that we are going across processes here, not worker_threads. And I would still like to keep the option open to support other non-js modules in the future, so would prefer not to be too tied into nodejs specific ipc quirks

thedist:
Increasing overall CPU usage is the main reason why I only did a very basic look into using threads for this, as it can still be a performance bottleneck (particularly on lower spec machines, and things like RaspberryPi's). The main point of this draft PR was to see if improvements can be made that'd reduce that overhead, such as ways to do it without transforming the data so many times (such as turning arrays into objects, and back into arrays, etc...) or iterating over the variables on too many different places.

PR #3259, opened 2025-02-06
feat: replace socket.io with trpc

Julusian:
I was originally unsure on whether this was a good idea, but I now think it is.

The TLDR is this starts on replacing the socket.io we use between the webui and backend with trpc, see the list of benefits this brings below the separator.

This PR so far contains the emulator and a couple of minor bits, my intention is that the rest of the ui can be worked through and merged piece by piece. While in progress nothing should break, but the code will be a bit more confusing and there will be the overhead of each ui running 2 websockets (should have no real impact)

-------------------------
Original description:

This is a POC that I am experimenting with. Not sure yet if it will be completed, it might get abandoned.

This experiment is looking into replacing the socket.io transport we use between the webui and the backend with [tRPC](https://trpc.io/). This is a modern and strongly type safe library designed for the kinds of things we are doing.

For now it is setup with a websocket (like socket.io), but we could look into changing that to be using SSE if we want to later, but I'm not sure if that will provide any benefits for us.

Once complete, this would:
* allow us to remove some of the glue we had to add around socket.io to be able to use it in a RPC fashion
* uses a slightly more standard (ie not invented here) approach to subscriptions for the ui
* have better guarantees that received messages are what we expect due to the zod based parameter validation
* have a more traceable handlers of rpc methods, as they will be constructed in a tree fashion of routers and child routers. unlike today where a socket.io handler can be anywhere (vscode intellisense should be able to direct you to the implementation via usual hotkeys)
* make it easier to reliably implement security on this api. different auth levels are very explicit with the 'publicProcedure' vs 'protectedProcedure'. middleware can be added to each of these to perform the appropriate auth checks. That said, we could easily add similar wrappers for the socket.io listeners as we are already using an abstraction. Some investigation is needed to figure out how subscriptions will behave and need to be affected when permissions for the socket change.
* A lot of our subscriptions can only be consumed in one place at a time, and rely on the client knowing that and doing it correctly.
* In various places we have invented other ways of ensuring predictable behaviour when multiple places want to subscribe to the same data, this is messy and tedious.

I'm currently a little unsure on whether this will bring enough benefit to justify the effort to change.


Julusian:
I am leaning towards committing to doing this conversion.
I have played around with converting the emulator as a better example of how the code can look.

Another good thing I am finding with this is that it makes handling the error states easier, as that gets left up to the trpc/react-query libraries to figure out and simply report the status to us. They then give us a function to trigger a 'reset' if we want to restart a subscription, something we don't currently have a good way to handle.  
This should resolve some issues we have with the ui not 'syncing' sometimes (which is causing a bunch of errors in sentry, as it does complete but is hitting the timeouts we have defined)

dnmeid:
If you think it's a good idea, please go ahead. The only point I want to throw in is that tRPC is very bound to Typescript and we will have a hard time reworking if we should decide in the (far) future to use a different language for the backend.

Julusian:
> The only point I want to throw in is that tRPC is very bound to Typescript and we will have a hard time reworking if we should decide in the (far) future to use a different language for the backend.

As I see it socket.io is no less bound to typescript. While there is some support in other languages, the ones I have seen are not very maintained.  
The one thing that socket.io does have going for it in this regard is that it is mostly just a websocket connection, so would be easier to translate to a new mostly websocket lib. But I would argue that if we are changing backend language, we probably want to adopt a transport that is designed to work well in that ecosystem. if c# signalr is basically their alternative to socket.io, I dont know about other languages. maybe others would make sense to go asyncapi and have to implement the nice bits ourselves?

And a large chunk of this rework is switching to using tanstack-query for fetching (which could be done separately from trpc), which is what makes a lot of this code 'nicer'. So having that in place should make it easier to replace the transport next time (you can see elsewhere in the codebase where it is being used already for a few http fetches)

TLDR; I see this as being neutral in that regard

dnmeid:
> So having that in place should make it easier to replace the transport next time

Agree

PR #3317, opened 2025-02-23
feat: Add generic length function, closes #3308

### LINKED ISSUE #3308: Add function "array length" to variable expressions

### Describe the feature

To go with array split etc.

Unsure of the best name for the function (length may be confusing given we've already got strlen)?

### Usecases

From Slack:
> Jeremy
>   Today at 5:31 AM
> Hi all wondering if someone can assist me with counting the amount of "objects" in a JSON that I've pulled from a REST API endpoint and storing that value to a variable?

To go with all the other array based functions.

peternewman:
Should this be targeted at main or develop?

Happy to hear other name suggestions, I've mirrored strlen for now.

I did wonder should I add push/pop at the same time (and possibly others...)


dnmeid:
Here are my thoughts:
I think the name of  `strlen()` is dated back to the time when variables had always and ever been strings and was chosen to make that more clear. I would keep strlen for legacy reasons and add a new generic function `length()`. That function length should work with strings (number of unicode graphemes), arrays (number of elements), JSON (number of properties). It could also work for integer numbers by returning the length of the string representation. We can then extend this if we introduce e.g. enums in the future.

As it is a feature that I think will not need super heaviy testing it should be targeted at main.

Feel free to add more useful functions.

Please write a little less minimalistic description in the getting started. Especially examples are very useful and the description needs to be suitable for non programmers. If you say a function "lenght" gives the lenght, that is not super useful. You need to explain more basic, like "count of elements".

dnmeid:
@peternewman still there?

shared-lib/test/expressions-functions.test.ts
```diff
@@ -2,6 +2,30 @@ import { describe, it, expect } from 'vitest'
 import { ExpressionFunctions } from '../lib/Expression/ExpressionFunctions.js'
 
 describe('functions', () => {
+	describe('general', () => {
+		it('length', () => {
+			expect(ExpressionFunctions.length()).toBe(0)
+			expect(ExpressionFunctions.length('')).toBe(0)
+			expect(ExpressionFunctions.length('a')).toBe(1)
+			expect(ExpressionFunctions.length('abc')).toBe(3)
```
dnmeid, line 10: expect(ExpressionFunctions.length('ä')).toBe(1) // codepoint U+00E4, one grapheme
expect(ExpressionFunctions.length('̈a')).toBe(1) // codepoints U+0308 U+0061, one grapheme
expect(ExpressionFunctions.length('́̈a')).toBe(1)  // codepoints U+0301 U+0308 U+0061, one grapheme

shared-lib/test/expressions-functions.test.ts
```diff
@@ -2,6 +2,30 @@ import { describe, it, expect } from 'vitest'
...
+			expect(ExpressionFunctions.length(9)).toBe(1)
+			expect(ExpressionFunctions.length(99)).toBe(2)
+			expect(ExpressionFunctions.length(-123)).toBe(4)
+			expect(ExpressionFunctions.length(3.14)).toBe(4)
+			expect(ExpressionFunctions.length([])).toBe(0)
+			expect(ExpressionFunctions.length([9])).toBe(1)
+			expect(ExpressionFunctions.length([99])).toBe(1)
+			expect(ExpressionFunctions.length(['abc'])).toBe(1)
+			expect(ExpressionFunctions.length([9, 'a'])).toBe(2)
+			expect(ExpressionFunctions.length(['a', 'c'])).toBe(2)
```
dnmeid, line 27: expect(ExpressionFunctions.length([1,,3])).toBe(3)

shared-lib/lib/Expression/ExpressionFunctions.ts
```diff
@@ -3,6 +3,23 @@ import { JSONPath } from 'jsonpath-plus'
...
+		let len = 0
+		if (v === undefined || v === null) {
+			len = 0
+		} else if (Array.isArray(v)) {
+			len = v.length
+		} else if (typeof v === 'object') {
+			len = Object.keys(v).length
+		} else if (typeof v === 'number') {
+			len = (v + '').length
+		} else {
```
dnmeid, line 25: Not all uncatched types have a lenght property, e.g. BIGINT

shared-lib/test/expressions-functions.test.ts
peternewman: Seems this is pretty hard to do entirely properly, but do you think of these is "good enough"?
https://mathiasbynens.be/notes/javascript-unicode

shared-lib/test/expressions-functions.test.ts
peternewman: Added

shared-lib/lib/Expression/ExpressionFunctions.ts
peternewman: I've added BigInt. What others are you aware (RegExp maybe) of/what do you think I should do in the default case?

peternewman: Added RegExp too now...

shared-lib/test/expressions-functions.test.ts
dnmeid: Actually node.js has the Intl.segmenter() since version 16 but that is not working 100% with combining marks.
https://www.npmjs.com/package/unicode-segmenter should do the trick and is even faster than the built in Intl.segmenter()

shared-lib/lib/Expression/ExpressionFunctions.ts
dnmeid: Hm, I think string should have its own case and then it is not part of the unknown rest any more. The unknown rest could maybe return NaN.
From the JS perspective there are also iterators and generators that are interesting for length() but currently we don't use them in variables, so if you want to implement them it would be a bonus for possible future use. Also we don't use regex data type currently and I'm not sure about the value of knowing the length of a regex.

Julusian: yeah I agree, a string case would be good and the fallback should always return NaN.

Another note is that `typeof new Regexp()` is `object`, so this will need to be above the object check

peternewman: Okay I've added/sorted those others, useful or not and made it return NaN as the default otherwise.

I guess ideally we need yet another type of object if we wish to test the NaN case...

shared-lib/test/expressions-functions.test.ts
peternewman: I think I've got unicode-segmenter going now, but it doesn't return the right result for all cases for some reason (unless I've made a stupid error somewhere)?

peternewman:
> @peternewman What's the state of this PR? Currently I see some failing tests and conflicts.

I've managed to resolve the conflicts.

I'm not sure what to do about the failing test though. As I mentioned in https://github.com/bitfocus/companion/pull/3317/files#r1981604888 it seems like it's a bug in unicode-segmenter or I'm not using it right or something? Do you have any suggestions?

> v4 is around the corner and I'd be happy to have this PR included.

Yes, me too.

Julusian:
I've just had the thought, should this be considering unicode like this?  
I ask because the other functions aren't which could lead to confusing results.

for example:
image

so if you are doing something that combines `length` and `substr` (or any other string function), that is going to give incredibly confusing and 'broken' behaviour.  
A likely scenario, is the naive trim off the last character `substr(my_str, 0, length(my_str)-1)`. 

So I think that short term this should not consider multi character unicode.  
And then a follow up can be done (I think its too late in 4.0 for this) which looks at all the functions and makes then consider unicode correctly. This could be a deep rabbithole, or could be pretty simple to achieve.

dnmeid:
> I've just had the thought, should this be considering unicode like this?

Yes, it should. That is a valid grapheme.

> so if you are doing something that combines length and substr (or any other string function), that is going to give incredibly confusing and 'broken' behaviour.

`substr` is not unicode aware, so you should combine it with the correct counterpart `strlen`.
I think the correct counterpart for `length` is `slice` which we don't have yet.

dnmeid:
@peternewman I found the issue with the count of 2
It is not a bug and we can keep this as a correct test result
The order of the two codepoints is wrong
U+0308 U+0061 should give 2 and
U+0061 U+0308 should give 1.
They are all looking the same, what makes it even more confusing considering that a grapheme should be what you see, but that is the unicode definition and the difference gets obvious when you add a regular char at the front. The combining mark has to be after the base char.

image

So you just need to adjust the test.

Julusian:
Ah, I forgot there was `strlen` already :facepalm: 

Then my only request is that the docs should make it really clear which functions consider unicode correctly, and which ones are naive string versions

shared-lib/test/expressions-functions.test.ts
peternewman: Resolved via https://github.com/bitfocus/companion/pull/3317#issuecomment-2928216455 and @dnmeid 's spot.

peternewman:
> @peternewman I found the issue with the count of 2 It is not a bug and we can keep this as a correct test result The order of the two codepoints is wrong U+0308 U+0061 should give 2 and U+0061 U+0308 should give 1. They are all looking the same, what makes it even more confusing considering that a grapheme should be what you see, but that is the unicode definition and the difference gets obvious when you add a regular char at the front. The combining mark has to be after the base char.

Hmm, that feels a bit to me like the current rendering is broken (in the web-browser) and it should have shown it as `..a` or that unicode-segmenter is broken, which doesn't sound like it from what you've said.

> `substr` is not unicode aware, so you should combine it with the correct counterpart `strlen`. I think the correct counterpart for `length` is `slice` which we don't have yet.

This seems to imply that slice isn't fully unicode compatible:
https://stackoverflow.com/a/70303029

There was mention of a unicode substring type tool elsewhere.

I'm quite happy to implement slice if it actually works as we want with unicode.

> I've just had the thought, should this be considering unicode like this? I ask because the other functions aren't which could lead to confusing results.

> Then my only request is that the docs should make it really clear which functions consider unicode correctly, and which ones are naive string versions

I've attempted to improve on this. Let me know if we've got preferred terms for grapheme and byte that might be more user-friendly? I think I've caught all the other broken functions, but let me know if I've missed any.

Julusian:
> This seems to imply that slice isn't fully unicode compatible:

our substr function is using slice internally, so Im pretty sure it isnt. But we could expose a method called slice (or whatever) that uses some other implementation internally.

PR #3342, opened 2025-03-19
feat: add surface Mirabox Stream Dock 293V3 and N4

dnmeid:
This adds support for the surfaces Mirabox Stream Dock 293V3 and Stream Dock N4.
It closes https://github.com/bitfocus/companion-module-requests/issues/1494

IMG_20250319_222403~2

There is maybe still a problem but I wasn't able to track it so far.
Sometimes a button jpg doesn't draw. Investigation showed that when that happens the USB packet is missing one byte. Unfortunately I'm currently developping on Ubuntu on WSL. WSL doesn't feature HID support, so I have to use USB forwarding from the host to the Ubuntu usblib. The Companion log doesn't show malformed packet sizes, so I assume it is a problem of the USB forwarding.
I tried to make a Windows build from WSL Ubuntu but this also failed. It seems that the build script thinks it is running on Windows and tries to unzip using power-shell commands.
```
/usr/bin/bash: line 1: Expand-Archive: command not found
/home/dnmeid/companion/companion/node_modules/zx/build/core.cjs:277
    const from = getCallerLocation();
```
@Julusian any idea about the build?


companion/lib/Resources/Util.ts
```diff
@@ -295,9 +295,8 @@ export async function transformButtonImage(
 
 	// pad, in case a button is non-square
 	const dimensions = image.getCurrentDimensions()
-	const maxDimension = Math.max(dimensions.width, dimensions.height)
-	const xOffset = (targetWidth - maxDimension) / 2
-	const yOffset = (targetHeight - maxDimension) / 2
+	const xOffset = (targetWidth - dimensions.width) / 2
```
Julusian, line 334: I'm wondering how this behaves when combined with rotation? Could that be why I did a `Math.max` here?

companion/lib/Surface/USB/MiraboxStreamDock.ts
```diff
@@ -0,0 +1,1414 @@
...
+	config: Record<string, any> = {}
+
+	readonly #streamDock: StreamDock
+
+	readonly #writeQueue: ImageWriteQueue<string, [number, number, ImageResult]>
+
+	/**
+	 * Whether to cleanup the deck on quit
+	 */
+	#shouldCleanupOnQuit = true
```
Julusian: This doesnt appear to get changed anywhere

companion/lib/Surface/USB/MiraboxStreamDock.ts
```diff
@@ -0,0 +1,1414 @@
...
+		this.config = {
+			brightness: 100,
+			rotation: 0,
+		}
+
+		this.#streamDock = streamDock
+
+		this.logger.debug(`Adding Mirabox ${this.#streamDock.productName} device: ${devicePath}`)
+
+		const configFields: CompanionSurfaceConfigField[] = [...OffsetConfigFields, BrightnessConfigField, LegacyRotationConfigField, ...LockConfigFields]
```
Julusian: I think this should use `RotationConfigField` instead of `LegacyRotationConfigField`.
The legacy version exists to avoid breaking existing configs, and should probably be phased out at some point. I don't think it makes sense to use for new surface types

companion/lib/Surface/USB/MiraboxStreamDock.ts
```diff
@@ -0,0 +1,1414 @@
...
+							this.emit('down', action)
+						}
+					} else if (action.type === 'push') {
+						this.emit('push', action)
+					} else if (action.type === 'rotateLeft') {
+						this.emit('rotate', action, -1)
+					} else if (action.type === 'rotateRight') {
+						this.emit('rotate', action, 1)
+					} else if (action.type === 'swipeLeft') {
+						this.emit('rotate', action, 1)
```
Julusian: should this be -1?

Julusian (approved):
no major complaints from me, just some minor comments

companion/lib/Resources/Util.ts
dnmeid: Good point, will check that.

Julusian:
> @Julusian any idea about the build?

I don't think I've tried to build in WSL in a long time because I typically need usb to work whenever I test on windows, so I have to have my machine setup to build&run outside of WSL.
That said, I'm surprised it thinks it should invoke powershell commands. Doesnt that mean it is running a windows version of node?

dnmeid:
> That said, I'm surprised it thinks it should invoke powershell commands. Doesnt that mean it is running a windows version of node?

No, not I'm aware of. I run `yarn windist` on WSL Ubuntu. It builds the tools and webui. Then downloads the node runtime and then tries to unzip it. The downloaded node runtime of course is for Windows. (`Expand-Archive /home/dnmeid/companion/companion/.cache/node/node-v22.14.0-win-x64.zip -DestinationPath /home/dnmeid/companion/companion/.cache/node-runtime/tmp-22.14.0`)

dnmeid:
> I don't think I've tried to build in WSL in a long time because I typically need usb to work whenever I test on windows

BTW since last November there is a WSL update available that makes HID available. There seem to be still some issues, that is the reason why I want to test with a native build.

Julusian:
> No, not I'm aware of. I run yarn windist on WSL Ubuntu. It builds the tools and webui. Then downloads the node runtime and then tries to unzip it. The downloaded node runtime of course is for Windows. (Expand-Archive /home/dnmeid/companion/companion/.cache/node/node-v22.14.0-win-x64.zip -DestinationPath /home/dnmeid/companion/companion/.cache/node-runtime/tmp-22.14.0)

Oh right yeah, the scripts don't really support for building for a different OS currently. https://github.com/bitfocus/companion/commit/79d257d871083dfed638d6ed32e3a107fe88de44 should resolve that immediate issue, no idea if it will hit other issues though

companion/lib/Resources/Util.ts
dnmeid: Nope, actually I have non square buttons at the strip of the N4 and rotation is only ok with this change.

alarm10101010:
Are the Mirabox devices supported in the recent szable Version?

I ask because I did not find any udev rules for them in the rules file.

alarm10101010:
> No, this is too new to be in the 3.5 releases. You will need to use the 4.0 betas for now

Die not find udef rules for mirabox in the rules files in 4.0 and 4.99 alpha package.

Maybe it was forgotten to add to the rules file?

Julusian:
yeah it looks like it was forgotten, I added it a few hours ago

PR #3401, opened 2025-04-27
feat: ui for grouping connections

Julusian:
Groups can be added, renamed and deleted.
Deleting a group moves every connection back to ungrouped.

drag and drop works as before, now being able to move between groups too. the group headers are drop zones too, and will add to the end of the group, making it easier to push things into groups without having everything expanded.

a heirarchy of groups can be formed, with everything draggable to make whatever structure is desired.

image

the ungrouped heading disappears when there is nothing in the ungrouped section.
image

Empty groups get a visible dropzone (when expanded). for the ungrouped, this will make the heading reappear too
image

when filtering by status, each group shows the hidden count (when expanded)

image

Note: import/export support for this is not currently implemented


phillipivan:
@Julusian Do you plan to / can you implement group scoped module enable/disable switches (and internal actions to do the same)?

Julusian:
> @Julusian Do you plan to / can you implement group scoped module enable/disable switches

I hadn't thought of doing that, but it is done now

phillipivan:
@Julusian a follow up q: Can these be nestable? Working in an environment where our companion control is facility wide, this would be quite useful.
Ie:

> - Venue 1
> | - Audio System
> | | - Radio Mics
> | | - Amps
> | | - Processors
> | - Broadcast System
> | | - PTZs
> | | - Graphics
> | | - Displays
> | | - Switching
> - Venue 2
> - Venue 3

Julusian:
I haven't thought about making them nestable, but I suppose that wouldn't be hard.

Getting the group enabled state to be its own thing is more complex than I was hoping, due to the async nature of the loop which runs and maintains the connections running. It allows for checking if a connection is running, but not to understand if the connection is starting or stopping. So figuring out what connections to start/stop when toggling a group is not simple..
I have a couple of ideas on how to solve this, but one feels wrong, and the other is a large refactor of the connection running logic

Julusian:
I'm worried about the complexity of the group enable/disable toggle and risk of introducing bad bugs (too risky for 4.0), so I've pulled that part out for now, that can be a follow up PR (for 4.1).

phillipivan:
Thanks for confirming. Will the nesting make it into 4.0?

Julusian:
@phillipivan https://github.com/bitfocus/companion/pull/3499
I'm glad I came back to that later, I took a vastly different approach that became much simpler

PR #3412, opened 2025-05-03
feat: support new upgrade execution for modules

Julusian:
Implements https://github.com/bitfocus/companion-module-base/pull/118

This has turned out to be a more complex state machine than I initially expected, so I am wondering if this should be held back for 4.1 (it could still be reverted in the module lib, so that the other changes can be released)

This api change is done so that how the options are stored can be different from how they are used, without needing modules to be the ones doing the parsing. 

This is a precursor to #2345, which I expect will build on this by further by changing the structure of the options object that companion and upgrade scripts use, but not for action/feedback execution or subscribe/unsubscribe calls.
At this stage, that means we can pre-parse variables in textinput fields (only those with the `useVariables` flag set for now) before passing the objects off to the modules. 

This should be non-breaking for modules, and only those using the updated base lib will hit the new flows, so any bugs found will have limited impact.

TODO
- [ ] try it with a module
- [ ] thoroughly test it with a module
- [ ] write some unit tests?

### DIFF
--- a/companion/lib/Controls/ActionRecorder.ts
+++ b/companion/lib/Controls/ActionRecorder.ts
@@ -346,6 +346,7 @@ export class ActionRecorder extends EventEmitter<ActionRecorderEvents> {
 					options: options,
 
 					uniquenessId,
+					upgradeIndex: undefined,
 				}
 				const delayAction: RecordActionEntityModel = {
 					type: EntityModelType.Action,
@@ -356,6 +357,7 @@ export class ActionRecorder extends EventEmitter<ActionRecorderEvents> {
 						time: delay,
 					},
 					uniquenessId: undefined,
+					upgradeIndex: undefined,
 				}
 
 				// Replace existing action with matching uniquenessId, or push to end of the list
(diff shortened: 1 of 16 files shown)


phillipivan:
@Julusian if you are doing preparsing of variables, do you plan to extend this to dropdowns with `allowCustom` enabled? Or perhaps turn it into `allowCustom: { variables: true },`. I have quite a few modules that can parse variables from a custom dropdown entry simply because it cuts down on the busy work of programming actions (with extra option definitions, `isVisible` statements, and then callback code to determine which option to use).

Julusian:
@phillipivan 2345 will add a way to make every field be an expression, which will allow for that. But that is a later step.

In this, any text fields with 'useVariables' set will be auto-parsed for you, but as we don't have the same flag on dropdowns it wont be auto-parsing those. I am leaning towards not making this work, purely because respecting the behaviour of that in the ui will be a nightmare.

so long term, this will be possible for you to cleanup (by using an upgrade script to convert existing values into expressions), but short term wont be addressed

PR #3441, opened 2025-05-30
add type handling to variables, closes #3335

### LINKED ISSUE #3335: Differentiation of custom variables into simple and objects/arrays in variables tab. Switch editing custom variable mode text/expression

### Describe the feature

Maybe there is an option to add a type switch similar to switching the button text modes between `text` and `expression` in custom variable editor. It would be extremely convenient.

### Usecases

In some cases, the ability to fix the type of a variable by the variable editor (the variables tab) is very much lacking. For example, I know for sure that my custom variable will be of the `object` type, and now, if I make any changes in this tab, the variable automatically becomes a `string` type and all the logic on pages breaks. But if I could switch edit mode of variable and get access to fields of my object instead of `[object Object]`

dnmeid:
This PR adds two things:

1. A type annotation to the readout of variable values. Users can now easily spot if this is number 42 or the string '42'.
When the variable is undefined, it is not omitted any more but the annotation will show 'N/A'  

image

2. The possibility to enter values in custom variables as JSON.  
A new button is added similar to the mode switch of the button text entry.  When the entry field doesn't have the focus, the button works similar to the type annotation and will show if the current variable value is a string or something else. Unlike the type annotation it switches only between string and object.  
In string mode the value is shown like it was before.  
In object mode, a stringified version of the value is shown.
The button and the field are updated live until you they get the focus (button and the field are one focus group). When they have focus you can change the mode or enter a new value. This resembles the current behaviour. Only difference is, when in object mode the variable will only be updated when the entry can be parsed.  
When switching from text to JSON mode, the current text is re-stringified. When switching from JSON to text mode and the value is a text, nothing happens. When the value is something else the raw entry will be interpreted as a string.  
When in JSON mode the text entry also parses JSON5 (properties don't need quotation, single ticks are allowed, spaces are allowed...) but the stringify part will always stringify to regular JSON. So it is e.g. possible to copy/paste JSON from a JS file.

image

closes #3335

todo:
- [x] update documentation
- [ ] the type CompanionVariableValues should be adjusted to reflect the actual possibilities


webui/src/Components/VariablesTable.tsx
```diff
@@ -176,6 +177,31 @@ const VariablesTableRow = observer(function VariablesTableRow({
 		}
 	})
 
+	let typeDescription = 'unknown'
```
Julusian: There is some duplication between here and CustomVariablesList that would be nice to avoid if possible.
This big elseif in particular, as it will be too easy to edit one and forget the other. Some of the jsx which renders the values too could hopefully be moved into a new reusable component, not 100% sure if that will be nice to do (I cant tell if they are the same or slightly different)

dnmeid:
> There is some duplication between here and CustomVariablesList that would be nice to avoid if possible.
This big elseif in particular, as it will be too easy to edit one and forget the other. Some of the jsx which renders the values too could hopefully be moved into a new reusable component, not 100% sure if that will be nice to do (I cant tell if they are the same or slightly different)

Is done now, but was a lot more effort than I hoped. In deed both were slightly different. The custom value display was just showing the value but the module variables hat the truncation option that could be collapsed or uncollapsed.

I now made both the same and you can collapse or uncollapse them.
The variables in the custom table collapse to their own state. The variables in the module table use the collapseHelper, what currently has no benefit because there is no collaps all control there.

Also I made the Mode/Less control a button instead of a link, and it sits now inside of the box for saving a little space.

image

PR #3444, opened 2025-05-31
feat(graphics-overhaul): add layer usage property

Julusian:
There was some discussion on how to 'downlevel' the config for use by the simple protocols or controls which want specific values in https://github.com/bitfocus/companion/pull/3299

This is my proposal on how this should operate.  

As agreed, there is automatic matching of the layer to use for each value. This currently defaults to the lowest layer of the appropriate type, but a manual selection of a 'usage' will override this.

image

I have opted for a rigid structure of usage values, as this means that the user doesn't need to guess or lookup elsewhere what the values should be. As the usage of these properties is purely internal at this stage I think this makes sense. This can easily be changed to a free-text field later if needed.

Internally, then everywhere we were using the old style object for the various protocols or surfaces, now uses an object `ImageResultProcessedStyle` which is built from matching these layers (or for the old style buttons, from the simple style object). 

With a simple implementation of converting that `ImageResultProcessedStyle` back into the old drawing style, the internal 'use another buttons style' feedback works pretty well and shows an approximation of what the apis will have access to:

image
In this:
- 1/0/3 is an old style button
- 1/0/4 is a new style button
- 1/1/3 is old style and using 'use another buttons style' of 1/0/3
- 1/1/4 is old style and using 'use another buttons style' of 1/0/4

I did this purely as an easy way to visualise the simplification process, this feedback implementation may not make it into a release.

-----

With this change, existing surfaces and satellite will handle these new buttons correctly.
The protocols which can modify style properties have not been updated yet, but I think will be fairly easy to update to modify buttons in the same properties

### DIFF
--- a/companion/lib/Cloud/Controller.ts
+++ b/companion/lib/Cloud/Controller.ts
@@ -704,7 +704,7 @@ export class CloudController {
 	 */
 	#updateBank(location: ControlLocation, render: ImageResult): void {
 		const bank = xyToOldBankIndex(location.column, location.row)
-		if (typeof render.style === 'object' && !render.style.cloud && bank) {
+		if (render.style.state && !render.style.state?.cloud && bank) {
 			const updateId = v4()
 			for (let region in this.#regionInstances) {
 				if (!!this.#regionInstances[region]?.socketTransmit) {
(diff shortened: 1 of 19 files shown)


Julusian:
@dnmeid what do you think of this?

I know that long term you want this processing/parsing to move into the surface/api 'modules', but until we have that structure, I think it makes sense to do it like this. When a surface needs a 'guage' value, we can easily extend this structure to include that.  
Whether we keep it rigid or make it looser long term is a discussion to have later, I don't expect us to get around to doing anything on this for a while.

PR #3445, opened 2025-05-31
feat: Add length function, closes #3308

### LINKED ISSUE #3308: Add function "array length" to variable expressions

peternewman:
Supersedes #3317


peternewman:
It seems I've now managed to resolve the original conflicts...

PR #3482, opened 2025-06-15
 feat(graphics-overhaul): support 'old feedbacks'

Julusian:
See a full explanation of the plan and reasoning in https://github.com/bitfocus/companion/discussions/3293#discussioncomment-13333908

This is my pitch on how to handle 'old feedbacks' (primarily advanced feedbacks) that don't fit into the new element based flow.  
I see this as a medium term solution, until we can drop support for versions of the module-api that support advanced feedbacks.
The hope here is to make it so that existing configs could be auto-converted across without affecting how buttons are drawn (for the simple square button case). This theory is not implemented, and needs testing to verify if it will work well enough.

image

TODO:
- [ ] Hook up to ui preview drawing

### DIFF
--- a/companion/lib/Controls/ControlTypes/Button/Base.ts
+++ b/companion/lib/Controls/ControlTypes/Button/Base.ts
@@ -8,6 +8,8 @@ import { EntityModelType } from '@companion-app/shared/Model/EntityModel.js'
 import { ActionSetId } from '@companion-app/shared/Model/ActionModel.js'
 import { DrawStyleButtonStateProps } from '@companion-app/shared/Model/StyleModel.js'
 import debounceFn from 'debounce-fn'
+import { CompanionVariableValues } from '@companion-module/base'
+import type { ParseVariablesResult } from '../../../Variables/Util.js'
 
 /**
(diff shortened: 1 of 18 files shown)


krocheck:
Forgive me ... I'm catching on the graphics overhul items.

We have no migration path for the existing feedbacks?  No way to in-place upgrade them?

Julusian:
For boolean feedbacks I think we will be able to manage migrating some or maybe all scenarios.  

But the challenge lies in that the new and old drawing properties look nothing alike, and the new flow doesn't allow for overriding in the same way. A core principal of the new flow is that instead of overriding properties with feedbacks, feedbacks provide local variables which can be used in expressions to compute or use logic to determine values.  

And because of this shift in how the drawing composition works, the override nature of advanced feedbacks simply doesnt translate.

I am a little worried that this might make simple buttons overly complex to create, so far I haven't had as much input on ux as I would like.

Julusian:
I have an alternate proposal #3620, which pushes the overhaul back into more familiar territory with overrides. It is a bit more manual than before, but I see that as the cost of the extra functionality we have in the drawing.

PR #3487, opened 2025-06-16
feat(graphics-overhaul): support old imagebuffers

Julusian:
This is a draft idea of how to handle imageBuffers in the new drawing scheme

The idea is based around deprecating imageBuffers, and pushing modules towards producing base64 pngs of unknown dimensions if they feel they have to produce an image. Ideally these images will be for things like previews, and that everything else they might want to do should be handled by our drawing system  
In doing this, it means we don't need a fancy flow for handling these buffers, just something good enough to replicate old behaviour. 

So this is simple handling of the buffers, during the parsing phase it composites all the buffers into a single png64. It does so assuming the button is square (like all drawing used to be), but factoring in the topbar (like it used to).
Hopefully this means that existing setups will draw almost pixel perfect to how it drew before.

As a consequence, this requires another usage type to figure out what element to put the composition into, which shares the image element type. (This is not fully implemented yet)

An example button, where the triangle is in the corner of the buffer in a button drawn for SDS:
image

This needs some analysis of existing modules to see how widespread the impact would be, and to see how big of a problem this will cause.

However, a key thing to note is that modules will be able to continue generating images by simply outputting this as a png instead. The downside being that it will be up to the module to figure out what dimensions to draw at, (or the user if they make the user specify that), and we won't guarantee drawing without up/downscaling.

-----

My only idea on how to handle them 'properly' is that we could create 'virtual' feedbacks to make the module give us the buffers at every dimension we need to draw. Then when drawing, we substitute in the correct bitmap.  
This leaves me with concerns about the cost of doing so (do we always request all sizes we might need?) and complexity due to the surface-type specific drawing.  In the ui, you can choose the aspect ratio to draw at, when changing that there will be a delay while the module prepares a new buffer (and this will require a virtual feedback to be created to get this buffer).
And now that we have a flexible drawing system, adding new elements to replicate what modules have been doing with buffers should be quick and easy to do.

## Summary by CodeRabbit

* **New Features**
  * Added support for deprecated "Image Buffers" property in button styles
  * Image buffers now stack and concatenate rather than replace
  * Enhanced element picker with schema-based validation

* **Bug Fixes**
  * Fixed pixel buffer color channel conversion to prevent data corruption

### DIFF
--- a/companion/lib/Graphics/Renderer.ts
+++ b/companion/lib/Graphics/Renderer.ts
@@ -574,7 +574,7 @@ export class GraphicsRenderer {
 	}
 
 	/**
-	 * Draw the image for a btuton
+	 * Flatten an array of imagebuffers into a single base64 image
 	 */
 	static async drawImageBuffers(showTopBar: boolean, imageBuffers: DrawImageBuffer[]): Promise<string> {
 		return GraphicsRenderer.#getCachedImage(


Julusian:
Modules which make reference to imageBuffer (as of 9 June 2025):

|  module | usage |
|-----|-----|
| anomes-millumin | progressbar |
| behringer-xair | meters |
| bmd-atem | media pool previews |
| cedar-dns8d | complex meters/status |
| companion-satellite | button bitmaps |
| cvmeventi-countdown | progress bar |
| discord-api | mic/headset mute status |
| ereca-stageracer | status circle. meters |
| focusrite-mp8r | meters |
| getontime-ontime | progressbar |
| kiloview-ndi | commented code |
| metus-player | media preview |
| newblue-captivate | unsure if this does anything |
| resolume-arena | meters? |
| rode-rcv | meters. visual status thing |
| sennheiser-digital6000 | complex status image |
| sennheiser-ewdx | complex status image |
| shure-scm820 | complex status image |
| stagetec-rcp | meters |
| studiocoast-vmix | meters. corner indicators |
| vbaudio-voicemeeter | meters |
| yamaha-rcp | meters |
| zinc-oscpoint | progressbar |

krocheck:
`shure-wireless` would not easily be able to work in PNGs as its programmed.  It uses a stripped down version of the old `image` class to create the canvas then draw various elements in from ARGB buffers.  It saves every combination of the rasterized "layers" as a buffer so that it can quick recall and send it back instead of building on-the-fly for every change.

I can set it up to draw for multiple rasters, but I'll have to establish XY coordinates for all the elements for each of those.

I did notice that scaling would be nice because I have these set up for 72x72 or 72x58 and they look horrible in web buttons and emulators, so being able to adapt it would be great.  I could probably find a library (or you know of one) to do the layer rasterization as PNG and send that back.  I'd just would like a recommendation so I can efficiently re-implement instead of spending gobs of time experimenting.

Julusian:
>  I could probably find a library (or you know of one) to do the layer rasterization as PNG and send that back. I'd just would like a recommendation so I can efficiently re-implement instead of spending gobs of time experimenting.

`@julusian/image-rs` has been reworked to support handling some image formats, so is able of taking in pixel buffers, doing some basic manipulation on them and outputting as a png or base64 png. (intended to solve problems similar to this)

Julusian:
Another thought on this; maybe we should allow modules to produce some 'presets' for the drawing elements.
As in, they expose a custom element type with a list of fields they specify, which gets substituted out for a list of elements that the module chooses, with the values from the fields they gave exposed as some local variables.

With that, I expect that modules would be able to replace almost all of their custom drawing with our new system, assuming we provide a good set of primitive elements for them to use.

Julusian:
Having now reimplemented this on top of the other feedback reworking, I am not entirely happy with it. I feel it makes quite a mess, causing _every_ preset (and existing button) with an advanced feedback to gain an extra image layer, with all feedbacks pointing to it, just in case they produce image buffers.

And internally, this is a few hacks to handle the stacking and fixing up the types. 

But with the idea I have on modules defining composite elements, I see this as a short-term solution. And hopefully we can start pushing modules away from advanced feedbacks, which will further mitigate this.

Issue #3491, opened 2025-06-18
Shuttle ring should trigger as repeated button presses (Contour Shuttle)


arikorn:
### Is this a feature relevant to companion itself, and not a specific module?

- [x] I believe this to be a feature for companion, and is not specific to a module

### Is there an existing issue for this?

- [x] I have searched for similiar existing issues

### Describe the feature

_Note: this is written for the Contour Shuttle since it's what I have. In theory it applies to X-keys shuttles if they work this way too._

The Contour Shuttle has a spring-loaded ring that serves as the "shuttle." Currently turning the shuttle emits a rotational action and sets an internal "shuttle" variable to indicate its position, emitting exactly one rotational action per "click" (the ring has seven "clicks" in each direction, intended to represent increasing speed). "Direction" is determined just like a knob, so moving from 1 to 2 emits "rotate-right" and returning from 2 to 1 emits "rotate left"

This behavior is somewhat at odds with the usage of a shuttle in video editing, in which a variable-rate "fast-forward" (or reverse) is accomplished by holding the ring in position and perhaps moving it slightly closer or further to zero to adjust the speed. In this method of operation, the direction of rotation is irrelevant: only the current position matters -- the action continues at the indicated speed dependent only on the current position of the dial, even if the dial is held still.

One (common) way to implement this would be repeated button presses with the following logic:

1. As the ring is rotated, update the `shuttle` variable (the speed)
2. If the shuttle ring moves off of zero, send repeated "keypresses." Either:
    2a. The shuttle variable determines the rate of keypress and each keypress uses a fixed increment or...
    2b. The keypress interval is fixed, and the `shuttle` variable determines the increment size
3. If the shuttle position is 0 release the button.

Notes:
1. When trying to implement this I found that 2b is currently a simpler option since Companion has no "sign()" function, which would be needed to implement 2a as a fixed, but signed, increment. (sign = abs(x)/x, which is a "mouthful" when "x" is "$(internal:shuttle)").
2. It is possible to program my desired behavior using actions and the new if..then.else logic-action: basically the rotation actions have to trigger the button-press action and the press action has to determine whether to (a) act and repeat after a wait or (b) trigger the button-release. Coding it this way is fairly "bulky" -- i.e. adds much unnecessary code and obscures the actual desired action. So a built-in feature would be more desirable.
3. If someone has done what I describe in note 2, then this proposal might break their button. One option to mitigate this problem would be to only emit a button-press if no rotation action was defined.
[...]

Issue #184, opened 2018-07-21
Request: Repeat button press when held


Andygoes:
**Describe the feature**
When pressing and holding a button to advance _n_ frames, give the option in Companion to repeat the command every _n_ ms 

**Is this platform dependent (windows, mac, ..)?**
no

**If documentation is required to implement, do you know where to find it?**
n/a

**Usecases**
The most obvious use case I can think of is incrementing/decrementing values - scrubbing media, increase/decrease volume, fade up/down.

Danie10:
Also would be used extensively in gaming for say moving forward, backwards, etc. Need to repeat while held and stop when released.

I've actually used xdotool on Linux with keydown (press) and keyup (after a long press duration) to toggle the repeat but it is not ideal as you need a longpress to stop, and just releasing with serve this use case request as well.

istnv:
@Danie10 
There are many ways to accomplish this in Companion.
Here's one for a fader up:
![image](https://github.com/user-attachments/assets/05887fe7-abd2-4b80-8b00-a5c49e1c061f)

AJolly:
> [@Danie10](https://github.com/Danie10) There are many ways to accomplish this in Companion. Here's one for a fader up: ![image](https://github.com/user-attachments/assets/05887fe7-abd2-4b80-8b00-a5c49e1c061f)
Sadly this doesnt seem to work for me.  It only increases once

![Image](https://github.com/user-attachments/assets/990b477c-33dc-4f76-ac77-334911cf5272)

Danie10:
I'm using the button action `internal: System: Run shell path (local)` with `xdotool keydown Right` for advancing frames on repeat quite well at the moment (in video editing).  I use it similarly for advancing in 1 second increments using a different key. This works for holding the button down until I release it.

Only problem is it does not have a various ms delay you can specify. It basically repeats like you are holing the key down on the keyboard (can't adjust it for a slower repeat rate. So I use the one button for the frames key, and use a different button to use the key for 1 second intervals (in DaVinci Resolve).

arikorn:
This very old issue may be worth revisiting.

_On the one hand_: new internal actions such as groups and if-then logic have laid the groundwork for adding a "repeat" action (i.e. an action that contains other actions)

_On the other hand_: there's currently a very natural way to implement repeat-while-pressed:

- Set up your button's _Press Actions_ as usual
- Add a _Duration Group_
- Click the "pencil" to edit its duration and enable "Execute while held" 
[...]

PR #3492, opened 2025-06-18
Contour Shuttle (Surface): Add repeating button-presses in response to the shuttle ring

arikorn:
Add repeating button-presses in response to the shuttle ring, to enable "shuttling" (i.e. fast-forward/reverse type of actions).

Resolves issue #3491 Please see the issue for more details.

### DIFF
--- a/docs/7_surfaces/contour_shuttle.md
+++ b/docs/7_surfaces/contour_shuttle.md
@@ -1,6 +1,6 @@
 It is possible to use the Contour ShuttleXpress and ShuttlePro v1 and v2 with Companion (since v3.1.0; note: support for very old versions of ShuttlePro v1 was added in v4.1.3)
 
-You must enable it in Companion's Settings > Surfaces section and rescan for USB devices. (You do not need to restart for the setting to "take"... since at least v4.0)
+You must enable it in Companion's Settings > Surfaces section and rescan for USB devices.
 
 The button layout closely matches the device.
 


Julusian:
I'm not keen on adding click events to this as well as the rotate events. That feels weird to me..

I'm not sure what I would prefer though.
I don't entirely like how this surface emits any rotate events, I would prefer it to do just the variable. Replicating this pr could be done with a pair of triggers off that variable, except that the current interval trigger event only runs to whole seconds.

arikorn:
@Julusian , would you accept the PR if I removed rotation events on shuttle altogether?  I certainly don't need or want them -- I just left it in for compatibility.  (Personally, requiring someone to add triggers, or do as I outlined in the issue, seems like a less desirable solution than this PR.)

FWIW, the manufacturer's "control panel" app does allow defining actions for "Transition from Left 3 to 2" and so on, as well as for "Shuttle in Left 2"  which could be considered another justification for allowing both rotational and press actions (i.e. the manufacture themselves supports it).  But again, my preference is for repeated presses, which seems the most natural way to define things.  In fact, the manufacturer's control panel uses repeated button-presses as its main way of handling the shuttle.

arikorn:
added commit: Modified code to allow 'click' event to set the "force" argument in the `controls.pressControl()` method

arikorn:
In thinking about "appropriate use" of the shuttle, I "surveyed" Countour Design's prepackaged configurations for the shuttle ring. Of 140 profiles specified for the Contour Shuttle (v1):

 - 6% react to transitions (i.e. differentiate between shuttle rotate-right and rotate-left)
 - 19% specify a single keypress for each position, i.e. don't care about rotation direction and don't need repeats
 - 75% specify repeated keypress or mouse actions for each position, i.e. don't care about rotation but do repeat at variable rates

(Examples of single-keypress include software in which each keypress increases the playback speed or where there's a different F-key for each speed.)

Now to be fair, these are all for software control on the local computer and the actions are limited to keypress or mouse action, so for example, my main use-case -- to change the value of a parameter directly and have the magnitude of change vary by shuttle position -- is not even considered. I concede, however, that my situation can be handled by variable-rate repeated actions.

**My take-aways**:
1. Since 75% of the situations used variable-rate repeated actions, that seems large enough to justify adding repeated actions as suggested in this PR.
2. Nevertheless, 25% is substantial enough that both options should be made available to the end-user. 

One simple way to allow both options is to provide non-repeating rotational actions and a repeating "press" action, as done in this PR. The user can then choose the appropriate behavior by whether they put their actions in rotational vs. press scripts.

**If you agree I can update this PR to**:
 - A. provide variable-rate repeated press actions. (instead of fixed-rate)
 - B. Add a "shuttle-sign" variable, to make it easier for button-press actions to select the direction. (see my notes on issue #3491)
 - C. (optional) only allow repeated press actions if a rotate action was not specified.

**Alternatives**:  
- accept PR as is, recognizing that devices controlled by Companion are more likely to be able to handle variable step-sizes (and may have trouble keeping up with too many network requests...)
- add a config option to enable/disable repeated actions on shuttle rotate-left/rotate-right actions. (A downside of this is that 94% of the use-cases in my survey don't care about rotational direction).
- add an `internal:repeat` action (as in issue #184)? (Again with downside about rotational direction.)
- do nothing (the current challenge is not insurmountable, merely clumsy/difficult to maintain)
- other ideas?...

arikorn:
@Julusian I am hoping we can resolve this PR before it becomes too stale… I believe that you are on the fence about this PR rather than completely opposed, so perhaps we can find common ground?
To summarize my opinion: 
1. Having the shuttle actions associated with a “button” (as it is now) is essential (a) for clarity, to keep it with the other buttons on the surface and (b) for functionality, to easily make the shuttle context-sensitive by putting the surface in a surface-group. For example, when the related surface -- such as a Stream Deck -- is on an audio page the shuttle/jog can adjust volume, whereas when it’s on a PiP page, it can adjust position. (This is something I actually do.)
2. The shuttle should allow repeated-press actions. If you consider the design of a spring-loaded shuttle-ring, it is a hybrid of a button and a knob: returning to “off” when released, but rotating as well. The conventional way to handle this is by allowing both repeated button-press actions and rotary actions, as can be seen in the manufacturer's own drivers, below. (If I had to choose, I would keep repeated-actions since as shown in the statistics above, it represents the majority of the use-cases.) 

**I would greatly appreciate your response on how we might move forward with this PR**!  For now, I will push an update that implements what I proposed in the previous comment, i.e. variable-rate repeated actions. (Note that while the update works, it will need additional work to merge with the current state of main.)

image
image

arikorn:
Recording a Slack dialog here, for continuity and posterity :grinning: ...
@arikorn: 
 I had another thought about the Shuttle: yet another reasonable alternative would be to use repeating rotary actions rather than button presses, but with a difference: rotate left would fire as long as the position is negative, and rotate right for positive, regardless of the direction of rotation -- i.e. regardless of whether increasing or decreasing. (And firing rate would still be proportional to magnitude.)

@Julusian: 
>use repeating rotary actions rather than button presses, but with a difference: rotate left would fire as long as the position is >negative, and rotate right for positive

yeah, I am on board with this idea.
I imagine that will give the repeating actions that you need in a way that feels helpful and isnt fighting the press vs rotation
I am not sure if you need something to happen when released and returning to 0? If so I don't know how that should fit into this. Maybe that should keep similar to the current pr where it triggers a press when starting the rotation events, and releases when returning to 0?
I do feel like this might want to be done at a new location rather than the current `const xy = this.modelInfo.shuttle`, so that it is possible to choose how you want it to behave
and one other comment from the pr is that you are only calling `clearInterval` when the shuttle moves, but not if the surface gets disconnected; which might cause the rotation to continue forever (or at least would cause them to be fired internally, even if it does get rejected elsewhere)

@arikorn asks clarification about location and @Julusian replies:
>Not sure what you mean by this. Doesn't it have to go in the contourShuttle.on('shuttle', (shuttle) callback?

I mean that it should be choosing its location from something like `const xy = this.modelInfo.shuttleRepeating` instead of the current  `const xy = this.modelInfo.shuttle`
That way, behaviour for existing users will not change, and it lets you choose whether you want the events of just when it changes, or the repeating behaviour

@arikorn:
The second button will have to be one over, though (2/4, I think) since the immediately adjacent one (2/3) is already taken by a Pro v2 button.  Another option that just occurred to me, would be to add an internal action "Enable Repeat" (could even use it for repeat-while-pressed on regular buttons... as a possibly simpler version than the while loop?). Just a half-baked thought at this point, though.

@Julusian:
that is annoying, but not the end of the world. I am kinda tempted to move that button, that will be pretty easy for users to fixup in their config
... and later ...
I'm not immediately convinced by a 'repeat' mode builtin to buttons, so I think for now I would prefer to avoid doing that, but that could of course change later

arikorn:
OK. that should do it, hopefully, for this PR.

The last commit is a "bonus" fix, renaming rotational `direction` to `rightward` in the core code for clarity (see the commit message for full justification). If you don't like it, it's easy enough to roll back!

(Also, the crazy line count is because of the replaced YAML config file...)

arikorn:
OK, the crazy "added lines" count bugged me, so I uploaded the template in all three formats -- each in its own commit. Choose whichever you prefer or stick with the compressed one in the final commit above.

companion/lib/Surface/USB/ContourShuttle.ts
```diff
@@ -34,8 +35,10 @@ const contourShuttleXpressInfo: ShuttleModelInfo = {
 	totalRows: 2,
 
 	// TODO(Someone with hardware): This mapping is guesswork and hasn't been tested
```
Julusian:
```suggestion
```
I am pretty sure I have tested this now and it was fine

docs/7_surfaces/contour_shuttle.md
```diff
@@ -1,8 +1,19 @@
-It is possible to use the Contour ShuttleXpress and ShuttlePro v1 and v2 with Companion (since v3.1.0).
+It is possible to use the Contour ShuttleXpress and ShuttlePro v1 and v2 with Companion (since v3.1.0; note: support for very old versions of ShuttlePro v1 was added in v4.1.3)
 
-Enable support for it in Companion's settings and rescan for USB devices.
+You must enable it in Companion's Settings > Surfaces section and rescan for USB devices. (You do not need to restart for the setting to "take"... since at least v4.0)
```
Julusian:
```suggestion
You must enable it in Companion's Settings > Surfaces section and rescan for USB devices.
```
The intention of the restart message at this point is really just because we don't close already open ones, or do a scan, so a restart is the way to get into a predictable state of what is opened.

(changing soon hopefully)

arikorn: Not so sure here. I tested it with my Contour Shuttle and it came right up. Maybe I did a manual rescan? Anyway, no harm in removing the instruction, so will do momentarily.

arikorn:
Thanks Julian, it feels good to finally have this one completed!!!

PR #3493, opened 2025-06-18
feat(graphics-overhaul): image library

Julusian:
This is a new page in the ui, for making a library of images:
image

This is intended to pair well with the new graphics, and at the same time to help users avoid bloating their config by having the same image being stored on multiple buttons.

For the MVP, I intend to expose the images as variables under names like `$(image:id-here)`. In this way, they can easily be used in the drawing, and especially for combining with conditional logic in expressions.  
I was not intending to do this originally, but I think this fits well into our current flow. A follow up could do some better integration to the elements ui to link things more easily, show previews, and make the process of importing more seamless.
image

When creating an image, you are prompted for just an id. After creating it you can upload the image, and you can easily replace the image with the same button later.

TODO:

- [x] user defined order of images
- [x] should id be called name, and name be description? (matching what we use for custom-variables)
- [x] implement 'collections'
- [x] expose the images as variables
- [x] implement import/export


dnmeid:
I do very much like the introduction of an image library.

This can nicely be extended to modules. Like custom-variables & module-variables. So modules could export images and use just a reference. The user will have much greater flexibility.
How about integrating this completely in the variables system? We could simply introduce a variable type `image` with the image data and some metadata. 

I have some code (don't remember if it is a local branch or remote) to allow rendering of SVG. Is the library prepared to use different image formats?

Julusian:
> This can nicely be extended to modules. Like custom-variables & module-variables. So modules could export images and use just a reference. The user will have much greater flexibility.

I suppose this would be a good reason to start adding a 'type' property to variable definitions, then we could handle them in better ways without requiring much effort from modules.

>  How about integrating this completely in the variables system? We could simply introduce a variable type image with the image data and some metadata.

Half way through doing this it did occur to me that these are essentially custom-variables but with a different ui. My feeling is that this is fine, as it means we can tailor a better experience for these vairables than we can with custom-variables, due to the certainty that they contain images.  
As they are ultimately consumed as variables, that means that a user can use custom-variables instead, at the cost of the less specific ui.

> I have some code (don't remember if it is a local branch or remote) to allow rendering of SVG. Is the library prepared to use different image formats?

Yes, I have tested this with png, webp, and svg. The canvas accepts all of these and I have opted to not limit the formats accepted yet. It is worth noting that svg rendering is not consistent with the browser previews, but it does draw them mostly fine.  
In fact this is a property of this graphics overhaul, the image layer in general will accept many formats (including svg), as long as they are presented as a base64 data url.

PR #3494, opened 2025-06-19
feat: Make the connection-add/edit-specific-connection pages deep-linkable

jswalden:


Julusian:
I'm not keen on how this is just using the router as a datasource for the page.
Could you rework it to use routing properly, the triggers ui should give a reference on how this can be done https://github.com/bitfocus/companion/tree/main/webui/src/routes/app/triggers. Use `<Outlet />` in the parent page to render the contents from the child route
I've made some layout changes which this pr is conflicting with a little (mostly because of the tabs at the top being removed)

jswalden:
I made changes to bring it seemingly within the style of the triggers code, but I'm very much not sure I understand normal React idioms as far as passing around state to outletted subcomponents goes.  Is this right?  Should I be defining a context that all aspects of the connections UI should share?  Happy to fix this up however is necessary for whatever misunderstandings I've had!

Julusian (approved):

Julusian:
> I made changes to bring it seemingly within the style of the triggers code, but I'm very much not sure I understand normal React idioms as far as passing around state to outletted subcomponents goes. Is this right? Should I be defining a context that all aspects of the connections UI should share? 

Yeah, but for these callbacks they are simple enough it is not worth it, duplicating is less complexity and way less effort, and is how I would have done it

PR #3498, opened 2025-06-21
feat: move version number into sidebar

Julusian:
~~I am not 100% certain on this, so throwing it here for any feedback.~~

Its always felt a bit weird/messy to have the version number so prominent on the top header bar. This moves it to the bottom of the sidebar:

image

For a stable build:
image

### DIFF
--- a/webui/src/Layout/Header.tsx
+++ b/webui/src/Layout/Header.tsx
@@ -1,14 +1,5 @@
 import React, { useContext } from 'react'
-import {
-	CHeader,
-	CHeaderBrand,
-	CHeaderNav,
-	CNavItem,
-	CNavLink,
-	CHeaderToggler,
-	CContainer,
-	CSpinner,
-} from '@coreui/react'
+import { CHeader, CHeaderBrand, CHeaderNav, CNavItem, CNavLink, CHeaderToggler, CContainer } from '@coreui/react'
 import { faBars, faLock, faTriangleExclamation } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { RootAppStoreContext } from '~/Stores/RootAppStore.js'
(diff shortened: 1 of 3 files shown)


bryce-seifert:
This is just a mockup for now, but maybe along with this we relocate the warning for updates / beta so it travels along with the version number? 
Screenshot 2025-06-21 at 9 40 22 AM

Julusian:
I'm open to moving that, but I don't mind that 'notification' being in the top. Largely because it looks tidy and designed.

I worry that putting it down there make it less visible and could mean that less users notice it when running a stable version?  
I don't imagine that many users care about seeing the version number that much, so hiding it a little seems reasonable

casperboone:
Good idea! Maybe it would make sense putting it next to the collapse button, then it doesn't take up any additional space, cannot be confused with a navigation button, and is a bit less prominent.

Something like this:
image

Julusian:
ok, now looking like:

image

that was way more tedious than I was expecting to get the css right for that..

arikorn:
1. I like your latest version best since it doesn't increase the size of `nav-secondary`, i.e., the non-scrolling portion. (Though in general, I'm ambivalent about moving it from the top, i.e. I would like it about the same either way.)
2. As long as you're modifying this region, could you change the background color of nav-secondary? I often get confused by the lack of differentiation between scrolling and fixed regions of the navbar.  (or should I just submit a separate PR?) I just tried something really simple and it actually seems OK, after a few moments getting used to it, but you may have stronger opinions and/or want the buttons to match.
~~~
	.nav-secondary {
		flex: none;

		background-color: #aaa;
~~~
image

Julusian:
@arikorn happy with this?

Its a bit more subtle but hopefully gives enough distinction. I've done some other tweaks to the sidebar look. contract/spacing there hasnt been the best 

image

arikorn:
Thank you @Julusian! -- and thanks also for accepting my PRs. Your style change is definitely subtle but I'm willing to try it: sometimes all that's needed is a subtle cue like that.  
I have to admit that I tried for several hours to change the color of those non-scrolling buttons with absolutely no effect. Even when the Chrome debugger said that the background color of the innermost element should be light, it was still black. (I do not consider myself an expert at CSS by any means, so no doubt this reflects my own shortcomings as well.) If you have any insight, I'd welcome it! (You can PM me in Slack if this is too much of a "hijack" of your PR...)

arikorn:
@Julusian -- I just noticed that you pushed the static-region style-change change separately. It looks great! (The static picture above doesn't do it justice...) Thank you again.

Issue #1676, opened 2021-07-28
[Feature Request] Unique ID per button for more scripting flexibility 


yukimizake:
As a user of companion in a heavily scripted environnement, I would like to be able to move buttons on my Streamdeck without having to edit the button position in my scripts http calls every time.
Being able to call a button press with a unique identifier (such as a uuid) rather than the button position may be a solution.

Exemple for a HTTP action : /press/uuid/8c00cb3f-1d31-4505-ad49-5bb812aeb4d2
This uuid being shown in the edit button page, and generated when creating/copying a new button, but not when moving one.

Julusian:
This is something I have been thinking about, but becomes quite a large overhaul of some internal structures.

In my mind, there should be a single list of all the 'button definition', which is then referenced by one or more buttons in the grid. This will provide exactly what you want as each button will have a unique and static id, as well as allowing for the common case of wanting to duplicate a button across pages which will use the same id to tie a button to the grid

willosof:
Cool idea! We should put it on the list for the next major :)

Julusian:
#2138 is about adding user defined aliases for buttons, to be able to reference them by a persistent friendly id.

Once that is done, should we expose every button by their internal ids, or if we should require buttons be 'exposed' to the apis by giving them an alias.  
This falls into some other ideas I have on rethinking the apis, and providing some kind of security in them. 

I don't know if that will make it too hard to use these apis sometimes, but it may be no worse than being forced to address buttons by unpredictable random ids. The benefit of allowing only by aliases is that its a low cost/complexity way for users to limit what can be triggered remotely, while also ensuring the protocol uses meaningful ids

PR #3500, opened 2025-06-21
Add action to quickly create a button shortcut

casperboone:
image

Add the ability to quickly create a button that executes the action of another button and mimicks its appearance. Useful when you need the exact same button in multiple places and don't want to duplicate the button/maintenance. 
We do this all the time to have a specific button available on a bunch of pages, for instance a lights/graphics clear button or a go back to home button. Or when we have create an event-specific page with the most-needed controls.

Uses existing actions/feedback to trigger/look like the target button.

An alternative approach would be to make a more fundamental "shortcut button" type that does not have its own configuration/actions/etc. For now, I chose this approach to avoid a larger maintenance burden by using existing concepts and to have quicker user feedback on the usefulness of this feature.


dnmeid:
There are a lot of requests for things like sticky buttons, or button references or master/slave buttons.
So far I'm not conviced of any of them. Usually the just build up on what we have now and that is far away from being a good UX. I don't think that we should make it even more convenient to create what you call button shortcuts. People will use them a lot and they will end up with more or less unmaintainable configurations. Move one button and everything collapses.

I think the most promising idea for this problem is another abstraction layer between the logic and the controls with the ability to 1:n mapping. Currently a button has a style and actions sets and one set of feedbacks. Currently they belong to the button. If we would introduce a pool of action sets and a pool of feedback sets and a pool of styles and you can use that on any button, then you can easily have as many buttons as you like with the same actions and the same look. This is also described here #2387 (quite long and old proposal, but still quite up to date)

Julusian:
I'm also a little worried about the brittleness of using this, but I'm not sure if that should stop this from being done.
The things proposed in 2387 is presumably years away from being usable, so as there is demand for this now it seems like something we should look into.  
Between now and 2387, maybe once #1676 is done this will change into using those ids to make these a bit less brittle. I have been wanting to make it so that a single control can be situated at 2 locations at the same time, which is another way this could change in future.

Julusian:
Closing this as #4339 is working on adding similar functionality via the new control type approach

PR #3505, opened 2025-06-24
Add a udev file for developing/running Companion on WSL

arikorn:
Add a udev file for developing/running Companion on WSL (Windows Subsystem for Linux).

In order to access surfaces in WSL, the user needs to open up both USB and HID devices. This rules file shows the simplest way to do it.

This PR is related to suggestion I have made for the Wiki page: [Setting up Developer Environment](https://github.com/bitfocus/companion/wiki/Setting-up-Developer-Environment)

Since I couldn't find a way to do a PR on wiki pages, I copied and edited it in Google Docs [here](https://docs.google.com/document/d/1jYic5nYimB9Jvv0Omv-IPfrx_tZpnKVZt-jvUOgVfag/edit?usp=sharing), as per Dorian's request to send him suggestions directly and he would check it in.  Here is the result of the edits (i.e. accepting all of my suggestions) in markdown: https://github.com/arikorn/companion-wiki/blob/main/README.md )

Note: the file _README_ should be renamed _README.md_ so formatting shows up in GitHub. I left the filename alone in order to make the diff usable.

### DIFF
--- a/assets/linux/50-companion-WSL.rules
+++ b/assets/linux/50-companion-WSL.rules
@@ -1,0 +1,2 @@
+SUBSYSTEM=="usb", GROUP="plugdev", MODE:="0666"
+SUBSYSTEM=="hidraw", GROUP="plugdev", MODE:="0666"
--- a/assets/linux/README
+++ b/assets/linux/README
@@ -24,10 +24,13 @@
 
    If you are running headless, run as root:
    $ cp 50-companion-headless.rules /etc/udev/rules.d/50-companion.rules
-   Or for desktop Companion, run as root:
+   Or for desktop Companion (except WSL), run as root:
    $ cp 50-companion-desktop.rules /etc/udev/rules.d/50-companion.rules
-
-   Apply the new rules:
+   Or for desktop Companion on WSL, (run as root):
+   $ sudo cp 50-companion-WSL.rules /etc/udev/rules.d/
+     (if this is your first time running on WSL, please see the instruction for using USB devices at https://github.com/bitfocus/companion/wiki/Setting-up-Developer-Environment)
+     
+   Apply the new rules (run as root):
    $ udevadm control --reload-rules
    And replugging any usb devices that Companion should be able to use.
 
--- a/assets/linux/README
+++ b/assets/linux/README
@@ -28,8 +28,8 @@
    $ cp 50-companion-desktop.rules /etc/udev/rules.d/50-companion.rules
    Or for desktop Companion on WSL, (run as root):
    $ sudo cp 50-companion-WSL.rules /etc/udev/rules.d/
-     (if this is your first time running on WSL, please see the instruction for using USB devices at https://github.com/bitfocus/companion/wiki/Setting-up-Developer-Environment)
-     
+   (if this is your first time running on WSL, please see the instruction for using USB devices at https://github.com/bitfocus/companion/wiki/Setting-up-Developer-Environment)
+
    Apply the new rules (run as root):
    $ udevadm control --reload-rules
    And replugging any usb devices that Companion should be able to use.


arikorn:
Converted Google Doc back to Markdown as per @dnmeid's request on Slack.
Here is a link (note that this is the "final" file, it does not show edits; please use diff to see the suggested changes.)

https://github.com/arikorn/companion-wiki/blob/main/README.md

will edit OP to add this link, too.

Julusian:
I'm wondering if it makes sense to have this udev file that gets shipped with the linux builds, or if it should be a codeblock inside the wiki page? After all, it sounds like we are expecting people to read the wiki for general instructions, so why not keep it all there

arikorn:
@Julusian wrote:
> I'm wondering if it makes sense to have this udev file that gets shipped with the linux builds, or if it should be a codeblock inside the wiki page? After all, it sounds like we are expecting people to read the wiki for general instructions, so why not keep it all there

 My thoughts: (TL;DR: we already do the same for Linux developers)
1. It's possible that someone is using WSL and not for development. This wouldn't be any different than someone using Linux.
2. This just makes it easier on the "beginning" developer -- no need to edit files with "unfamiliar" editors, as in the current instructions, and 
3. it makes the instructions in [Setting up Developer Environment](https://github.com/bitfocus/companion/wiki/Setting-up-Developer-Environment) more uniform -- step 2a currently asks Linux developers to go to the assets/linux page, but then has to say "but if you're in WSL do this but don't do that" Much simpler to just make it all the same.

Julusian:
1. I'm not convinced that it something we should encourage, or why anyone would want to. This trickery to be able to use usb in WSL it not something I am keen on and would discourage it for production.
2. True
3.  Well, the reason for that on linux is so that we don't have somewhere else to maintain, as those rules need updating every time a new streamdeck is released, or a new usb device gets support. Maintaining just those 2 has been a little error prone at times. But this wsl one is is so catchall it wont ever need to change.

Maybe I am questioning the wrong thing; maybe I should be questioning why that udev rules is being so catchall as to make every usb device in wsl be read/write by everyone, and why can't it use the linux-headless rules (I assume its a case of the companion group the rules use doesn't exist, and your user isnt a member of it)

arikorn:
Well, @Julusian , 1 out of 3 is a good start... one more and 2 out of three should be enough? ;)

1. I actually agree with you in principle, but don't see why it shouldn't be made available. If a person needs to use WSL, then they need to do it! I don't see this PR as an "encouragement" but rather as a call for consistency... and I've long since learned that just because I don't see a reason to do something (i.e. use WSL to run Companion), doesn't mean others won't either!
2. We agree.
3. Here I agree with @dnmeid , who [wrote in Slack](https://bitfocusio.slack.com/archives/CAV2ACADU/p1750791734692769?thread_ts=1750199120.030479&cid=CAV2ACADU): 

> I don't see the point of using the fine grained access possibilities for WSL when the user has to attach every device individually anyway

(It's also why we can get away with the simplified rules: the only usb devices visible to WSL are the ones added with usbipd.)

FWIW, I did try the "generic" udev rules but it didn't work, partly because the USB subsystem in WSL is `usb`, not `input` - no way around that. Also, I suspect that `TAG+="uaccess"` isn't working in WSL or something else like that (and, yes, we use "plugdev" as the group instead of "input", which as you imply, is the least problem). 

In the end, this PR isn't a big deal: the main point was to make it easier for a Windows person to become a contributor to core development, -- in my mind every little bit counts -- but if you feel that it's too much change for a relatively small benefit, then of course I defer to your collective opinions.

arikorn:
Good news: I found a way to make (almost) everyone happy: the udev rule is back to one line and doesn't need an editor to install.
 Please see my suggestions for the Wiki page here (since I see no way to make a PR on that part of the Wiki).  If you would prefer, I can shift this to an Issue instead of a PR...

Here's the diffs: https://github.com/arikorn/companion-wiki/commit/bbabe7a38c635723ed2efafa20d140ca1042f8f1
(note: missing from the overview: fix various yarn arguments.)

And here's the edited page: https://github.com/arikorn/companion-wiki/blob/bbabe7a38c635723ed2efafa20d140ca1042f8f1/Setting-up-Developer-Environment.md

Julusian:
> I don't see the point of using the fine grained access possibilities for WSL when the user has to attach every device individually anyway

Yeah that makes sense. 

> Good news: I found a way to make (almost) everyone happy: the udev rule is back to one line and doesn't need an editor to install.

I much prefer this, simply because it keeps it all to one wiki page

> I actually agree with you in principle, but don't see why it shouldn't be made available. If a person needs to use WSL, then they need to do it!

I see your point, but if we encourage it then that means we need to support it to some degree. Perhaps then the headless linux install script should be handling this setup too?  
And for 99%+ of users, running in wsl is the wrong choice, so I don't want it to appear like it could be something we recommend, and cause them more effort and pain.

If someone has a reason for doing this, then I want to understand why. Perhaps it is we are missing something that is trivial to solve, or is there something special about their setup/env that will need to be considered so that we don't break it by accident?

arikorn:
Thank you @Julusian for your feedback here and in my "fork" of the Wiki page. I've updated the page according to our discussion as well as some slight edits. Here's the latest diff: https://github.com/arikorn/companion-wiki/commit/bcf7d148db4cbed2455fcbf224803af6976e71c7

The complete version is at: https://github.com/arikorn/companion-wiki/blob/main/Setting-up-Developer-Environment.md

I did add one question: _when is python needed_? (if it's still needed for Companion?)

If you and/or @dnmeid approve perhaps you could copy it over to the Companion Wiki page and then we can close this out?

Julusian:
wiki page is updated

arikorn:
Thank you @Julusian!  A minor note: it appears that the brackets around `https://learn.microsoft.com/en-us/windows/wsl/setup/environment.` (second paragraph under "Installing WSL") are missing, so that instead of "(Microsoft’s installation instructions)" coming out as a link it is followed without a space by the URL.

dnmeid:
> it appears that the brackets ...

Fixed

arikorn:
Thank you Dorian!

PR #3506, opened 2025-06-24
Feat/syslog-transport

phillipivan:
Adds syslog transport to winston.

Configuration via flags on startup so that transport can be setup immediately, rather than say, waiting for the service APIs to be initialized, and there by missing much of the logging as Companion spools up.

Configuration via electron loader will come in a subsequent PR.


Julusian:
> Configuration via electron loader will come in a subsequent PR.

Maybe this line answers it, but is this something that is wanted for desktop or headless usage?

Because for headless wouldn't it make more sense to use some system tool to collect the system logs and pipe them somewhere? I have used things like filebeat in the past to collect logs for similar into kibana. I assume there is a standard tool to do the same but as this syslog-transport.

If this is just for desktop, it might make more sense to include this extra logging entirely within the launcher.

This is how the current logging is setup; companion expects something above it to handle writing the logs to disk (on linux likely systemd, in docker leaving that to docker, on desktop the launcher).  The reason being that then the logs will catch unhandled errors and won't miss anything at startup/shutdown of the process (and it avoids needing to implement logging twice, as we also want to write logs from the launcher)

-----

regarding configuring this in the launcher. I feel like the best way to do this will be to replace the way the dev modules folder is setup today, to turn that cog to open a 'settings' window.  
however as this launcher ui is a tiny amount of handwritten html/js/css making that ui won't be the nicest. Making that in react does feel a bit overkill though, so dont feel like you need to do that.

phillipivan:
We presently use companion on windows and linux VMs, so that is the use case I had in mind, rather than fully headless - though I guess at some point we should investigate running it in a docker container again.

I hadn't thought of running the syslog integration at the launcher level (I'm only starting to get my head around the internal companion architecture). I can look into that, though the preexisting winston module sure does make the current solution very straightforward. 

I think we are on a similar page regarding expanding the function of the cog, as that was more of less my plan.

Julusian:
> I hadn't thought of running the syslog integration at the launcher level (I'm only starting to get my head around the internal companion architecture). I can look into that, though the preexisting winston module sure does make the current solution very straightforward.

Oh, I thought it was winston in there too, but apparently not. I suppose hookint into the main companion part makes more sense then, especially as I dont see any general purpose libraries which could be utilised

phillipivan:
I used a syslog library for the `generic-syslog` module, however it was very stale and I had to publish my own [release ](https://github.com/phillipivan/node-syslog-client )simply to get around some deprecated NodeJS calls, adequate for what is likely a very esoteric module, but I think the winston module would be a safer bet long term to include in the core companion code.

Edit to add: looking over that library again, I see the bigger issue was that no release had been published or tagged since 2017, so whiles some commits had been made to address the deprecation a new release was not published. Not that that really changes my assessment.

companion/lib/Log/Controller.ts
```diff
@@ -251,7 +251,7 @@ class LogController {
 		try {
 			options.app_name = 'Companion'
 			this.#logger.add(new Syslog(options))
-			this.#logger.debug(`Syslog transport initialized`)
+			this.#logger.debug(`Syslog transport initialized. Options: ${JSON.stringify(options)}`)
```
Julusian, line 254: probably shouldn't be logging the password in this?

phillipivan: There won't be any password there. Possible fields include host, port, protocol, local host name, and app_name. Apart from app_name all of the above are potentially undefined depending what flags have been used to start companion. When undefined the module reverts to defaults.

yarn.lock
```diff
@@ -16589,6 +16628,17 @@ asn1@evs-broadcast/node-asn1:
   languageName: node
   linkType: hard
 
+"unix-dgram@npm:2.0.6":
```
Julusian: This library is going to be a problem.
But it looks like we probably don't need it, so maybe it should be patched out (either as a yarn patch, or replacing the package with a placeholder). I can deal with this unless you think it is needed?

The problem with it is that it is a compiled library, that our build system will have to compile. And we will need to make some build changes to ensure it builds the correct arch for each build.

phillipivan: It looks like that dependency is only used if the protocol is set to `unix` or `unix-connect`. I have only exposed the `udp4` and `tcp4` protocol options (udp4 is default and there is a flag to switch to tcp4) so I think it is not required.

PR #3512, opened 2025-06-26
feat: improved backups control #2961

### LINKED ISSUE #2961: Config Autosave

### Describe the feature

I would like to be able to set companion to save out config backups at a set time interval (ie 1 hour, 1 day, or 1 week). Keeping a specified number of backups, which when reached, the oldest is discarded as the new one is saved.

Similar functionality is available in other software, such as Riedel's Director which we have setup to backup the configuration of the intercom matrix once a day, for a period of 90 days. Allowing us to step back if problematic config changes have been made.

### Usecases

Backup against configuration errors and programming mistakes, etc, in complex setups, or being able to easily step back temporary configuration changes. Obviously not a primary backup tool against total loss of system, but a useful tool in the toolbox.

Julusian:
closes #2961

This allows the user to define some auto-backup on interval
image

The idea is that the user can define different rules, into different paths on different intervals.  
One could be a daily one into dropbox keeping 90 days worth, another could be hourly elsewhere keeping the last 2.

This supports the usual export formats, and the 'raw database' format, which is a full sqlite db. For now that is the default, but if we get to a point where a normal export contains 'everything', then we could change the default format.

### DIFF
--- a/companion/lib/Data/Database.ts
+++ b/companion/lib/Data/Database.ts
@@ -4,6 +4,7 @@ import { upgradeStartup } from './Upgrade.js'
 import { createTables as createTablesV1 } from './Schema/v1.js'
 import { createTables as createTablesV8 } from './Schema/v8.js'
 import type { UserConfigModel } from '@companion-app/shared/Model/UserConfigModel.js'
+import fs from 'fs/promises'
 
 export interface DataDatabaseDefaultTable {
 	page_config_version: number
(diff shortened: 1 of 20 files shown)


phillipivan:
Picking up from your question in the related issue, how does one restore these backups? Can they be imported via the web ui now? Or is a manual job in the filesystem to replace the primary db when companion isn't running?

Julusian:
If you choose the 'raw database' format, then that is a manual filesystem shuffle to restore. It also offers the usual export formats

Issue #3522, opened 2025-06-29
[Internal] Remove deprecated webui Sass rules


arikorn:
### Is this a feature relevant to companion itself, and not a specific module?

- [x] I believe this to be a feature for companion, and is not specific to a module

### Is there an existing issue for this?

- [x] I have searched for similiar existing issues

### Describe the feature

Internally, the webui code uses deprecated `@import` rules, which cause a large number of warnings every time one compiles the webui during development.

While the `@import` rules may not be removed for some time, it seems prudent (and less annoying) to remove them now.  I will submit a PR on this.

### Usecases

(see above)

PR #3523, opened 2025-06-29
Modernize sass code

arikorn:
Remove deprecated `@import` code, as recommended by Sass; completes issue #3522. This PR consists of two commits. **Main changes are in support of replacing `@import` with `@use.`** In the final commit:

1. `@coreui` is loaded in the file *scss/variables.scss* so the Companion variables can override @coreui defaults; Consequently, *scss/variables.scss* **must be** the very first file loaded in *Apps.css*, so that the overrides work. (See the notes, below.)
2. Certain files have `@use` added to the beginning of the file. (See the notes.)
3. (additional cleanup: removed extraneous `.form-control` clauses from _common.scss)
4. (additional cleanup: added `appearance: none;` to `.form-check:not()` for compatibility.)

(Item 1 is the main difference between this and the first commit in the PR: in the first commit local variable definitions are moved from *_variables.scss* to *App.scss*. While I think this first version is a bit clearer for subsequent development, this current commit is more consistent with the existing file architecture.)

Notes:
To minimize changes I’ve opted to import coreui into the global environment: any variables referenced w/o a package identifier are from coreui. (See below.)
The `@use` directive differs subtly from `@import` Here is a detailed explanation:
1.	By default, `@use` establishes a separate namespace for the imported variables (note: css rules appear to load into the global namespace). This can be largely circumvented by writing `@use 'file' as *` with important caveats:
2.	 Two imported files cannot define the same variable into global space. For example, the files *`_variables`* and *`@coreui`* cannot both define `$primary-color` if both are imported using `as *` As a result it is not possible to override default variables in a module such as `@coreui` by simply loading the overrides into global space. Instead:
3.	To override default values in an external module (such as `coreui`), the `@use` line must send the variables using `@use 'module' with ($var1: 'a', $var2: 'b', …)` However if `$var2` does not exist in the module, Sass generates a very misleading error about multiple definitions. On the good-news side,
4.	A module will keep the same configuration (or lack of configuration) even if it’s loaded multiple times. Because of this, `@use ... with` can only be used once per module, the first time that module is loaded. (Quoted from Sass docs):  so submodules don’t need to use the `with` clause.
5.	Unlike `@import` which acts like a `#include`, code in files imported with @use is not “aware” of the file into which they are being inserted. This means that if the sub-file references other variables, css rules, `@extend` or `@include` clauses it will need to import the relevant modules/files with `@use`. (note: `as *` does not appear to be needed to reference imported css rules.)
5a.	`@use` uses the file's location as its current directory, so whereas the path in *Apps.scss* is `scss/subfile`, the path in other subfiles is just `subfile`
6. Note: two warnings are unrelated to Sass: the warnings about `alert-variant()` and `avatar()` appear to be generated by `@coreui` regardless of if they're being used.

### DIFF
--- a/webui/src/App.scss
+++ b/webui/src/App.scss
@@ -1,72 +1,37 @@
-@use 'sass:color';
+// IMPORTANT: 'scss/variables' MUST be the first @use rule: it overrides the defaults in @coreui
+//  See Sass @use documentation: "A module will keep the same configuration ... even if it’s loaded multiple times"
 @use 'scss/variables'; // local definitions not related to @coreui definitions
 
-$local-header-bg: #d50215; // needed for both $header-bg and the adjust() call, below.
+// let coreui (with the overrides defined in _variables.scss) be the default source of variables.
+@use '@coreui/coreui/scss/coreui' as *;
 
-// pass all remaining local definitions that are meant to override @coreui defaults:
-// note that Sass will issue a strange error message if a variable specified here is not already in coreui. (As of Sass 1.89)
-@use '@coreui/coreui/scss/coreui' as * with (
-	$primary: #d50215,
-	$secondary: #ccc,
-	$body-bg: #333,
-	$light: #e9e9e9,
+// note: if a variable is defined in the following imports, they will need to be reference below by module name.
+@use 'scss/react-time-picker';
+@use 'scss/react-date-picker';
 
-	$header-bg: $local-header-bg,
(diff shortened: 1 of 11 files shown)


arikorn:
More good news, @Julusian.  My PR to CoreUI has been accepted:  with the next version of `@coreui` we should be able to compile webui with no deprecation warnings.

PR #3527, opened 2025-07-01
feat: custom variables overhaul

Julusian:
This is not complete but enough of the plan is visible for any input on the feature/ux.  
I am thinking of backporting this to 4.1, unless it turns out to be too tangled into the graphics-overhaul.

This builds upon the changes made for [local variables](https://github.com/bitfocus/companion/pull/3314). The result is a pretty much from scratch rewrite of the custom-variables system.  

The basic idea is that these variables are now feedback based. To support this, a new 'value' feedback type is added (not yet supported from modules, that will be a follow up to this). The 'traditional' custom variable is then reimplemented as a value feedback (with some special handling for a few flows), and a new `internal: Evaluate Expression` is added to expose a 'live' expression as a variable.

These generally behave like any other variable, with the execution cycle of any other feedback.

Not final or polished ui, but it looks very similar to before:
image

The 'headline' field of the editor has been overridden for these variable feedbacks, to include the variable name.

TODO:

- [ ] Testing
- [ ] upgrade old configs
- [ ] upgrade old exports
- [ ] Make sure the old internal actions work
- [ ] Fix ui layout bugs
- [ ] Show the current value in the editor ui?
- [ ] fix drag and drop/order
- [ ] Verify the ui only offers internal actions for user_value ones


dnmeid:
So to summarize, the plan is to have a new way of generating and passing values in Companion.  
A value can be set as a literal like the current variables, it can be the result of an expression or it can be the output of a feedback (currently only boolean but other values in the future as well).  
Values can have different "scopes": button/local - the prefix is `local` and the name can be the same on different buttons; module - the prefix is the module name and actually this is a global accessible value then; global - the prefix is `custom` and the name needs to be unique.  
Values can be used in expressions or variable-enabled fields by referencing the name. Within one scope the name has to be unique and the user is responsible for this. We can invent a system of checking and refusing duplicate names on entry but care needs also be taken of duplicate names during import.

My current thoughts are:
- I think the image library should be merged with this. An image would then just be a value. It would be an object like `{companionValueType: 'image', metadata: {mime-type: 'png', dimensions: {x: 64, y: 120}}, value:'base64 encoded png'}`. Then users could store their images in the global scope or module scope or button scope. When the UI sees a value of type object with the property companionValueType we know that it needs some special treatment and presentation.

- I'm not sure yet but tend to that the non-custom values should also be part of this. E.g. like a read-only, no rename version of a value, but presented in the same list.

- The UI needs definitely to be streamlined before this can go in production. A lot of users tend to have hundreds of custom variables because previously arrays or objects wasn't available. We need a slim design for presenting long lists. A lot of the icons should go away. The drag icon is not needed, the item can just be draggable when picked anywhere within the item but outside of an input/button. The trash icons can be replaced with a select and delete workflow with a keyboard delete key press or a global delete icon. The expand/collapse icons should be more a tree style triangle, that that serves a a collapse/expand button but at the same time shows that this is a node of the tree and also the level of the node. The edit icon is not needed. Text can just be editable when clicking on the text.  
Icons and basically all elements of a node should not change positions when collaped or expanded. There should only be some things that are revealed or not.

- One thing I'm not a big fan of is the "ungrouped" behaviour of items in a list when they are not in a collection. My suggestion is that a list can have individual items and also groups of items in any order. It should be possible to have e.g. item1, item2, group1, item3, group2, group3, item4... No need to have a collections section and one special collection for ungrouped items. When a group is deleted, all children in the group should also be deleted and not end up somwhere else. I think a confirmation like "do you really want to delete the group and all of its content?" should be enough.

These UI considerations are not special to this PR, I think a more uniform and lightweight style of presenting lists would be very beneficial. I think a tree design is well established and every user will immediately know hot to work with it. Here are some general guidlines: https://design.mindsphere.io/patterns/tree.html  
I think we can use a little more blocky design to adopt a little to the current UI, like this: 
image
We would need pseudo-items for "add new..", I think we should keep this behaviour.
With a tree it is easy to allow the user to have any number of child objects or child groups, but it will be a little challenge how to handle our nodes that have a fixed number and type of children. Like the conditional has exactly three child lists: one for the condition, one for true-branch, one for false-branch. It is no problem to add these when creating a conditional, but we need a visual hint for the user that they are actually one entity and one cannot just delete or move e.g. the conditional.

Julusian:
> I think the image library should be merged with this. An image would then just be a value. When the UI sees a value of type object with the property companionValueType we know that it needs some special treatment and presentation.

I am not immediately convinced by this. As you say elsewhere, the ui needs streamlining, and I think that whatever we can do on this will not be as nice to use as something intended to show thumbnails.  
And by having the image library as its own concept, that will help reduce the number of items we have to display in this ui.

> I'm not sure yet but tend to that the non-custom values should also be part of this. E.g. like a read-only, no rename version of a value, but presented in the same list.

I don't quite follow what this would entail, but I think we need to be careful there, as we dont currently handle the number of variables some modules have well, I have experienced the slowness it can cause or crashing of the ui.  
But this sounds like a follow up to figure out.

> The UI needs definitely to be streamlined before this can go in production. A lot of users tend to have hundreds of custom variables because previously arrays or objects wasn't available. 

I dont think the ui is any worse for this in 4.1/this pr than it was before, the fact that we have groups makes it more managable surely?

> Icons and basically all elements of a node should not change positions when collaped or expanded. There should only be some things that are revealed or not.

That is mostly how it was working before this, which meant that some icons which maybe should be hidden when collapsed are not.  
So this partially conflicts with saying to hide icons.

This could be a separate discussion on whether to move some of the buttons we show in the left list to be buttons at the top of the editor. It wouldn’t be unreasonable for them to be placed there

>  The drag icon is not needed, the item can just be draggable when picked anywhere within the item but outside of an input/button. The edit icon is not needed. Text can just be editable when clicking on the text.

I'm open to trying it, but I'm a little worried about the UX of this. Differentiating between clicks and click+drag is going to cause some pain for both us and users.  

> The trash icons can be replaced with a select and delete workflow with a keyboard delete key press or a global delete icon. 
> The expand/collapse icons should be more a tree style triangle

I don't have a problem with these, other than this is reusing components from the action/feedback editors elsewhere, so will need some care to not break those (unless they also want to change

> When a group is deleted, all children in the group should also be deleted and not end up somwhere else. I think a confirmation like "do you really want to delete the group and all of its content?" should be enough.

One thing to note is that this collections ui was originally written for connections (with the implementation being reused across triggers and custom variables), and having that mass deleting there makes me a bit nervous. But if we want differing implementations of certain portions that would be doable.

> One thing I'm not a big fan of is the "ungrouped" behaviour of items in a list when they are not in a collection. 

A simple option here is to simply remove that heading. I think that dates from before groups could be nested, so it made sense to have the 'fake group', but perhaps not so much when behaviour inside a group is a little different

> My suggestion is that a list can have individual items and also groups of items in any order. It should be possible to have e.g. item1, item2, group1, item3, group2, group3, item4...

I have a technical argument for it being the way it is; interleaving the lists like that likely means rewriting various chunks of code to either rework existing logic/data structures to handle this, or be able to handle the interleaved sorting between the two lists.  
This may not be that bad to do, but as I am the kind of person who usually has my file browser set to show folders at the top it didn't occur to me that this might be wanted.  
~~In particular, the custom variables are entities, and share the same entity implementations as elsewhere. that makes putting them in collections at all a pain~~


>  I think we can use a little more blocky design to adopt a little to the current UI, like this:

So far I have been trying to avoid anything that looks too different to what we already have. I am fine to do something else, a mockup/screenshot of how to make it look would be appreciated to get me most of the way there.  
Of course we should also think about where else will want to get the same look to make things consistent

> Like the conditional has exactly three child lists: one for the condition, one for true-branch, one for false-branch. It is no problem to add these when creating a conditional, but we need a visual hint for the user that they are actually one entity and one cannot just delete or move e.g. the conditional.

This part sounds like you are proposing how we display all actions/feedbacks and their children? I didn't get that impression from elsewhere.  
I'm not opposed to making those nested lists of entities collapsible, that might be a good thing.

------

As already said, few of these things are larger changes to our 'design language' for pages/tables like this, so arent exactly in scope of this pr, and could be a precursor or follow up.

A thought I have just had, maybe this custom-variables page should now use the 2 pane layout. With the left being the tree with everything always in the single line view. (much like we have for connections/triggers)
That would make it consistent with what we have elsewhere, and would probably make some things a bit easier.
Perhaps I should restart this again, making each custom-variable its own 'control' would fit with that approach, and avoid some of my current complexity/concerns.

Julusian:
Something along these lines:
image


~~That single feedback in the editor needs some more unwrapping/special rendering to integrate it better with the rest of the page (and remove the bits of ui which arent relevant for it).~~  I think this will mitigate some issues with the list being too long, as the rows in the left panel cannot be expanded to be any taller than they are shown here.

One downside I see is that it is more clicks to create and setup the variable (it starts off with no 'feedback' selected as its root element), and it is no longer possible to drag feedbacks between variables. (Which is likely not a significant issue)  
And this makes it a bit different to how local-variables are shown/arranged, due to the more limited space available within the button editor (and the collections probably wont be necessary there)

dnmeid:
I didn't have time so far to think about it a lot, but what about this idea: the "add custom variable" could be a dropdown like we have it for actions and so. The constant returner (do we have a name for it?) could be the default choice and then you can create it with one click again. For the other types the user would have to select a different type. Of course I don't want to have different types of custom variables, I'm just talking about prepopulating a first item.

Julusian:
> I didn't have time so far to think about it a lot, but what about this idea: the "add custom variable" could be a dropdown like we have it for actions and so. The constant returner (do we have a name for it?) could be the default choice and then you can create it with one click again. For the other types the user would have to select a different type. Of course I don't want to have different types of custom variables, I'm just talking about prepopulating a first item.

The old behaviour is called 'User value' currently (as we agreed upon for local variables).

I'm not sure about making the add button a dropdown or anything, I worry that will be cluttered/messy to do nicely.
But I am wondering if putting two add buttons 'add simple variable' and 'add computed variable'. That will give the easy button to get the old/simple behaviour and then one for everything else.

Or something I am now wondering, perhaps this should be a fixed entity, and each custom-variable should itself have local-variables? I can see value in that, as you could use similar composition logic to elsewhere but without needing to do it all in one feedback/expression.

eg:
```
$(local:input) = feedback returning the input number on a videomixer
$(local:something) = something based around a logic:and, using $(local:input) as a value to one of the feedbacks

the output being an expression like: $(local:something) ? $(local:input) : '-'
```
this could be done by breaking out more custom-variables, but being able to use local variables makes it self-contained composition. It also avoids polluting the global space for these intermediary values, and being self-contained makes them easier to duplicate and tweak.

If we go this route of supporting local-variables, then maybe there are only 3 possible types (aka entity at the root) of these custom-variables:
1) user value
2) expression value
3) logic (maybe?)

I know we pushed away from this rigidity for local-variables, but maybe it makes sense to do so here. 
We could keep the root entity as being anything (like it is right now), but I think that makes having local-variables here more confusing, and I worry could mean we will need to make it possible to change the root entity of the variable.

If we do support changing the root entity, should we dispose of the local-variables when clearing this entity, or when selecting a new entity type that doesn't support them (making for some messy UX&code to do that prompt).

Julusian:
For now I am starting with having two create buttons, the 'computed' one isnt currently filling in an entity, but I am still wondering about doing that.
image

I am also intending to expose local variables inside of here too, I think that will make these become really powerful reusable computations

Julusian:
A slightly stupid example, but it works. And I think this shows off why I want local-variables here. Without local variables, I would have to do that `$(local:abc)` as its own custom-variable, which could then be used here to translate that boolean into text. This allows me to achieve the same result, much more self-contained.  
And this will get more important when modules start making 'value' feedbacks, as they will need to be stored into a variable before they can be used by an expression; again local variables allows that to be done locally

image

lukescott:
I think [Add Simple Variable] [Add Computed Variable] may be a bit confusing. Especially when there is a "Set Variable" action to set the value. I would assume that these computed variables shouldn't show up in an action because they are read-only, correct? (Or many of these would have a read-only property?)

Perhaps it would make more sense to call these 'computed values'. So your buttons could be:

[Add Variable] [Add Computed Value]

The UI is awesome for defining variables / computed values. I love the organization with grouping and sorting. Even if sorting serves no other purpose than to please my OCD, it allows you to keep things organized.

The one thing that is lost by this is the ability to quickly lookup values. You have to drill down into the definitions. So I think it might make sense to call this 'Value Definitions' or "Variable Definitions" and then have some separate "Values" screen where you can search for variables / computed values and see them in a simple table.

By the way, I've been playing with the local variables in buttons, and this is really exciting! I'm assuming modules will also be able to define these computed variables/values? If Feedbacks are simplified, I almost wonder if it makes sense to call 'new feedbacks' 'layers' since the simplified version can just refer to these new variables/value types.

Another random thought, perhaps module variables could be customizable a bit more. Like a module could define a variable as settable in the 'Variables' UI. Perhaps provide a dropdown of possible values. Could replace a good chunk of the configuration, making the values addressable as values in buttons.

Julusian:
Thanks for the input!

> The one thing that is lost by this is the ability to quickly lookup values. You have to drill down into the definitions. So I think it might make sense to call this 'Value Definitions' or "Variable Definitions" and then have some separate "Values" screen where you can search for variables / computed values and see them in a simple table.

Yeah thats a fair point. I suppose the question is whether to do this in a modal like offered in the connections page, or to keep the old page under the variables heading. Or both is also doable, but might be a bit overkill 

Or maybe it could be added to the list, but I don't see any space that would be a good place to put it on each row

> I would assume that these computed variables shouldn't show up in an action because they are read-only, correct?

Correct, only the 'simple' ones will show up in those places which want to set them.

> Perhaps it would make more sense to call these 'computed values'. So your buttons could be:

I'm not convinced by 'values', as these are still called variables elsewhere in the system (also the modules have variables, even though they arent settable). I am open to the prefixes of both of these buttons though; maybe 'simple' should be dropped, but I worry that would be more confusing.

> By the way, I've been playing with the local variables in buttons, and this is really exciting! I'm assuming modules will also be able to define these computed variables/values? If Feedbacks are simplified, I almost wonder if it makes sense to call 'new feedbacks' 'layers' since the simplified version can just refer to these new variables/value types.

Yes, once there is something for local-variables exposed to users then the same 'value' feedback type that the 'internal: Evaluate expression' uses will be exposed to modules.

lukescott:
> maybe 'simple' should be dropped, but I worry that would be more confusing.

I think for me, this was a signal to take a step back and re-evaluate things a bit. In the spirit of that:

> modules have variables, even though they arent settable

Why not? If we break this down a bit and look at the potential value, why not let modules define variables that can be set by the user? If I were to look at the pure benefits of that:

- This could replace a lot of configuration. (Configuration fields already overlap with actions and feedbacks)
- Module configuration values could be addressable in actions/feedbacks/rules/etc.
- A module wouldn't have to copy configuration into a variable if it was defined like that to start.
- A module could define a custom type of settable variable. Perhaps an enum, with a drop down of possible values.

Then you can start to look at 'variables' and 'computed values' as two separate things. They could share the same 'references' foundation of being able to reference these 'variables' or 'values' in expressions (as in `$(prefix:name)`). 

I want to expand on the comment of 'they aren't settable'. Currently, you can't set them in the 'variables' screen, but you can set them with a 'set variable' action. To me, this further reinforces the need for these to be separate first-class things (that share the same foundation). -- If a module wanted a read-only variable, instead of using 'setVariableValues', it should be providing a getter on the variable/value/reference definition. If it is a settable variable, it should be able to provide a default value in the definition, but later be able to update it with 'setVariableValues'. But using 'setVariableValues' on a computed value _should_ be an error.

I do want to clarify when I mean 'replace a lot of configuration' I mean the configuration presented in the connections tab. A module can save values in the configuration that are not defined in the connections tab, and I think there is some separate value in allowing for that. Like persisting and saving values that can be used internally for computed values.

Julusian:
> Why not? If we break this down a bit and look at the potential value, why not let modules define variables that can be set by the user?

That pitch is sounding a lot like https://github.com/bitfocus/companion/issues/1187, which is something I would still like to do.  
I do have some thoughts on more specific bits, but this is rather tangential so not important to figure out here.

> Then you can start to look at 'variables' and 'computed values' as two separate things.

Maybe so. But I am not sure that it is beneficial to users to start calling some things 'values'. For consistency we should probably be calling all module variables values now, and so many parts of the ui will need to be updated to reflect this differentiation.  In most cases (everywhere except the places which can change a variable and not a value), it wont matter to the user whether `$(custom:abc)` is a variable or a value, so is it important to highlight the difference?

And I worry that 'value' is too generic of a word for this. A 'value' is something I put in an action option field, so calling these 'variables' 'values' will confuse some users. (many times users don't understand that not all 'variables' are 'custom variables')

So I am keen to keep these scoped as some kind of variable, as it keeps things simpler and clarifies what they are.  
I did consider 'add value variable', but perhaps that implies that the 'computed variables' dont have values. I did consider 'add variable' but that suggests that computed variable is a subset of that which isnt the case. So decided on 'add simple variable'. But I am very open to changing the wording of this if we can agree on what it should be

thedist:
So to give some feedback, I'm not a huge fan of this change, and here's an example of something I just quickly made:

chrome_Z1ardxOxkr

Say that I create a set of variables, and at some time later want to go back and look through them. What are their values? I have to click through each and every one. Not a big deal for 3 variables but for larger setups it becomes a significantly worse user experience. What type are these variables? Again, need to click through each of them, and only then would it be shown that `$(custom:anotherComputedVar)` isn't a computed variable at all but a simple variable.

Contrast this so the current system. If a user wants to check on some variables, a custom variable is still a variable so they go to the 'Variables' page, then the Custom Variables page as hierarchically that makes sense, and at a glance they can see variables, their values, their type. All of these are immediately editable and can be scrolled through. So in this design aspect, this branch is a detriment to user experience as you're adding additional steps to view data, reducing how much data is immediately available to users, and adding complexity variable creation.

As for the simple variable / computed variable naming debate. Why do they need to be separate? Could it not just be a "Custom Variable", and have a "Type" field with a dropdown for text, number, boolean, JSON, Expression, etc... This would also allow for the advantage that if a user has created a variable and used it throughout there setup it would be easier to change it from a text string to an expression, or vice-versa (yes this could be done by renaming the old custom variable and creating a new one in it's place so any actions and such would not need changing, but that's still more steps for end users).

As for users changing variables in modules, and replacing configuration with variables, I'm not quite sure I see the benefits and if the plan is to replace some of the user configuration with user set variables, will you also need to make changes to the upgrade scripts system as if a module is utilizing user-defined data (be in from the Config screen, or through some variable the user modified the value of) then it could result in breaking changes if module developers can make changes to that data like they currently can to actions/feedbacks/config.

Julusian:
> Say that I create a set of variables, and at some time later want to go back and look through them. What are their values? I have to click through each and every one. Not a big deal for 3 variables but for larger setups it becomes a significantly worse user experience. What type are these variables? Again, need to click through each of them, and only then would it be shown that $(custom:anotherComputedVar) isn't a computed variable at all but a simple variable.

I hear these concerns, but I have no idea on how to arrange it in this table nicely. A sketch would be appreciated.

> Why do they need to be separate? Could it not just be a "Custom Variable", and have a "Type" field with a dropdown for text, number, boolean, JSON, Expression, etc... 

I almost did do this, but then had the thought of that the local variables can only be used with the expression type, and should be hidden for the other types (to avoid confusion). So changing to/from expression mode will show/hide the local-variables section. And in doing so, it really ought to be disposing of any defined local-variables which needs to have a confirmation modal to make it clear that this is destructive. I think this will lead to more confusing UX than having them as separate 'types'.

thedist:
> I hear these concerns, but I have no idea on how to arrange it in this table nicely. A sketch would be appreciated.

So to break down some of the issues I would have as a user. In this pic you can see there is a LOT of empty space, it's not providing any control to the user or displaying information. (Yes, long names/descriptions would fill some of the area on the left, and for fancy variables some of the area on the right has room for local variables, but still largely wasted space).

Also the circled area in the top left, the expand/collapse icons are different widths, so when toggled it shifts the collection text left/right. This should be a fixed width.
chrome_iEK8LjZ4J7

Here I've removed the entire right panel, it's now just a long list of Variables. Description moved to the same line as name (although wouldn't be bold, it just is here as it was a quick change).

The <br> was removed making each variable item more compact and displaying more on a page.

Basic information about the variable is clear and displayed immediately such as type, value, persist, startup, etc... Full control over the variable could be done through clicking on that item and it expanding and giving the capability to add local variables, edit name/description, etc... then when not selected it would collapse back into the more compact display. To keep it organized when in the compact view we could also hide any overflow, so that long descriptions don't distort the sizes of each custom variable box, but when expanded show the full text of anything that had overflow hidden.

If there was an error with a variable (such as the type being set to JSON and the JSON being invalid) there should be some sort of red background to indicate an error, and an exclamation mark symbol with a tooltip giving an error message. (As a separate issue, the JSON type needs fixing as it's erroring on completely valid JSON and when clicking out of it resetting to the previous value).

Collection expand/collapse fixed to 10px, toggling it wont move other elements in that row.

Example of some of the changes, keeping in mind that if this was populated with actual names, descriptions, values, persist checkbox, and all of that then that white space would actually be filled with useful information
chrome_dsZTAJw3fy


While I'm sure there are many features and capabilities we would like to give end users we need to keep in mind that most users are not power users, and every additional click a user has to do, or page they need to go through, to view/edit what should be very straightforward and basic functionality as Custom Variables, will be a detriment.

Julusian:
> In this pic you can see there is a LOT of empty space

The empty line was unintentional and I simply hadnt noticed as it doesnt appear in firefox. Turns out triggers had the same issue.

> Also the circled area in the top left, the expand/collapse icons are different widths

I was sure I had already fixed that, but I guess it broke during some refactoring :disappointed: 

> Here I've removed the entire right panel, it's now just a long list of Variables.

The reason I added the right panel here was partly because it was forming a mess in the left panel when doing a computed variable, and partly to match what we do elsewhere (buttons, triggers and surfaces are all this setup).  

And tbh, more than half screen width usually makes these things be either weirdly spaced out or ends up with lots of empty horizontal space between things you want to interact with. You can see that in your screenshot; having the clone/delete buttons all the way to the right is so far away from anything else, its going to be tedious mouse movement to use those.

I don't see the computed ones as being nice to use without the right panel. The options I see here are:
* Expand the row. but with it being easily multiple screen heights tall among a long list that will make navigating it annoying as it wont have its own scrollbar (no easy way to jump to the top, or see how far through you are)
* Use a modal for the editing. Which we have just removed from surfaces due to it being inconsistent with everywhere else
* Bring in the right panel when needed. Which could mess up the scroll position, and crush the content of other rows in the table to be a mess.

I'm curious how you think the computed variables should be shown in your table.  Would that second line just be type and value, or would you want the multiline expression there too?  
We don't currently store a type, the ui is either in string or json mode with the type then inferred from the current value. But this is changeable, or could behave the same

> keeping in mind that if this was populated with actual names, descriptions, values, persist checkbox, and all of that then that white space would actually be filled with useful information

How will that degrade on smaller screens/windows? Should it start splitting those values across multiple lines as it gets narrower? Also consider that the value(s) can themselves be shown multiline 

Dynamic heights of rows in these tables makes things a pain, it messes with the drag and drop logic, and scroll positions, so I am keen to avoid that if possible. (although it might be a hazard of adding the value back to this table

Julusian:
Another option to consider is perhaps this shouldnt be custom-variables, but an entirely new concept with its own page. That way the simple variables can keep a simple in-table view, and the new page can have this split view.  
This is kind of what I did for #3493, as the motivation for that being separate was mostly to be able to present the table as a grid of thumbnails instead.

The downside there is splitting out even more pages likely means more overwhelming for some users, which I think this has a higher hazard of as they also might not understand what page is for

thedist:
example1
example2

So a couple examples of an Expression Variable (I think that's probably a more suitable name than computed variable, as it uses expressions, users know what expressions are already and will be a less confusing transition). In the first pic the variable isn't selected so it's only show a 1 line of the text box for Expression (I only thought of after photoshopping that it'd probably be better in the collapsed design to just show the value, not the Expression).

But when selected it would expand to show all of the configuration details for that variable.

So, at a glance you can immediately see all variables and their values. If you want to edit anything, click it and it'll expand.

As for usability on small screens, keep in mind that you've currently got 2 panels, and all of the padding/margins that come with those to take up space, go down to 1 panel and that gives you a little more room to begin with. Secondly you could have elements such as current value and startup value side by side if the screen width allows, but after a certain point it would just go to individual label: entry, on each line. On smaller displays than that you could even go to splitting it so the label and text entry are on different lines.

As for efficient use of space on small screens, keep in mind that the UI uses lots of padding everywhere, for example in these next 2 screenshots there's the default padding, and on the second image it's the exact same content but all I've done is tweak some of the padding values so text input boxes aren't so large, which also has the added benefits elsewhere in the UI such as triggers, or buttons with lots of actions/feedbacks. as it allows the user to keep much more relevant data to what they're currently doing on screen and prevent them needing to scroll quite so much (ever try to show someone screenshots of a button to help with with setup, and you have to take 3 or 4 shots just because it doesn't all fit on 1 page?).

chrome_so58gE1Lgd

chrome_63ywbtkiAC

So there's things that can be done with the CSS of the Web UI in general to improve display on small screens. To give a real world example of an App doing such tweaks, have a look at Discord where they have options to alter the css between 'compact', 'default', and 'spacious', where spacious would be equivalent to what Companion does for spacing.

Julusian:
> But when selected it would expand to show all of the configuration details for that variable.

ok so this option:

> Expand the row. but with it being easily multiple screen heights tall among a long list that will make navigating it annoying as it wont have its own scrollbar (no easy way to jump to the top, or see how far through you are)

A common scenario I am thinking of here, is adding a new local-variable, and then using that in the expression. That means scrolling to the bottom of the expanded row and then back to the top. Being able to do that without precise scrolling will make that much smoother to do.

But the biggest concern I have with this though is consistency. It think it is bad UX for this page to do one thing and for everywhere else to do something else. 

> As for usability on small screens, keep in mind that you've currently got 2 panels, and all of the padding/margins that come with those to take up space, go down to 1 panel and that gives you a little more room to begin with. 

Sure, but on large screens it gives way too much space. And having it will encourage using it, then needing the thought on how to degrade.

Have you noticed how the settings pages are only 50% width of the page? That is a similar thing of they look silly being 100% width, but I have no idea what should go to the right of them, so its just empty space.

> Secondly you could have elements such as current value and startup value side by side if the screen width allows, but after a certain point it would just go to individual label: entry, on each line. On smaller displays than that you could even go to splitting it so the label and text entry are on different lines.

On smaller displays I would probably want to hide some of those options too. I rarely care about persisted and startup values, so seeing them taking up 33% of the row height is excessive. But maybe that wont be possible if all editing is inline..

> As for efficient use of space on small screens, keep in mind that the UI uses lots of padding everywhere,

No real argument there other than I am not good at design, so you won't find me trying to restyle everything.  
If I do anything towards this, it will be replacing the ui toolkit we are using with something else, and that may or may not change spacing (but would help on the darkmode). And that will likely result in a bunch of other changes that people disagree with so I haven't yet decided it as a good idea (even though I really hate the current library)

lukescott:
Sorry still catching up/reading some of the other comments. If I missed a point that was already address, sorry about that.

> Say that I create a set of variables, and at some time later want to go back and look through them. What are their values?

@thedist  I think this is easily solved by having two separate screens:

- Define Custom Variables
- View Variables

I'm not trying to come up with great names here, but the general idea is "View Variables" is basically very similar to what the current 'Variables' screen currently is, but just for viewing the current state. You wouldn't have inputs of any kind. You could have a search box and filters to find the one you are currently looking for.

The 'Define Custom Variables' screen is what @Julusian currently has in this PR. It would purely be meant for defining custom user variables or computed variables.

I don't think trying to cram everything into one screen makes sense.

> That pitch is sounding a lot like https://github.com/bitfocus/companion/issues/1187, which is something I would still like to do.
I do have some thoughts on more specific bits, but this is rather tangential so not important to figure out here.

I like this a lot, we should definitely be doing that.

In the short-term for 4.1, could we expose local variables in buttons as the 'value' type for now? Similar to what 4.0 has for defining global variables. I think that in the short term, even simple local variables brings a lot of value to Companion.

Computed variables/values are extremely useful, but I think the UX probably needs more time to bake before it gets rolled out.

> Maybe so. But I am not sure that it is beneficial to users to start calling some things 'values'. For consistency we should probably be calling all module variables values now, and so many parts of the ui will need to be updated to reflect this differentiation. In most cases (everywhere except the places which can change a variable and not a value), it wont matter to the user whether $(custom:abc) is a variable or a value, so is it important to highlight the difference?
>
> And I worry that 'value' is too generic of a word for this. A 'value' is something I put in an action option field, so calling these 'variables' 'values' will confuse some users. (many times users don't understand that not all 'variables' are 'custom variables')

As I said, naming is hard 😄 . Going along with the ideas in #1187, what if they were called properties?

In general, I think the problems that need to be solved are:

- Making sure adding custom 'value' variables are simple, and require fewer steps for the user, similar to how it is now. It doesn't have to be exactly how it is now, but it should require as much or less cognitive thought.

- Avoiding qualifiers like 'simple' as that isn't really clear. As in, what does 'simple' mean? You an I understand it, but a general use will not. It just signals to me the need to be concise and clear about the terms/verbiage being used.

And we don't necessarily have to use two buttons here. We can still have an 'Add Variable' or 'Add Property' button and simply default to the 'value' type. Have a checkbox or input selector (drop down?) to select the advanced cases.

thedist:
> I don't think trying to cram everything into one screen makes sense.

The current Variables screen, which the community is used to, does have both display of current values and capability to edit all on one page. So moving to a system where users have to click a each variable to see the value, or having to jump between different pages to see variables and change their values, is going to be a detriment to the user experience. 

If that detriment is outweighed by the benefits that is yet to be seen, as keep in mind the majority of Companion users are not power users so would only experience this detriment and not the benefits of Expression Variables.

More capabilities is great, but I'm just hoping people don't get tunnel visioned on making fancy stuff and ignoring the impact to users. There's splitting 'Custom Variables', into 'Simple Variables', and 'Computed Values', there's the discussion of potentially changing the naming scheme of things, splitting variable viewing/editing over 2 pages, introducing the ambiguity of variables, values, and properties, etc...

To me it should be as simple as `x = y`. Which makes `x`  a variable, doesn't matter if it's defined by a module, Companion, or the user, or its type, it's a variable and it references some value. `y` is a value, doesn't matter where it comes from, it's what is ultimately parsed and given to options, button text, whatever. This largely falls in line with what we currently have, it's just variables have the distinction of instance variables, custom variables, or internal variables.

What benefit does it have to end users to rename what they are used to, not just from past Companion experience but from other apps/coding too where the variable/value terminology is quite standard.

lukescott:
@thedist I share your concerns. I think the experience will have to change, but I want to make sure the experience is not a big leap from what is there now. I'm looking at this as different problems to solve:

1. What variables do I have and what are their values? Great for debugging actions/buttons that interact with variables.
2. Defining new custom 'simple' variables, close to what is there now. Without extra steps.
3. Adding a 'computed' expression that provides a read-only value. But doing this in such a way where it benefits power users without confusing non-power users.

I'm not sure what the final solution will be. This really isn't a code problem, but a UX problem. It needs more time for careful consideration.

I think there is just too many good ideas to consider that we should look at breaking these apart into separate releasable chunks. Essentially make the problem smaller, give yourself more time in areas that need it.

Local 'simple' button variables are one potential chunk with a lot of immediate value. Collections are another one, at least from an organizational perspective.

Taking some of these ideas and expanding module functionality would be a good opportunity to dog food this stuff with developers (power users) before changing anything for general users. It would allow some of these concepts to be built and tested behind the scenes while allowing time to work on UX.

thedist:
> Adding a 'computed' expression that provides a read-only value.

Why does this need to be read-only? It's essentially a custom variable, where the value is an expression that's evaluated when the variable is used. We have Actions like `internal: Custom Variable: Set with expression`, could it not be a matter of updating that so that if set to a normal custom variable it evaluates the expression and sets the result to the variable, but if used on an expression variable it doesn't evaluate the expression and instead sets that as the value. This would of course need updating the descriptions on the action and documentation, but I see no reason why we should force a user defined variable to be read only.

Julusian:
> In the short-term for 4.1, could we expose local variables in buttons as the 'value' type for now? Similar to what 4.0 has for defining global variables. I think that in the short term, even simple local variables brings a lot of value to Companion.

The big problem with local variables is that probably less than 10% of the modules support them at the moment. I know it is a bit of a chicken and egg problem, as there isnt much reason for modules to do the work to be compatible, so I am trying to hold off until we can drop a bunch of reasons to do it.  
That said I would like to add the 'value' feedback type before companion 4.1, to make it easier for modules to support in the future

> Taking some of these ideas and expanding module functionality would be a good opportunity to dog food this stuff with developers (power users) before changing anything for general users. It would allow some of these concepts to be built and tested behind the scenes while allowing time to work on UX.

The challenge here is that until we have some kind of ui in companion for it, it wont be usable by anyone and in this case module devs wont have a way to test the feedbacks they are creating. And there is risk of confusing the less experienced module devs by offering them a bunch of stuff they can implement which does nothing.  
And as we don't have a way to control who sees what ui features (ie no license system or anything), if we put something in the ui then everyone can see it, including those normal users who are using a beta for some reason.

> More capabilities is great, but I'm just hoping people don't get tunnel visioned on making fancy stuff and ignoring the impact to users. 

Thats why this is still a draft and not merged ;)  
I just don't have the answers on how to solve this, so I do appreciate the input (even if I don't agree with some of it)


> Another option to consider is perhaps this shouldnt be custom-variables, but an entirely new concept with its own page. That way the simple variables can keep a simple in-table view, and the new page can have this split view.

This idea is growing on me.. 

Maybe something like this (pending better icon selection):
image
or keep it simpler:
image
Yeah its still some extra stuff thrown at users, but hopefully not too in your face. 


I suppose my concern of 'what if we change our minds and want to merge it later' is exaggerated, it wouldnt be hard to do some renaming and alias the old names for a while (like we are doing for `$(internal:custom_*)` to `$(custom:*)`). There is some risk of collisions, but probably unlikely if we prefix them too (eg `$(computed:*)` to `$(custom:computed_*)`)

lukescott:
> The big problem with local variables is that probably less than 10% of the modules support them at the moment. I know it is a bit of a chicken and egg problem, as there isnt much reason for modules to do the work to be compatible, so I am trying to hold off until we can drop a bunch of reasons to do it.
That said I would like to add the 'value' feedback type before companion 4.1, to make it easier for modules to support in the future

I was thinking less about module support, and more that this is a generally useful feature for users where global variables have to be used. Even with the internal actions/feedbacks.

In my personal use of companion, I have a case where I have 50+ variables, and they all use internal actions/feedbacks. I have that many because I need 2 variables per button. Even having the simple variables without module support would greatly increase the manageability of that use-case.

By support, I'm assuming you mean having the variables be selectable within a module action or feedback? Any particular reason a module has to opt-in for local variables or dictate what kind of variables can/can't be used? Doesn't companion have an opt-in already (useLocal) for using `$(this:*)` variables?

> The challenge here is that until we have some kind of ui in companion for it, it wont be usable by anyone and in this case module devs wont have a way to test the feedbacks they are creating. And there is risk of confusing the less experienced module devs by offering them a bunch of stuff they can implement which does nothing.
And as we don't have a way to control who sees what ui features (ie no license system or anything), if we put something in the ui then everyone can see it, including those normal users who are using a beta for some reason.

With this I was thinking more about #1187. Wouldn't the UI/UX work be pretty minimal? The current variables screen would more or less be what it is now, wouldn't it? Showing what the current values are.

This somewhat relates to the post you made above:

> No real argument there other than I am not good at design, so you won't find me trying to restyle everything.
If I do anything towards this, it will be replacing the ui toolkit we are using with something else, and that may or may not change spacing (but would help on the darkmode). And that will likely result in a bunch of other changes that people disagree with so I haven't yet decided it as a good idea (even though I really hate the current library)

I think perhaps we should be looking at separating these into chunks of work that do not require major UX/UI work from the ones that do. And strategically release the parts that bring user / module developer value.

I think there are just so many potentially good things we could do that the scope of work has grown too large to be manageable.

To expand on the chicken and the egg problem, I think that is from so many inner dependencies. You can'd to X without doing Y, and you can'd do Y without Z, and so on. I think that alone is a worthy problem to consider and solve. Asking why these are so closely linked, and what can be done to break them apart so they aren't so dependent on each other. -- We want to be in a place where you don't have to update modules at all.

Julusian:
> By support, I'm assuming you mean having the variables be selectable within a module action or feedback? Any particular reason a module has to opt-in for local variables or dictate what kind of variables can/can't be used? Doesn't companion have an opt-in already (useLocal) for using $(this:*) variables?

Its using the same opt-in, but modules also have to use the correct `parseVariablesInString` method for it to work. Generally a trivial change, but something that modules need to do. (The upcoming module-api does also handle this for them, but modules would need to update to that to benefit)

> Even having the simple variables without module support would greatly increase the manageability of that use-case.

My concern is that if we ship something that is useless to a large proportion of the users, they will get frustrated by this limitation and then not look at it again for ages if ever. Meaning that shipping it early could be a disservice to them. We often see users not following other changes that were made, and keep doing things the old way.

> With this I was thinking more about https://github.com/bitfocus/companion/issues/1187. Wouldn't the UI/UX work be pretty minimal? The current variables screen would more or less be what it is now, wouldn't it? Showing what the current values are.

Oh right. that is so out of my mind at this point.  
But yes, the intention was for this to be almost invisible to the user, with the design work needing to be done in the module-api. (in a backwards compatible fashion)

lukescott:
> Its using the same opt-in, but modules also have to use the correct parseVariablesInString method for it to work. Generally a trivial change, but something that modules need to do. (The upcoming module-api does also handle this for them, but modules would need to update to that to benefit)

Would it be possible to make the old version work without having to update the modules? And then mark the old way as deprecated. What could be done to get to a place where modules wouldn't have to be updated moving forward?

> My concern is that if we ship something that is useless to a large proportion of the users, they will get frustrated by this limitation and then not look at it again for ages if ever. Meaning that shipping it early could be a disservice to them. We often see users not following other changes that were made, and keep doing things the old way.

Maybe I'm in the minority here, but I would be happy to have the feature, even if I have to wait for it to be expanded later. The old way for me currently is not ideal :).

> Oh right. that is so out of my mind at this point.
But yes, the intention was for this to be almost invisible to the user, with the design work needing to be done in the module-api. (in a backwards compatible fashion)

Not sure what the lift would be, but to me this is just as valuable as computed variables. Could this be split into two chunks? First chunk the 'getValue' only version that lets modules define read-only properties. Second chunk adding 'setValue', which allows users to change the values.

I think overall, I'm trying to break these features into small deliverable chunks. Making each challenge smaller, easier to solve.

Julusian:
> Would it be possible to make the old version work without having to update the modules? And then mark the old way as deprecated. What could be done to get to a place where modules wouldn't have to be updated moving forward?

Nope, its a problem that the method calls the module makes don't have enough context to know what action/feedback they are for, so there is no way to know what local variables to use.  
Thinking about this, maybe it would be possible to solve for the old method by utilising asyncLocalStorage? But that would still require updating the module-api library, and the latest version has started on moving variable handling out of the module, so becomes rather redundant

> Could this be split into two chunks? First chunk the 'getValue' only version that lets modules define read-only properties. Second chunk adding 'setValue', which allows users to change the values.

Lets move this discussion to #1187, as it isnt really related to the pr here.

Julusian:
As it has been a month and noone is screaming no to my idea of splitting this out to a new 'computed variables' page and restoring the current custom-variables page as is, I shall do that.  
Shout quickly if you disagree with that plan.

Then in parallel I shall start looking into what can be done for UX of the existing custom-variables page.

Julusian:
That said, I am now a little worried about potential confusion here.

We have:
* Custom variables: user defined values, settable manually or in actions
* Computed variables: user defined expressions (which can use local variables)
* Local variables (on each control & computed variable): which can be either a user value (like a custom variable), Evaluate Expression (like a computed variable), or be from any other feedback

My worry (and the reason I originally did this as a rework of custom variables instead of its own thing) is that this difference between local-variables and custom/computed variables is a bit confusing, as they are similar in functionality but are exposed very differently and things are named differently.  
Maybe that is fine. Originally I had not intended to give computed variables their own local variables, but I think that is useful to have, and perhaps is enough to justify the differences.

I am wondering if this should be expression-variables instead of computed. I havent yet decided on the exact terminology.

Julusian:
ok, this is now its own page and concept. please voice any remaining concerns that need to be resolved before this can be merged, or I shall merge it tomorrow

Issue #223, opened 2018-09-08
Present the UI within the Electron app view


DrewLX:
**Describe the feature**
I would prefer it if the control UI didn't require a browser window, but rather displayed that within the Electron app window that already exists. This could be easily implemented - even as an iFrame if you wanted to be super lazy.

**Usecases**
Many of my show machines I wouldn't have a browser running, so it's a bit cleaner to my mind, and I know a few other Companion users that have made similar comments.

DrewLX:
I'm not talking about removing any features - I like the ability to do it, just not that its the only way to do it. Its a very small easy feature to implement and makes it a simpler and easier program to use.

willosof:
I’ll think about it :)

willosof:
Unfortuneately we need to prioritize other more important issues before we can look into this. But you're welcome to do the changes in the code and make a pull request :)

Julusian:
I did look at this and think 'this sounds pretty easy to do', but then realised there is more to it than I first thought.

Opening it as a separate window would be easy to do, but where should that be done from? I can't figure out where to put a button that wont cause confusion or make the launcher feel cluttered.

Do we need to be careful about navigation inside of this window? I expect it wont have a back button or other typical browser things.

In conclusion, I don't think this will work very well without causing confusion so I am thinking it shouldnt be done. If someone has a good plan on how it could be done, please provide some screenshots or a PR

PR #3534, opened 2025-07-06
feat: configure syslog from launcher

phillipivan:
Follow on from #3506 allowing configuration of syslog transport from the electron launcher. It is exposed and hidden by the same `cog` that reveals the `Devloper modules path`.

image

Occasionally after changing parameters an error is returned:
image
This doesn't seem to be contingent upon if the parameter is valid, and will occasionally trigger from simply enabling `syslog`. I do not understand why this is.

The logs prior to one such error:

```
launcher-enable-syslog: true
2025-07-06T09:30:01.386Z Application: trigger companion restart
2025-07-06T09:30:01.386Z Application: trigger companion restart
2025-07-06T09:30:01.714Z Application: Companion exited with code: 1
2025-07-06T09:30:01.714Z Application: Restart Count: 4
2025-07-06T09:31:16.649Z Application: Companion process stopped
2025-07-06T09:31:16.664Z Application: Companion process started
```

Other than that it seems to work.
image

@Julusian I'm sure you had something far more elegant in mind, UI wise. If you can point me towards some examples of what you had in mind I will do my best.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7883,21 +7883,21 @@ asn1@evs-broadcast/node-asn1:
   linkType: hard
 
 "brace-expansion@npm:^1.1.7":
-  version: 1.1.11
-  resolution: "brace-expansion@npm:1.1.11"
+  version: 1.1.12
+  resolution: "brace-expansion@npm:1.1.12"
   dependencies:
     balanced-match: "npm:^1.0.0"
     concat-map: "npm:0.0.1"
-  checksum: 10c0/695a56cd058096a7cb71fb09d9d6a7070113c7be516699ed361317aca2ec169f618e28b8af352e02ab4233fb54eb0168460a40dc320bab0034b36ab59aaad668
+  checksum: 10c0/975fecac2bb7758c062c20d0b3b6288c7cc895219ee25f0a64a9de662dbac981ff0b6e89909c3897c1f84fa353113a721923afdec5f8b2350255b097f12b1f73
   languageName: node
   linkType: hard
 
 "brace-expansion@npm:^2.0.1":
-  version: 2.0.1
-  resolution: "brace-expansion@npm:2.0.1"
+  version: 2.0.2
+  resolution: "brace-expansion@npm:2.0.2"
   dependencies:
     balanced-match: "npm:^1.0.0"
-  checksum: 10c0/b358f2fe060e2d7a87aa015979ecea07f3c37d4018f8d6deb5bd4c229ad3a0384fe6029bb76cd8be63c81e516ee52d1a0673edbe2023d53a5191732ae3c3e49f
+  checksum: 10c0/6d117a4c793488af86b83172deb6af143e94c17bc53b0b3cec259733923b4ca84679d506ac261f4ba3c7ed37c46018e2ff442f9ce453af8643ecd64f4a54e6cf
   languageName: node
   linkType: hard
 
@@ -9267,9 +9267,9 @@ asn1@evs-broadcast/node-asn1:
   linkType: hard
 
 "devalue@npm:^5.1.1":
-  version: 5.1.1
-  resolution: "devalue@npm:5.1.1"
-  checksum: 10c0/f6717a856fd54216959abd341cb189e47a9b37d72d8419e055ae77567ff4ed0fb683b1ffb6a71067f645adae5991bffabe6468a3e2385937bff49273e71c1f51
+  version: 5.3.2
+  resolution: "devalue@npm:5.3.2"
+  checksum: 10c0/2dab403779233224285afe4b30eaded038df10cb89b8f2c1e41dd855a8e6b634aa24175b87f64df665204bb9a6a6e7758d172682719b9c5cf3cef336ff9fa507
   languageName: node
   linkType: hard
 


Julusian:
I think this needs to be done in a settings modal/window.
With the amount of fields this adds, it makes this window rather messy, especially when only a tiny fraction of users will need to interact with these fields.

Unfortunately I don't have any examples to point to for this, as this current ui of the launcher is incredibly minimal and basic. It might be time to make this ui be its own vite project, or maybe the settings window could be done as its own html file inside the same webui bundle? (the challenge then would be serving that ui when the companion http isnt running)
It would make sense to use the same ui style as the main ui (or it could be styled on its own.

If you don't feel up to figuring out that workflow, I can deal with this (ideally the module path would move there too) then you can add the syslog stuff there once the structure is in place

dnmeid:
Wow, Companion v36! Future is calling.

What do you think about replacing the cog that reveals all options by an accordion?
There could be one section for the developer modules and one for the syslog.
This would also be good for another 2-4 sections before we need to rethink the overall design of the launcher.

Julusian:
> What do you think about replacing the cog that reveals all options by an accordion?

I suspect that will be tricky to make nice. I remember when first making this dual height it was a bit fiddly to get that to work correctly on every os. When the cog is clicked that triggers a size check and explicit resize of the window. But it would be fine if done without animations.  
But if making it an accordian is not too painful to do then I am open to that

phillipivan:
> If you don't feel up to figuring out that workflow, I can deal with this (ideally the module path would move there too) then you can add the syslog stuff there once the structure is in place

Pragmatically this is going to produce the fastest and best result, whether a modal is used or an accordion. This is really my first stab at anything 'frontendy' so Im sure it will be full of misteps.

arikorn:
As long as you're talking about redesigning this, it might be nice to give that window the standard title bar and caption controls. (I just figured out, while typing this, how to minimize the launcher window -- silly how much  we depend on the standard cues, but that's life...)
BTW, in terms of layout, increasing width, rather than height of the window for this new stuff, might look neater -- although maybe not compatible with accordion?

arikorn:
Another issue: if you turn the cog "off" (i.e. hide the dev modules path), it disables the dev module. This behavior made sense in previous version where it would just revert to the non-dev version of the module. But in v4.0+ it results in an error instead, since you explicitly select the dev module in the Connections tab and it doesn't automatically revert to a non-dev module (regardless of Update Policy). See screenshot.

My preference would be to allow the dev module regardless of the cog state. I could probably fix this myself, but don't want to cause a code-conflict with this work. Let me know either way if one of you can do it, or I should raise a separate issue and fix it myself, or....? Thanks!

image

Julusian:
I am taking a bit of a look at this, while I still have some motivation for UX polishing. I havent really touched the settings side of this yet, but tweaking the other bits:
image


>  BTW, in terms of layout, increasing width, rather than height of the window for this new stuff, might look neater -- although maybe not compatible with accordion?

This might be an idea, a toggle to show/hide a right panel attached to this, for the advanced configuration stuff.  

But I worry that this will make this panel feel messy, and a bit jarring. I like the simplicity and sleek look of this window currently, so am still leaning towards a 'full' larger config window. I dont know if there are more things we will add to it over time, but there are things that probably should be defined in it.  
For example, if we want to start down the route of making companion actually 'secure' we should offer an option to disable executing shell scripts, that would only work if defined in this launcher.  
Maybe we should let the user change the loglevel that gets written to disk?
I've been wanting a way to let the user cleanup the dbs from old companion versions, which would be better to do here than inside companion  
Im sure there are more, so making this into more a more fleshed out 'environment settings' ui would make sense

arikorn:
That looks nice, @Julusian, and I assume it would address the "cog 'off' error" as well?

Perhaps you could put the "Companion v##.##.##" in the title bar (or just "Bitfocus Companion"?) -- or is it just the icons, not a real title bar?

Note - that your mock-up (?) doesn't show the "Run at login" checkbox -- though I cover it below, FWIW.

With your list of possible additions, I wonder if putting in a menu bar instead might not be a good option. Something like:
- File
    - Set Dev Folder...
    - Configure Syslog...
    - Clean up DBs...
    - Exit
- Options
    - Start minimized  [_note: added after my initial post_]
    - Run at login
    - Disable  Shell Scripts
    - Change Log Level...
- View
     - Show Dev Folder
- Help
     - Getting Started...
     - About...

Or somesuch ("File" and "Options" could be merged; "Help" is perhaps aspirational). Point being it would both be a minimal-looking change and would make it look more like a standard app-window.

 (I like the idea of showing the dev folder where it currently is in the main window, at least as an option...)

Might look like (with *Show Dev Folder* enabled; mock-up done in Photoshop)

image

Julusian:
> Perhaps you could put the "Companion v##.##.##" in the title bar (or just "Bitfocus Companion"?) -- or is it just the icons, not a real title bar?

correct, its not a real titlebar. I like the simple look of this window without the titlebar, so am trying to keep that urrently

> I assume it would address the "cog 'off' error" as well?

Yeah, that behaviour wont make sense once it is in a new window.

> Note - that your mock-up (?) doesn't show the "Run at login" checkbox -- though I cover it below, FWIW.

Electron doesnt support that on linux, so we dont show it on linux

> (I like the idea of showing the dev folder where it currently is in the main window, at least as an option...)

I don't think we should treat this one option as special. I was debating moving the interface and port selection to a settings window too, but think that is too far.

> With your list of possible additions, I wonder if putting in a menu bar instead might not be a good option. Something like:

I think a single large settings page will look nicer and will be easier.  
For reference, satellite used to use a few tray options to change ip and port, but each one of those windows is its own html page, so it ends up being more code and effort to do than having a single page (and thats ignoring the UX aspect)

arikorn:
Fair enough @Julusian. Needless to say, I think my mockup looks pretty sharp , but a preferences window would definitely look nice too.

FWIW, another advantage of the menu system is that you essentially get the UI for "free" in terms of coding -- create the menu from template, and almost all of the interactions are just standard menu check boxes or system dialogs.  And very easy to expand for new settings that may come up in the future.

Some additional comments:

1. Optionally displaying info. The following may be worth considering, though it's mostly me "thinking out loud."
> > (I like the idea of showing the dev folder where it currently is in the main window, at least as an option...)
> 
> I don't think we should treat this one option as special.

I didn't mean it to be exclusive. The syslog information could be summarized in one or two lines, and made optionally visible too (see mockup below). Maybe other things too. (With the menu system, these are the only items it would need; with the settings window, possibly others too?). I definitely wouldn't recommend it as the default view, but for people who want it up-front it could be a nice touch. 

And of course the text could be made active so that clicking on them takes the user to the correct settings-window tab or system dialog.

Again, just to be clear: consider this mostly "thinking out loud," in case any of it sounds interesting...

2. Not that I use mac, but for completeness...

> correct, its not a real titlebar. I like the simple look of this window without the titlebar, so am trying to keep that currently

Does it work correctly with Mac's "traffic light buttons? (I.e. both the icons and the position on the left instead of right)

3. Another useful feature would be to raise the Companion GUI which, at least for people like me, is often buried in the large pile of tabs and browser windows I have open. This could be done by adding a "Window" menu in my scheme, or else by changing the "Launch GUI" button to "Show GUI" (or some other verb) after launching the GUI.

4. Yet another feature: it would be great if  "Launch" could open the Companion GUI in its own window rather than a tab in the last-visited browser window.  Something like HTML's `target="_blank"` or `--new-window` or equivalent javascript?

Thanks, as always, for your time!

mockup related to first point:
image

arikorn:
(Apologies for the multiple posts.) Here's my latest mockup of the launch window with the new developer info (on the left) just for comparison with the current launch screen, and also aligned so you can see what it would look like with the standard top part instead of my title bar & menu bar setup. It's surprisingly compact (same width; a bit shorter) and it could actually be used to implement all of @phillipivan's functionality, but the main point is to illustrate that the information can fit neatly on the screen, regardless of whether it is set here or in a separate screen.  On the flip side, if this window _is_ used to set all the parameters, the menu bar would probably be unnecessary for now. What do you guys think? Pretty sure I could even do it if there was interest...

On a side-note, now that I looked a little closer, 
1. I would not even expose the TCP vs. UDP choice -- I know there are cases where UDP is preferred but don't believe it would be here -- perhaps that's just my prejudice against UDP?
2. I'm not sure what "Companion Hostname" is; it may need a clearer label?

----

image

phillipivan:
>     1. I would not even expose the TCP vs. UDP choice -- I know there are cases where UDP is preferred but don't believe it would be here -- perhaps that's just my prejudice against UDP?
> 

UDP is still common for syslog. I hid a lot of possible settings, but I think that one is worth exposing. 


>     2. I'm not sure what "Companion Hostname" is; it may need a clearer label?


The reported host name of the Companion instance in the syslog message. In my screen shot above it would have been set to `Phillip's Laptop Companion`. Since we run multiple Companion instances in the same facility it is important to me that the messages from each can be easily distinguished. Open to a different property name; or just clarifying in the docs.

arikorn:
> The reported host name of the Companion instance in the syslog message. In my screen shot above it would have been set to Phillip's Laptop Companion. Since we run multiple Companion instances in the same facility it is important to me that the messages from each can be easily distinguished. Open to a different property name; or just clarifying in the docs.

Ah, that makes sense, thanks @phillipivan. Maybe the example value was what confused me.  Maybe just call it "Local Hostname" and have it default to `os.hostname()`?

Another question: could you say if/what this solution captures that wouldn't make it into the GUI "Log" page?  (Put another way: is this solution strictly for remote monitoring/multiple instances or would I benefit even if just running it on my local computer for a single installation?)

phillipivan:
It's for centralised log collection. You can run it on `localhost`, and while I did for testing, I don't know why you would do so in production.

Julusian:
> >   correct, its not a real titlebar. I like the simple look of this window without the titlebar, so am trying to keep that currently
> 
> Does it work correctly with Mac's "traffic light buttons? (I.e. both the icons and the position on the left instead of right)

No, but trying the real buttons now and I am not sure we will be able to customise them to match current behaviour. 

On linux, this works, but there is a maximise button there that does nothing.
On macos, the minimise button minimises it to the dock (not the tray) and I dont appear to be able to change that. Hiding the window while/once it is minimising or attempting to abort the minimise does nothing.  
On linux it isnt possible to disable minimise.  Maybe these linux quirks are fine, as I dont expect many users to use this there.
On windows, the buttons appear to be as customisable as I want.

So if going this route, maybe we should embrace the flow that macos wants and is used elsewhere, to 'close' the window into the tray, and to make the only way to quit be the tray icon 

image
image

image

-----

> FWIW, another advantage of the menu system is that you essentially get the UI for "free" in terms of coding -- create the menu from template, and almost all of the interactions are just standard menu check boxes or system dialogs. And very easy to expand for new settings that may come up in the future.

nope, electron doesnt give us a way to show something like an input dialog, we have to make that html page ourselves.  
If we were rewriting this launcher in another language then maybe, but that is not something I am in a rush to do.  
And tbh, even if we could use native ui components for this stuff, I am not convinced that will be a nicer experience than doing this as a single html window.

>     Another useful feature would be to raise the Companion GUI which, at least for people like me, is often buried in the large pile of tabs and browser windows I have open. This could be done by adding a "Window" menu in my scheme, or else by changing the "Launch GUI" button to "Show GUI" (or some other verb) after launching the GUI.
>
>    Yet another feature: it would be great if "Launch" could open the Companion GUI in its own window rather than a tab in the last-visited browser window. Something like HTML's target="_blank" or --new-window or equivalent javascript?

I dont think we can do either of these. The [openExternalUrl](https://www.electronjs.org/docs/latest/api/shell#shellopenexternalurl-options) method doesnt appear to have an option to choose it being in a window.
And once it is open, I dont see how we could tell the browser to bring a window/tab into focus. Maybe it is possible by interacting with each browsers api, but that will be a lot of effort.  
The other option is to stop using a browser for this and to host this as a window of the application. But I am not sure that is a good idea. (There was an issue for this a while ago #223, but there wasnt much desire for it and I had concerns)

Julusian:
This is what I am thinking about, and have in a branch:
image

syslog can be added here with as many options as it wants, and more can come later as needed

arikorn:
1. The settings window looks just like I imagined it, @Julusian, and looks good! ...except for the color scheme. (unless it's just using your system default, in which case it's perfect.) Oh, and is it really necessary to "enable" developer modules? It seems to me that if someone specifies the folder, then they want it enabled and conversely if no folder was specified or the folder is empty (or doesn't contain a dev version of module x) then "dev" doesn't show up in the Module Version" pulldown in the GUI, anyway.

2. Regarding my tangential requests for launching... oh well! I had seen that `openExternalUrl` doesn't support new windows but was hoping there was some work-around. But let's set all that aside for now, since it's not terribly relevant to the current PR. (Apologies for pirating it!) The one part that does pique my interest is the idea of running the GUI in electron. Technicalities aside, your new "advanced settings" window, would be the perfect place to put such an option, i.e. run in standalone vs. run in browser.

     Incidentally, your comment that there wasn't much desire for running standalone made me chuckle: some people in the GoStream user group say  they won't use Companion because they don't like that it runs in a browser window. In other words, this repo may select against the audience who "really wants" Companion to run standalone... (Hopefully that convoluted logic makes some sense...) Anyway, if I have further thoughts maybe I'll pick it up in #223?

3. Regarding
> nope, electron doesnt give us a way to show something like an input dialog, 

My bad! But there is `electron-prompt`, which gives an input dialog, or better yet, [`custom-electron-prompt` ](https://github.com/Araxeus/custom-electron-prompt )which might actually do it all, but I agree that getting all of @phillipivan's info is a bit of the fly-in-the-ointment to my claim that menus are simpler. Though I do believe I could handle his input needs pretty neatly on the main window, i.e. using [the last mockup I showed above](https://github.com/bitfocus/companion/pull/3534#issuecomment-3153386119) (actually, I would have a pull-down or spinner for udp vs. tcp, then an input or two for the "host:port). OK, that's just me being academic again. But even with your settings window, if the info can be displayed as in my mockup, then the user could click on it to get to the right settings tab quickly.

4. And regarding

> > Does it work correctly with Mac's "traffic light buttons? (I.e. both the icons and the position on the left instead of right)
> 
> No, but trying the real buttons now and I am not sure we will be able to customise them to match current behaviour.
> 
It looks like you got it covered, thanks! I have no objection to keeping a non-working maximize button. (Or, conversely, actually allowing the user to increase the window's size, just for consistency's sake.) Likewise, leaving in a quit button is perfectly fine with me. And of course this whole tangent started with the idea of adding the standard "minimize" button, so that's really the essential part for me.

arikorn:
This discussion continues in PR #3593.

peternewman:
> > The reported host name of the Companion instance in the syslog message. In my screen shot above it would have been set to Phillip's Laptop Companion. Since we run multiple Companion instances in the same facility it is important to me that the messages from each can be easily distinguished. Open to a different property name; or just clarifying in the docs.
> 
> Ah, that makes sense, thanks @phillipivan. Maybe the example value was what confused me. Maybe just call it "Local Hostname" and have it default to `os.hostname()`?

Would instance name be better or something then? Surely this should be the app name, not the hostname (which is a known/defined/understood thing). If you're really letting people overwrite the hostname, then as @arikorn suggested it should at least start with the machine hostname, not some generic example (which will by definition be the same on multiple hosts). Although getting a bit deeper there, you almost want it to default to undefined and substitute that for the current hostname in case it changes/when copied to another machine.

phillipivan:
When #3593 is merged I will revise this and let the `Companion Hostname` default to `os.hostname()` when empty. FYI `app_name` field is already set to Companion [internally ](https://github.com/bitfocus/companion/blob/c4786a4c6a93d805604ac3c4e84d24f1e51f5658/companion/lib/Log/Controller.ts#L260)

PR #3535, opened 2025-07-06
feat: rework surfaces page

Julusian:
This reworks the surfaces page to be more like the other pages.  
Instead of using a modal for configuring a surface, this brings the design of  the page to be inline with elsewhere.  
The arrangement of data in the table has been reshuffled a bit to better fit the space.

Screenshot From 2025-07-06 16-03-04

As part of this, the sidepanel has been enabled even for offline surfaces which allows for configuring some properties which used to be unconfigurable when the surface is offline:

image


arikorn:
Nice @Julusian and thank you!  I particularly appreciate:

> As part of this, the sidepanel has been enabled even for offline surfaces which allows for configuring some properties which used to be unconfigurable when the surface is offline

which I was about to ~~request~~ suggest...

Issue #3340, opened 2025-03-16
Function display


dve2021:
### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Hello,

I just noticed that there are more actions possible by typing the actions directly than by going through the folder on the right, which allows you to browse the actions for each module.

Is this intentional?

![Image](https://github.com/user-attachments/assets/f9eb3358-a50f-46e3-9508-f264d1ebaf9d)
![Image](https://github.com/user-attachments/assets/726e7c52-dbf0-487c-9821-e92ee88fd6fd)

### Environment (please complete the following information)

```markdown
- OS:
- Browser:
- Companion Version:
```

dve2021:
Thank you in advance for your reply.

Julusian:
What version of companion are you using?

Can you give an example of an action that is missing? I'm not noticing anything myself

dnmeid:
Your first picture shows **actions** search field and your second picture shows **feedbacks** browser. Maybe that is your problem. actions ≠ feedbacks.

dve2021:
Hello Dorian and Julian,

Indeed, I must have made a mistake because I can't find the problem I had earlier.

I probably added a filter that I forgot to remove.

I apologize for this mistake.

However, I'd like to take this opportunity to point out another issue.

When reimporting a configuration backup on companion
the OSC Listener is disabled in the settings page.

Do you have the same problem?

I'm on version 3.5.2.

Thank you for your reply.

Julusian:
Yes, it is currently expected that any settings arent imported. https://github.com/bitfocus/companion/pull/2875 will resolve this when complete

arikorn:
@dve2021 wrote:
> However, I'd like to take this opportunity to point out another issue.
>When reimporting a configuration backup on companion the OSC Listener is disabled in the settings page.

This secondary issue is now resolved by PR #3557 - user has option to import without the "full reset"

Issue #3329, opened 2025-03-09
Include surface settings in Configuration Import/Export


heisjolly:
### Is this a feature relevant to companion itself, and not a module?

- [x] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the feature

This is something I find a bit frustrating - every time I export and import configuration (as I often edit it on other computers to re-import on the actual server) I have to change the surface pages and settings again. This breaks all my tallies, phone emulators I provide to people so they can perform duties wirelessly etc.

Can these settings please be included in the configuration backup?

Thank you!

### Usecases

Quality of life, as per description.

jswalden:
https://github.com/bitfocus/companion/pull/2875 is probably relevant to this.

davidjoshuaford:
I experienced this too, although I notice that the surface Row and Column count is maintained, but not the Offset.

arikorn:
It turns out that this issue is **_not_** related to saving config settings (#2875, #3538), but is in fact a bug in the current import/export code!
 If you look at human-readable versions of the export files, it turns out that surface settings are already being saved on export and read on import (current beta 4.1, but no doubt 4.0 stable and even earlier). The main problem is that page numbers/id are not being saved properly and/or rematched. So in YAML for example, page id's are not saved in the page definition but the `nanoid()` of the page is being used in the surface config for start/last page.

~~~yaml
pages:
  "1":
    name: Page 1
...
...

  streamdeck:<serialID>:
    groupConfig:
      name: Unnamed group
      last_page_id: rihBOgFWIpgpE8NQqBfLP
      startup_page_id: rihBOgFWIpgpE8NQqBfLP
      use_last_page: false
 ...
~~~

I started tracing where best to fix it but would be just as happy if one of the experts take it on. (If you want me to do it, just LMK/assign this to me.)

arikorn:
And here's a bit of irony (or what in tech jargon is called the source of a "reversion," I believe)...  whenever this deprecation happened is likely when this Issue became an issue. So now we'll have to convert to/from page numbers when exporting/importing, I suppose... (Using internal ID's isn't a bad thing, if it allows pages to be tracked when they are moved...) This change appears to have been added in v3.5 as part of the upgraded page handling.
*shared-lib/lib/model/Surface.ts*
~~~typescript
export interface SurfaceGroupConfig {
	name: string
	last_page_id: string
	startup_page_id: string
	use_last_page: boolean

[...]

Issue #3276, opened 2025-02-09
[BUG] The "Default Export File Name" setting is omitted in configuration exports (and is reset during importing)


jswalden:
### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

The setting to set the default config-export filename isn't saved in those same exports.

That's maybe annoying, but it's not the worst thing in the world on its own.

What makes it much worse is that importing an exported configuration _resets_ that default filename setting.  So any time you make a mistake and use your backup to revert, the next time you export you have to set the very thing that was supposed to be a default...which makes the setting itself kind of useless.

I might try to fix this.  OTOH last time I tried looking at the export/import code, to try fixing #2688, I got confused by seemingly snarly code in it, so I easily might not, too.

### Environment (please complete the following information)

```markdown

```

jswalden:
So there's plainly some vague idea of including userconfig details (settings) in exports in the code.  But it's just commented out -- as if it is believed to be undesirable to include any settings information (including this filename) -- or perhaps more to the point, all of it indiscriminately -- in exports.

https://github.com/bitfocus/companion/blob/4c58d1db77ad5e60eb8ad46b6e557d9dfdab4007/webui/src/ImportExport/Export.tsx#L26
https://github.com/bitfocus/companion/blob/4c58d1db77ad5e60eb8ad46b6e557d9dfdab4007/webui/src/ImportExport/Export.tsx#L77
https://github.com/bitfocus/companion/blob/4c58d1db77ad5e60eb8ad46b6e557d9dfdab4007/webui/src/ImportExport/Export.tsx#L173

It's intentional that [`UserConfigModel`](https://github.com/bitfocus/companion/blob/4c58d1db77ad5e60eb8ad46b6e557d9dfdab4007/shared-lib/lib/Model/UserConfigModel.ts) not be saved wholesale in exports or overwritten wholesale in imports.  I can easily imagine justification for some of it, for example the HTTPS cert stuff probably at least would demand some opt-in.  But I don't have the background to really understand the philosophy the code should embody: which settings would one reasonably want to export, which ones would one want to exclude, what the defaults should be, etc.  And without that background, and presuming the undesirability of just bulling ahead to update the export format for _exactly one setting_, this seems beyond my capacity given my still-meager understanding of the project aims here.

Julusian:
[...]

Issue #3146, opened 2024-11-15
Import / Export setting workflow refinement


AJolly:
### Is this a feature relevant to companion itself, and not a module?

- [x] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the feature

I would like to speed up the export import setting workflow:

1) Don't force the user to go through the startup configure wizard every time you import settings
2) Remember all settings as part of the export and import.  For example, surface detection, button grid size, remote services, password, etc. 

3) I'd love if we could move to using json to hold the settings, rather than the sqlite.db (I understand why this may not be desired)

### Usecases

I run companion on my laptop and desktop. I'm treating it as a faster replacement for stream deck software.
I would like to keep my settings synchronized between the two. 
I frequently  import and export my config between the two machines.  

This takes much longer than required:
1)  There should be no reason to force a user to go through the config wizard every time they import the configure, after all importing a configuration implies that they've already been using companion and shouldn't need to go through the entire process.
2)  there is no reason the import export should not include surfaces and  all settings.   this is especially confusing considering import/exporting the config specifically asks you if you want to export surfaces.

3) I'd love json  just because it'd be easier to synchronize between machines, (synchronization software is much better at dealing with json to resolve conflicts).

Julusian:
1) & 2) There is a draft PR that should resolve most/all of this part https://github.com/bitfocus/companion/pull/2875

3) We have only changed the db to be sqlite (https://github.com/bitfocus/companion/pull/3045), not yet in a stable release. This was done partly in a new effort to mitigate corruption reported by some windows users, also because some peoples configs are becoming longer than nodejs can store in a string making their configs unreadable, and to minimise the cost of making config changes, to minimise the amount of data we write to disk.

arikorn:
Note: this feature request is mostly resolved with PR #3557, i.e., the user can choose a partial reset, which will keep the current Settings, and not require going through the settings wizard.

PR #3538, opened 2025-07-07
feat: add userconfig to import-export (updating PR #2875 to v4)

arikorn:
This is a cherry-picked version of @krocheck's PR #2875, with the code upgraded to v4

The original PR is old enough that SmartGit (and maybe git?) made it sound too scary to try and check it out, so instead I cherry-picked the two commits in that PR, resolved all conflicts, and then fixed the remaining compilation errors. This remains Keith's PR.

The code appears to work as expected, although it revealed a bug that is unrelated to saving "settings" -- surface import/export is buggy (essentially, issue  #3329) . (_The bug is now fixed/merged in PR #3542_)

This PR does resolve issues such as not having to re-enable settings that are disabled by default (#3146,  #3340 and my own experience with Contour Shuttle)

~~_**Note**_: I left *check.svg* (and *indeterminate.svg*) in *public/img* but am not entirely clear whether it should be there or in *src/scss/img*. Moving it resulted in a Sass warning that it couldn't resolve the file locations of the svg files.~~ [Edit: moved these files to *src/scss/img* in [27b45a9](https://github.com/bitfocus/companion/pull/3538/commits/27b45a9672f16418d938b2bd51646d9a7b3fc698), below, and converted to data urls]

This PR addresses some of the issues mentioned in: #3146, #3276,  #3340 (but not #3329, as mentioned above)

One last comment: this PR/issue has me wondering if another option would be not to do a full-reset when importing settings. (_implemented as a user-option; merged in PR #3557_)


arikorn:
Et voilà, whatever was bothering Github on the original PR is gone now!

arikorn:
A comment and a question:
1. Although Keith (@krocheck) suggested investigating a more granular approach for saving settings, please consider merging his functionality (in this PR) without the granularity. I suspect that the use-cases for more granularity are rare -- generally one wants to save everything so it all can be restored properly. Let someone raise an issue if they need to save only the PIN code but not the surface support, for example! 
(Also, some of the sub-categories can be confusing to the end-user since names like "buttons" and "surface support" can be  ambiguous, as seen in issue #3329, where "surface settings" is confused with the settings addressed by this PR.)
2. What does `"setup_wizard": 34,` mean? Is that correct?

arikorn:
Updated original comment to reflect my latest commit: moved the two svg files to src/scss/img in [27b45a9](https://github.com/bitfocus/companion/pull/3538/commits/27b45a9672f16418d938b2bd51646d9a7b3fc698), and converted to data urls. (They worked as regular URLs too, but I'm assuming it breaks in the dist or something?)

bryce-seifert:
> What does "setup_wizard": 34, mean? Is that correct?


I think this stems back to here https://github.com/bitfocus/companion/pull/1945#issuecomment-1055489345
Looks like it's the Companion version number as just a two digit number to track when new things are added to the wizard / it was last run. You can see the versions here too, with 34 being the most recent:
https://github.com/bitfocus/companion/blob/main/webui/src/Wizard/Constants.ts#L5

arikorn:
Thank you @bryce-seifert! That makes sense and also means it takes care of preventing the wizard from running when importing an exported file with that value in it.

arikorn:
I see that Keith (@krocheck) closed PR #2875 -- the "parent" of this PR -- a couple of days ago.

I take this as an endorsement from him on this PR -- thank you Keith! (And if I'm over-interpreting, please correct me.)

As such I am very eager to see this PR merged and/or reviewed soon -- before it goes stale again, -- since it seems to be a much-needed improvement to the import/export process. Currently, for example, a user who has enabled scanning for a non-default-enabled surface such as the Contour has to re-enable it and restart Companion every time they import a full config -- something that is both time-consuming and error-prone (i.e. forgetting to re-enable and restart and then wondering why the Contour isn't showing up as connected in the Surfaces page...).  On a more idealistic level, this PR removes a major inconsistency in the "full" import/export process, i.e. that it does a full reset but doesn't save/restore all the things it is resetting. (So in my mind, this PR is fixing a bug, FWIW.)

I have tested its basic functionality and found no obvious faults. (And see also my [comment, above](https://github.com/bitfocus/companion/pull/3538#issuecomment-3045590318).)

Many thanks!

krocheck:
Hah.   Not an endorsement.  You don't see the commits not sent to add the selection of settings groups.  Now that there are settings groups as menu navigation, those should be selectable as both export and import.  Plus, as part of surfaces we were including the surfaces table.  The indeterminate checkbox was added explicitly because this implementation required the "Settings" checkbox to be indeterminate because sub checkboxes needed to determine that main settings checkbox state.

arikorn:
Thank for the reply @krocheck.  Two questions:
1. what does this sentence mean: "Plus, as part of surfaces we were including the surfaces table."?
2. what is your timeline for adding selection of settings groups? i.e. days? weeks? months? years?

I am "campaigning" for the current result to be released ASAP and then you can add subcategories at your leisure. The work that you've already pushed in this PR is very important and significant on its own!!!  (And as the old saying goes... "Don't let perfect be the enemy of the good" -- in this case of course you can have it both ways: merge the "good" now then push the "perfect" later.)

companion/lib/ImportExport/Controller.ts
```diff
@@ -371,6 +372,11 @@ export class ImportExportController {
 				// Destroy old stuff
 				await this.#reset(undefined, !config || config.buttons)
 
+				// import userconfig
+				if (!config || config.userconfig) {
+					this.#userConfigController.setKeys(data.userconfig || {})
```
Julusian: I suspect we should be excluding the `admin_lockout` and `admin_password` fields here, to make it harder to accidentally lock yourself out when importing a config

arikorn:
Changed PR title for clarity and consistency (now that #2875 is closed)

arikorn:
For the sake of clear communication:
1. This is Keith's (@krocheck) PR. My only contributions have been to modernize the code and fix problems created by my other recent PRs. Especially now that Keith has indicated that he's actively working on this PR, I have no intention of contributing further code to this PR, i.e. doing anything that may interfere with what Keith is doing. (Of course, if another PR of mine causes problems with this one, I will make amends!)
2. I retract any of my comments encouraging swift action: The sense of urgency that I expressed in some earlier comments has been addressed by the recent addition of the option for partial reset upon import
3. @krocheck, I'm still concerned with, and hoping you can clarify your comment "Plus, as part of surfaces we were including the surfaces table." [mentioned above](https://github.com/bitfocus/companion/pull/3538/#issuecomment-3067588557). Surfaces should not be affected by this PR, other than the global setting enabling/disabling scanning for them. (Plus I simply don't understand what the sentence is saying: the current behavior of surface export/import appears to be correct and without "excess" -- and all parts work now that Julian and I fixed it a couple of days ago. Conversely, this PR as it currently stands on GitHub, correctly does not appear to do anything with surface configs, if that's what  you mean by "surfaces table".)

companion/lib/ImportExport/Controller.ts
arikorn: This makes sense, yet the opposite is true too: if a user loses their password every time they import a full config, it will be very frustrating and confusing.  Perhaps add a dialog on import if the import file's password settings are different from the current config? Something like: 
~~~
      This config file specifies an admin password that differs from your current settings.
      Do you want to: Clear/Replace/Keep your current password settings? 
~~~

arikorn:
The more I think about this PR, the less comfortable I am in having it being attributed to me. TBH, I created this PR on the assumption that @krocheck had abandoned PR #2875, considering how long it had been dormant. Keith mentioned above that he is still working on it, so he should retain the primary "ownership" and "credit" for this contribution. I'm therefore closing this PR but not deleting anything, so Keith can merge the updated code back into his original PR, if he so desires. Apologies for the inconvenience and misunderstanding!

PR #3542, opened 2025-07-10
Fix: Include surface settings in Configuration Import/Export #3329

arikorn:
Fix: Include surface settings in Configuration Import/Export #3329

Although not currently marked as such, issue #3329 is actually a bug report: The bug was introduced in commit [3af86b4](https://github.com/bitfocus/companion/commit/3af86b4) when Companion v3.5 changed the internal representation of the surface `startup_page_id` and `last_page_id`,  See my three comments starting here: https://github.com/bitfocus/companion/issues/3329#issuecomment-3043561606

  This PR includes two commits to fix two related but independent bugs

1. The first commit converts `startup_page_id` and `last_page_id` to/from the export/import page format (i.e. to/from page#/nanoid) when saving/loading the files, which is the bug mentioned above.

2. The second commit fixes a bug revealed by the previous commit:
When importing a companionconfig while a surface is attached, the surface's page settings are not updated. As a result, the surface's current page/start page are reset to the default (i.e., page 1, start on last page enabled). 
_[edit: see [comment, below](https://github.com/bitfocus/companion/pull/3542#issuecomment-3064457590) for a more complete description.]_

on branches fix_export_import_surface_config, personal/fix_export_import_surface_config


arikorn:
Just confirming again that this is a bug fix and not a new feature: saving startup page/current page worked correctly in Companion 3.4.3+7347-stable-23764ac2 (I just happened to be cleaning my storage and ran across it!).  

The second problem solved here appears to have been a bug in that version as well. (i.e. When importing a companionconfig while a surface is attached)

A more complete description of the second bug solved here:  When importing a companionconfig while a surface is attached, the surface's page settings are not updated. As a result, the surface's current page/start page are reset to the default (i.e., page 1, start on last page enabled).  However the imported settings _do_ take effect on the next restart. The one visible effect was that import reset the surface to page 1 instead of "current page" (i.e. `last_page_id`). Oddly, if one immediately exited, then the "current page" did revert to what was in the import file, and not to page 1, which was the actual current page upon exit.  This PR resolves the problems.

shared-lib/lib/Model/ExportModel.ts
```diff
@@ -19,8 +19,8 @@ export interface ExportFullv6 extends ExportBase<'full'> {
 	customVariablesCollections?: CustomVariableCollection[] // Added in v4.1
 	instances?: ExportInstancesv6
 	connectionCollections?: ConnectionCollection[] // Added in v4.1
-	surfaces?: unknown
-	surfaceGroups?: unknown
+	surfaces?: unknown // Record<number, SurfaceConfig>
```
Julusian, line 22: It would be good to set these to be the types in the comments instead of being `unknown`. 

Although I am leaning towards these using new 'Export*' types, to accomodate some of the other changes

companion/lib/ImportExport/Export.ts
```diff
@@ -492,8 +493,25 @@ export class ExportController {
...
 		if (!config || !isFalsey(config.surfaces)) {
-			exp.surfaces = this.#surfacesController.exportAll()
-			exp.surfaceGroups = this.#surfacesController.exportAllGroups()
+			const surfaces = this.#surfacesController.exportAll()
+			const surfaceGroups = this.#surfacesController.exportAllGroups()
+			const findPage = (id: string) => this.#pagesController.getPageNumber(id)
+			// Convert internal page refs to "export-record id's", i.e. page numbers.
+			// Note that `page`, which is a holdover from previous times, is already recorded as a number...
+			const setExportPageId = (groupConfig: SurfaceGroupConfig) => {
+				groupConfig.last_page_id = String(findPage(groupConfig.last_page_id) ?? '1')
```
Julusian: I am not keen on these on using the `last_page_id`, as it doesn't contain an id but a number. 

I think it would be better for the export logic to transform the SurfaceGroupConfig into an ExportSurfaceGroupConfig, which would not have these `_id` fields, and would have use some better named (the old naming?) for these properties.

As a bonus of using the old naming, that should fix the import of these page numbers from 3.4 and earlier, (although this isnt an important case to consider, just a nice one to do if easy)

companion/lib/ImportExport/Export.ts
```diff
@@ -438,10 +439,10 @@ export class ExportController {
 		const referencedConnectionIds = new Set<string>()
 		const referencedConnectionLabels = new Set<string>()
 
+		const pageInfos = this.#pagesController.getAll() // needed also for surfaces.
```
Julusian:
```suggestion
		const pageInfos = this.#pagesController.getAll()
```

I dont think this comment adds anything other than potential confusion

companion/lib/Surface/Controller.ts
```diff
@@ -1227,11 +1227,11 @@ export class SurfaceController extends EventEmitter<SurfaceControllerEvents> {
 		}
 	}
 
-	exportAll(): any {
+	exportAll(): Record<number, SurfaceConfig> {
```
Julusian, line 1339: nice! I don't know why these weren't using real types already..

companion/lib/ImportExport/Controller.ts
```diff
@@ -413,6 +414,22 @@ export class ImportExportController {
 				}
 
 				if (!config || config.surfaces) {
+					const surfaces = data.surfaces as Record<number, SurfaceConfig>
+					const surfaceGroups = data.surfaceGroups as Record<number, SurfaceGroupConfig>
+					const getPageId = (val: string) => this.#pagesController.getPageId(Number(val))
+					const fixPageId = (groupConfig: SurfaceGroupConfig) => {
+						groupConfig.last_page_id = getPageId(groupConfig.last_page_id) ?? ''
```
Julusian: Instead of '', a better default would be to use page 1 (the result of `getFirstPageId()` on the pagesController

companion/lib/Surface/Controller.ts
```diff
@@ -1266,6 +1266,10 @@ export class SurfaceController extends EventEmitter<SurfaceControllerEvents> {
 		for (const [surfaceId, surfaceConfig] of Object.entries(surfaces)) {
 			const surface = this.#getSurfaceHandlerForId(surfaceId, true)
 			if (surface) {
+				// it appears that #surfaceHandlers and #surfaceGroups have independent copies of `groupConfig`...
+				//  and the one in #surfaceGroups is the one that controls the surface's ..page.. values.
+				const group = this.#surfaceGroups.get(surfaceId)
```
Julusian: should this be up here or below the `this.#attachSurfaceToGroup(surface)`, because that call can change whether this will find a group

companion/lib/Surface/Controller.ts
```diff
@@ -1278,6 +1282,16 @@ export class SurfaceController extends EventEmitter<SurfaceControllerEvents> {
 					surface.setGroupId(newGroupId)
 					this.#attachSurfaceToGroup(surface)
 				}
+
+				// now copy the surfaceGroup into #surfaceGroups (does it matter that we don't copy `name`?)
+				// it appears that `surfaceHandlers` is empty if a surface is in a user-specified group
+				//  (note: I tried moving `&& group.surfaceHandlers.length > 0` to `group.#isAutoGroup` in Group.ts,
+				//  but it resulted in bogus groups being created when a device was attached -- and these groups only show up on next restart or on export.)
+				if (group && group.surfaceHandlers.length > 0) {
+					group.setGroupConfigValue('use_last_page', surfaceConfig.groupConfig.use_last_page)
```
Julusian: maybe this should be using the same logic as the `// Sync config` block near the start of this method?

Julusian (changes requested):

companion/lib/ImportExport/Controller.ts
arikorn: Changed.

companion/lib/ImportExport/Export.ts
arikorn: Removed and reverted the entire line to its original location. (this was left over from an intermediate revision)

companion/lib/ImportExport/Export.ts
arikorn: Perhaps I can convince you otherwise?  The abbreviation "ID" stands for "identifier. The identifier is typically the "key" in a map or object. For pages in the export file the ID _is_ the page number. For example:
~~~yaml
pages:
  "1":
    name: Page 1
    controls:
~~~
So, even on idealistic grounds, I'd have to say that using page number in the export file, and calling it an ID is correct.

But also on practical grounds, changing the name would disrupt importing the most commonly expected export file format (i.e. those saved recently). And regarding backwards compatibility, there is already code for transforming `last_page` and `startup_page` to the new names in *companion/lib/Surface/Groups.ts*, so keeping *page_id nomenclature would be less disruptive and would still preserve backwards compatibility.

companion/lib/Surface/Controller.ts
arikorn: Thank you!  My guess is it probably relates to the typedef... or that it just wasn't needed previously.

companion/lib/Surface/Controller.ts
arikorn: I suspect that if `this.#attachSurfaceToGroup(surface)` was called, then my fix wouldn't be necessary since it appears to recreate the whole object. Nevertheless, I agree that moving that  line is a good idea.
Done!

companion/lib/ImportExport/Export.ts
arikorn: OK, my confidence in *companion/lib/Surface/Groups.ts* was misplaced (it's not called before my code runs), but I still believe it best to keep property names as they are. 
I added code to handle importing files exported from v3.4, now that I have an example.

companion/lib/Surface/Controller.ts
arikorn: Not sure: `setGroupConfigValue` has a number of side-effects that I suspect we want to have. Especially when setting `last_page_id`

shared-lib/lib/Model/ExportModel.ts
arikorn: Setting to the types shown here results in several runtime errors in *companion/lib/Data/Upgrades* -- Maybe make changing the type a separate issue (or just leave it as-is)? (FWIW, I added the comment just as a reminder of what it becomes.)

I made my argument in favor of keeping things as they are earlier, and can't help thinking that creating parallel `Export*` types is not very DRY. The export format necessarily has to follow the internal format pretty closely anyway, so I would expect that adding export types will mostly create headaches with little benefit. Again, maybe at best it should be a separate issue?

arikorn:
@Julusian thank you for your thorough review! [edit: delete garbled text] Github presents your comments in a different order than in the "Files Changed" tab. Some of the comments will make more sense if viewed in that context.

shared-lib/lib/Model/ExportModel.ts
Julusian: > Setting to the types shown here results in several runtime errors in companion/lib/Data/Upgrades

thats fair, I can take a look at that at some point after merging this

> can't help thinking that creating parallel Export* types is not very DRY. The export format necessarily has to follow the internal format pretty closely anyway

My argument for having separate types is to minimise risk of error when the internal types change, and because of different stability requirements. Because we want to support old exports 'forever' (in theory exports from 1.3 will still work fine), so at points the internal and export formats do differ.

But tbh a lot of the other export types are not using isolated types, and I havent figured out a workflow to do any proper static/unit test analysis of the formats to ensure that changes don't break anything. So it is a bit all over the place at the moment (there is also the problem of half of it predating us using typescript, so is lacking types)

companion/lib/Surface/Controller.ts
arikorn: OK. I must have been looking at the code cross-eyed when I wrote the previous reply.
Code updated as per your request.

shared-lib/lib/Model/ExportModel.ts
arikorn: Thanks @Julusian. I'm still not convinced that there a benefit in adding new types, since we already have a mechanism in place for handling the outdated  properties by converting and removing them from the object during the import process.  Regardless, it sounds like we agree to defer any activity on these concerns until after merging this fix, so I'm going to mark the conversation as resolved for now. (Please re-open it if I misunderstood your comment!)

companion/lib/ImportExport/Export.ts
arikorn: @Julusian, I'm guessing that your other comment here (https://github.com/bitfocus/companion/pull/3542#discussion_r2203492998) would apply to this comment as well, so we can consider this one resolved too. But since I also added a bit of code to support v3.4, could you mark this as conversation as resolved, if you are satisfied for the purpose of this PR? Then should it be ready to merge?

Julusian: Sure, they are both identifiers, but they different kinds of identifiers. The converting to and from string indicates this and is itself a bit of a smell.

In fact, perhaps we ought to be preserving both the page number and page id in the exports? At some point we hope to redo a lot of the import logic to do more granular imports, in which case it might be possible during an import to match up the ids in the export with the ids of the pages even if their numbers have changed.

Considering this, perhaps the import logic should do:
1) If *_page_id is set, try and match it to a current page by id
2) If *_page is set, convert that number to a page id
3) Otherwise set to the first page.

arikorn: In somewhat reverse order:

- @Julusian, the problem with import-logic option 1 that page IDs are not saved to the pages section export file, so there is no way to match page_id in the surfaces section to a current page by id -- that's the very reason I wrote this PR!

    FWIW: I don't see an advantage to saving both page number and page id (implied by your first import-logic option) since the export file is static, so the relationship between page number and page id doesn't change. (Also, it seems likely to me that page_id would change on import, so even if they were saved to the file it would be difficult to match during import.)

- I totally support "At some point we hope to redo a lot of the import logic" but hope adding new features won't hold back this bug fix!

- As far as the smell, yes it should have been saved as a number, not a string but doing things this way preserves the current import/export formats. In retrospect, page_id should never have been added here...  I suppose at this point, if you really insisted we could reverse all of the logic and deprecate page_id, but this seems like way  more work than is justified by a "smell." I would suggest that when you're redoing the import logic we "deodorize" this as well. I'd be happy to help.

Julusian: > there is no way to match page_id in the surfaces section to a current page by id 

I am thinking of the future case where an export was taken, and some changes were made. Then the export is partially reimported; perhaps just some of the surfaces are imported. So there are some of the same pages, in a different order to before. So ideally the surface will stick to the page id rather than number in this scenario.  
I admit, this case is a bit odd/niche, but could be combined with imports of some of the pages

I have just added page ids to the exports, which will better allow for flows like this at some point.


> I totally support "At some point we hope to redo a lot of the import logic" but hope adding new features won't hold back this bug fix!

No Im not trying to avoid this fix, just make sure that this fix will align nicely with that. My worry is that putting a number in page_id means that in the future we will likely have to separate this out anyway, and then perform more steps to figure out how to interpret this value.

arikorn: @Julusian:  Sadly, as expected adding page ids to the exports didn't work because the page ids change on import. (See screenshot at end from the latest code in main). Just so you know: there is an oddity in the page id assignments in that page 1 (aka page 0) doesn't always get a new id, for reasons I couldn't figure out.

So options now include:
1. Change **_page import_** code to use imported ids (I'm not volunteering for this)
2. Don't save page ids (or ignore them) -- they really don't add any information, as I suggested earlier.
3. Keep page ids, but add code to convert them to the new ids upon import.

- If 1, then the only conversion bits submitted here still needed are if supporting v3.4 fields. (but see "regardless" below)
- If 2. then two more possibilities:
    -  2a. Change surface export algorithm to export `startup_page`/`last_page` (the deprecated version). The import algorithm will correctly fix those.
    - 2b. Just keep my code as it is and everything will work (needless to say: this is my recommendation).
- if 3. then the conversion code will be very similar to this PR, but will search `data.pages` to find the page number and then convert from page number back to the new page id.

Regardless, the other fix in this PR is still relevant too.

image

arikorn: Oh wait I forgot an option keep things as they are and do conversion of ids on import.  Edited above as option 3

Julusian: yeah 3 is what I am intending as a long term option (but preparing for it now), and 2a is what I am after short-term

>  Just so you know: there is an oddity in the page id assignments in that page 1 (aka page 0) doesn't always get a new id, for reasons I couldn't figure out.

I would guess that is a case of we aren't ever deleting that page, and instead clear it during the reset phase, not a big deal (and could explain why I didnt notice this lack of fixup originally)


> Don't save page ids (or ignore them) -- they really don't add any information, as I suggested earlier.

In isolation they don't but in the same way that we use them internally instead of numbers for some things, they can give some more stable references. We really need some actions that can either reference pages by id (or maybe some other friendly identifier). Once we have that (or anything else which wants to refer to ids) we will need the ids to be able to fix those up correctly (in the same way we do for connectionIds).

So yes today they may not be useful, but at some point they will be and having in the exports now means we might be able to import older configs better

arikorn: OK, implemented 2a and also fixed an aspect of the second bug that I had previously missed (emulators not being restored to `last_page`).

I actually tried doing option 3, but it was just too messy and raised some issues that made me wonder if we should ever go there.  In particular:
* Allowing both *page and *page_id, is ambiguous, especially if *page_id is required (as you can see, even in my 2a I had to make it optional for things to work smoothly).
* If only page_id is specified, what happens if the user only imports surfaces? All *page information will be lost!

Anyway, those are just musings for the future. The current state of this PR now implements 2a (save as *page only), which you indicated is your goal for it. If all is good, please resolve this conversation and hopefully it's ready to merge

arikorn:
I do not understand the strange conflict that needs to be resolved, but since it's not related to my code, and it's there even after I resolved the bizarre conflicts after merging with main, I'll leave it to you, @Julusian. Thanks!

companion/lib/ImportExport/Export.ts
Julusian: I've undone the part that makes page_id optional, as that suggests that our internal data structure supports that (because the types are shared). And I see no harm including both in the export (in the same way I added ids to the page exports, for future use).

Other than that I am happy now :)

Julusian (approved):

arikorn:
Thank you @Julusian !!!  This has certainly been a long slog even though it took less than a week.  I very much appreciate your guidance and patience throughout the process.

A final comment and a tangential question:
1. Thinking about using *page_id, I believe that the[ option 3](https://github.com/bitfocus/companion/pull/3542#discussion_r2205678545) is not viable, since if the surfaces are imported without the pages, the ids will be meaningless, i.e. the information is lost w/ no hope of recovery. And if you save both page# and page_id then it's simply redundant. Put in the positive sense: option 1 (don't update ids) seems like the way to go if you want to support importing page_id. Just my musings, feel free to ignore!
2. There were some pretty massive changes to the import/export code (and probably other parts too?) in the intervening week. It appears to be primarily under-the-hood, (introducing zod?) so I'm curious what was the motivation and, more generally, is there a place where this one in particular, and other such efforts are discussed or described?

Again, many thanks for your help here!

Julusian:
1) I think that 3 could be viable with some larger work to handle various scenarios. Having the unused ids there is small and harmless, and means that when/if we do that larger work then it can work with older exports too.

2) There was a rework of the library used for webui to backend communication (https://github.com/bitfocus/companion/pull/3550 & https://github.com/bitfocus/companion/pull/3259) which was a very sprawling change due to having all the api logic scattered among the codebase.  
    This code likely isnt in its final form, so far it is working an will be tidied further as I feel like it

PR #3543, opened 2025-07-11
Improve "visibility" of groups in the new surfaces page

arikorn:
(This idea could be applied to the current stable version too.)

Now that I've had a chance to play with the new Surfaces page, I believe that indentation would really help: Even though the "NO" column indicates groups, that column is easy to overlook. This PR indents members of a group under the group heading to make group membership clearer (especially for single-member groups).

image

### DIFF
--- a/webui/src/Surfaces/KnownSurfacesTable.tsx
+++ b/webui/src/Surfaces/KnownSurfacesTable.tsx
@@ -239,7 +239,7 @@ const SurfaceRow = observer(function SurfaceRow({
 			onClick={handleSurfaceClick}
 		>
 			<div className="grid-cell">{index !== undefined ? `#${index}` : ''}</div>
-			<div className="grid-cell" style={{ paddingLeft: index === undefined ? '2em' : 'default' }}>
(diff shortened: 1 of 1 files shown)


webui/src/Surfaces/KnownSurfacesTable.tsx
```diff
@@ -239,7 +239,7 @@ const SurfaceRow = observer(function SurfaceRow({
 			onClick={handleSurfaceClick}
 		>
 			<div className="grid-cell">{index !== undefined ? `#${index}` : ''}</div>
-			<div className="grid-cell">
+			<div className="grid-cell" style={{ paddingLeft: index === undefined ? '2em' : 'default' }}>
```
Julusian: It would be nicer for this to use the I think `ps-4` (or some other number) class for this.
Something like:
```
className={classNames('grid-cell', {
	'ps-4': index === undefined,
})}
```
(a few lines above is another example of this approach)

arikorn: Thank you for your guidance @Julusian !!
Done.

PR #3550, opened 2025-07-13
feat: replace socket.io with trpc

Julusian:
Follow up to https://github.com/bitfocus/companion/pull/3259 that completes the conversion

I was intending to do this more gradually, but this was a good low effort task to do when filling time.

This needs some more testing to make sure it behaves the same as before, I tried to test things as I went and they appeared to work ok, but there could well be things I missed.  
And some bits such as behaviour around connection loss were not easy to test when socket.io was still being used, so could do with some more attention now.


arikorn:
@Julusian, thank you for pointing me to these PRs -- it definitely helps in understanding what changed this past week.

Small question: when running `yarn install` it now shows

> @companion-app/webui@workspace:webui doesn't provide @trpc/server (p2c61c), requested by @trpc/client and other dependencies.

Is that something to be concerned about? Does it indicate that I need to rebuild webui? ...

Thanks!

Julusian:
> Is that something to be concerned about? Does it indicate that I need to rebuild webui? ...

No, trpc is complaining about things being spread across multiple packages in the monorepo

Issue #3348, opened 2025-03-20
[Request/Clarification] Ability to import components WITHOUT a full reset


ChrisSW58:
### Is this a feature relevant to companion itself, and not a module?

- [x] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the feature

If I wish to import buttons, triggers or custom variables to an existing configuration, it appears I can only do this with a complete reset of the config. The result is that I am left with only the components I imported. Is this as planned?

I would have expected a scenario where I could import one of those components and add/overwrite an existing config, rather than clear the existing config.

I'm not sure how this function was designed to be used. Willing to be educated!

### Usecases

When beta testing, it would be very helpful to be able to work on, say, custom variable documentation while triggers or buttons were still buggy, and then simply reimport the current buttons/triggers when the bugs were resolved.

arikorn:
Here's my thoughts: although there may be some interdependencies between the import categories, it seems to me that resetting all is certain to break them whereas leaving the unimported sections alone makes it likely that things won't break...

The fix turns out to be trivially simple because under the hood reset is already done by category.   I will push a PR shortly.

PR #3557, opened 2025-07-15
Ability to import components without a full reset

arikorn:
This commit/PR implements the feature described in #3348

Note that although there may be some interdependencies between the import categories, it seems to me that resetting all is certain to break them when importing only a subset of all the categories, whereas leaving the unimported sections alone makes it likely that things won't break...  (The same argument goes for userconfig: since it currently isn't saved, it shouldn't be reset.)

### DIFF
--- a/companion/lib/ImportExport/Controller.ts
+++ b/companion/lib/ImportExport/Controller.ts
@@ -470,7 +470,7 @@ export class ImportExportController {
 				}),
 
 			importFull: publicProcedure
-				.input(z.object({ choices: zodClientImportSelection, fullReset: z.boolean() }))
+				.input(z.object({ choices: zodClientImportSelection.nullable(), fullReset: z.boolean() }))
 				.mutation(async (opts) => {
 					// the following minimizes the changes below, by preserving the meaning of `input`
 					const ctx = opts.ctx
@@ -483,7 +483,7 @@ export class ImportExportController {
 
 						if (data.type !== 'full') throw new Error('Invalid import object')
 
-						const resetArg = fullReset ? undefined : { ...input, connections: true, userconfig: false }
+						const resetArg = fullReset || !input ? null : { ...input, connections: true, userconfig: false }
 
 						// Destroy old stuff
 						await this.#reset(resetArg, !input || input.buttons)
@@ -621,7 +621,7 @@ export class ImportExportController {
 		}
 	}
 
-	async #reset(config: ClientResetSelection | undefined, skipNavButtons = false): Promise<'ok'> {
+	async #reset(config: ClientResetSelection | null, skipNavButtons = false): Promise<'ok'> {
 		const controls = this.#controlsController.getAllControls()
 
 		if (!config || config.buttons) {
--- a/companion/lib/ImportExport/Controller.ts
+++ b/companion/lib/ImportExport/Controller.ts
@@ -477,7 +477,7 @@ export class ImportExportController {
 					if (data.type !== 'full') throw new Error('Invalid import object')
 
 					// Destroy old stuff
-					await this.#reset(undefined, !input || input.buttons)
+					await this.#reset({ ...input, connections: true, userconfig: false }, !input || input.buttons)
 
 					// import custom variables
 					if (!input || input.customVariables) {
--- a/companion/lib/ImportExport/Controller.ts
+++ b/companion/lib/ImportExport/Controller.ts
@@ -470,12 +470,10 @@ export class ImportExportController {
 				}),
 
 			importFull: publicProcedure
-				.input(z.object({ choices: zodClientImportSelection.nullable(), fullReset: z.boolean() }))
-				.mutation(async (opts) => {
+				.input(z.object({ config: zodClientImportSelection.nullable(), fullReset: z.boolean() }))
+				.mutation(async ({ input, ctx }) => {
 					// the following minimizes the changes below, by preserving the meaning of `input`
-					const ctx = opts.ctx
-					const input = opts.input.choices
-					const fullReset = opts.input.fullReset
+					const { config, fullReset } = input
 
 					return this.#checkOrRunImportTask('import', async () => {
 						const data = ctx.pendingImport?.object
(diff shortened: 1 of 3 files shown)


arikorn:
And of course if a user really wants to do a full reset before importing only one category, the "Reset Configuration" button is right there immediately below the Import button, so there's really no loss of functionality with this PR.

Julusian:
> And of course if a user really wants to do a full reset before importing only one category, the "Reset Configuration" button is right there immediately below the Import button, so there's really no loss of functionality with this PR.

There isn't but I worry that the change will cause confusion/frustration for those who are used to it performing a full reset.
I see this going one of 4 ways for users:
1) They don't realise it did a partial reset until later when they have started using the imported state
1) They realise immediately after importing that it did a partial reset, and have to do a reset and repeat the import
1) They read the text and realise they have to go back to perform the reset then restart the import process (an extra 8+ clicks)
1) They know to perform a full reset first, so do that first (even though this is 5+ clicks longer than the old import flow)

So I think it would be best to add this as another button at the bottom, so that there is:
1) Full reset and import
2) Partial reset and import

That gives users both options, and also avoids an unexpected behaviour change.

arikorn:
No doubt you're just suggesting this to force me to expand my skills (including the new tRPC code).  ;)

All done now. Here are the notes in the current commit plus additional questions:
Notes:
- the changes in *webui/.../import/Full.tsx* were copied to parallel *webui/.../import/Triggers.tsx*. However, using braces instead of quotation marks around `true` and `false` resulted in a CSS error (in the web browser console) that prevented it from working. Not sure why it works for *Triggers* but not here?
- In order to preserve `input` as used in the `importFull: publicProcedure` callback, I had to manually unpack the arg in `.mutation(async ()`
- the compiler doesn't complain about `input` possibly being undefined, so it's not clear to me if `if (!input || ...` is actually necessary.

Additional Questions:
- ~~tRPC seems to hide much of the code "traceability" from VS Code (while coding, not during runtime). Is there a plugin that helps with tRPC?~~ (Nevermind, it seems to be working now.)
- I've been having trouble attaching/debugging the React code, even with auto attach enabled. Do you have any advice?

arikorn:
One more note: there seem to be quite a few whitespace and formatting changes -- that's just what `yarn format` did. Sorry about that!

Julusian:
> the compiler doesn't complain about input possibly being undefined, so it's not clear to me if if (!input || ... is actually necessary.

good point, that it something this is supposed to support but I accidentally broken during the trpc conversion.
I've push a commit to this that fixes that

> One more note: there seem to be quite a few whitespace and formatting changes -- that's just what yarn format did. Sorry about that!

:shrug: that happens. github has a toggle to hide the whitespace changes when reviewing a pr so that all disappears when I look at it.

> In order to preserve input as used in the importFull: publicProcedure callback, I had to manually unpack the arg in .mutation(async ()

I wouldnt object to updating the lines below to use a new name, it only just changed to input instead of config for a similar reason

> the changes in webui/.../import/Full.tsx were copied to parallel webui/.../import/Triggers.tsx. However, using braces instead of quotation marks around true and false resulted in a CSS error (in the web browser console) that prevented it from working. Not sure why it works for Triggers but not here?

I would guess a naming thing, as the one in Full.tsx is not called `data-*`. so the browser is likely complaining about an unknown property. https://developer.mozilla.org/en-US/docs/Web/HTML/How_to/Use_data_attributes

> tRPC seems to hide much of the code "traceability" from VS Code (while coding, not during runtime). Is there a plugin that helps with tRPC? (Nevermind, it seems to be working now.)

Im glad you figured that out, it should just work. maybe it just needed a restart of vscode?

> I've been having trouble attaching/debugging the React code, even with auto attach enabled. Do you have any advice?

I do not, I rarely use a debugger here. (Largely because a lot of my projects involve some kind of connections that are timing sensitive which a debugger often breaks, so I have gotten used to not being able to use one)

arikorn:
Thank you @Julusian.  Just pushed the fix for the css name and reverted `input` to `config` on the server-side.

- Personally I would have preferred "choice" instead of "config" (especially since "config" also is used to mean "Settings), but there is way too much precedent in the code for using "config", so that's fine.
- Curious: what's the case in which "config" would be null?
- Sometimes when I do `yarn dev` (maybe the first time after `yarn install` possibly only after changing branches, running `yarn install` then `yarn dist:webui` or such?) I get:
~~~
file:///home/ari/Git-projects/companion/shared-lib/dist/Entity.js:1
import { FeedbackEntitySubType } from './Model/EntityModel.js';
         ^^^^^^^^^^^^^^^^^^^^^
SyntaxError: The requested module './Model/EntityModel.js' does not provide an export named 'FeedbackEntitySubType'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:182:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:266:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:644:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5)
~~~

the code still runs after that.

Julusian:
> Personally I would have preferred "choice" instead of "config" (especially since "config" also is used to mean "Settings), but there is way too much precedent in the code for using "config", so that's fine.

Changing it to that is fine with me, I dont have any attachment to any particular naming. Let me know if you want to do that, the pr looks ready to merge

> Curious: what's the case in which "config" would be null?

I think this was intended as an easy way to perform a full import without having to set all the boolean flags to true. I don't know if it should still exist, but its very little code so I don't see harm in keeping it

> Sometimes when I do yarn dev (maybe the first time after yarn install possibly only after changing branches, running yarn install then yarn dist:webui or such?) I get:

Yeah yarn dev does that, but it will recover. dist:webui should perform better now

arikorn:
@Julusian wrote:

> Changing it to that is fine with me, I dont have any attachment to any particular naming. Let me know if you want to do that, the pr looks ready to merge

I'm good with "config" so let's merge. It may, coincidentally, fix the failed checks with the other import/export PR!

arikorn:
Thanks @Julusian, and I forgot to add in the previous note: I appreciate the offer to rename config!

arikorn:
For completeness: this PR "mostly" resolves the issues mentioned in #3146 and #3276 as well as resolving the secondary issue in #3340  (which is already closed)

Julusian:
I've hidden this new partial import button in 4.1.2 as it is often resetting things that it isnt supposed to be.
I have tried to fix some up, but more bits keep revealing themselves.
Mostly this is happening when the import doesnt contain things, or prior to my first attempt to fix #3639 it was always resetting and reimporting the connections

PR #3567, opened 2025-07-21
fix: Edit-Button Load-PNG not loading updated files

arikorn:
(Note: I can post an issue if you prefer, or do so in the future -- this one just removes a minor though persistent annoyance so it didn't seem worth it.)

When the contents of a PNG file is updated, replacing the button image from the file of the same name doesn't always work, when using the PNG "folder" button:
    image

Specifically,
1. Load a PNG file using the PNG "folder" button.
2. Stay on the Buttons page
3. Modify the PNG file. (Switching apps doesn't matter, just as long as "Bitfocus Companion - Admin" remains on the same page)
4. Try to update the button-image  using the PNG "folder" button and selecting the updated file (same name as before).

Moving to a different buttons page doesn't help, though moving to a different URL or restarting will solve the problem.

This commit fixes the issue, so one can do rapid "protoyping" of the PNG image without leaving the "Buttons" page.

This PR works by erasing the form's `value` property. While in theory, one should test for when to erase the existing value, doing it all the time doesn't appear to harm anything in this case, even if the user cancels the file dialog.

### DIFF
--- a/webui/src/Components/PNGInputField.tsx
+++ b/webui/src/Components/PNGInputField.tsx
@@ -47,62 +47,56 @@ export function PNGInputField({ min, max, onSelect, onError }: PNGInputFieldProp
 		return canvas.toDataURL()
 	}
 
-	const onClick = useCallback(
-		(e: React.MouseEvent) => {
-			onError(null)
-			if (inputRef.current) {
-				const form = e.currentTarget.getElementsByClassName('form-control')[0] as HTMLInputElement
-				form.value = '' //files = null didn't work
-				inputRef.current.click()
+	const onClick = useCallback(() => {
+		const fileForm = inputRef.current // "file" type of input form
+		onError(null)
+		if (fileForm) {
+			fileForm.value = ''
+			fileForm.files = new DataTransfer().files // you can't create a FileList object directly (even though compiler doesn't complain)
+			fileForm.click()
+		}
+	}, [onError])
+	const onChange = useCallback(() => {
+		const newFiles = inputRef.current?.files
+		console.log('change', newFiles)
+
+		//check whether browser fully supports all File API
+		if (apiIsSupported) {
+			if (!newFiles || !newFiles.length || newFiles[0].type !== 'image/png') {
+				onError('Sorry. Only proper PNG files are supported.')
+				return
 			}
-		},
-		[onError]
-	)
-	const onChange = useCallback(
-		(e: React.ChangeEvent<HTMLInputElement>) => {
-			const newFiles = e.currentTarget.files
-			e.currentTarget.files = null // note: this doesn't do anything (files is unaffected)
-			console.log('change', newFiles)
 
-			//check whether browser fully supports all File API
-			if (apiIsSupported) {
-				if (!newFiles || !newFiles.length || newFiles[0].type !== 'image/png') {
-					onError('Sorry. Only proper PNG files are supported.')
-					return
-				}
+			Promise.resolve()
+				.then(async () => {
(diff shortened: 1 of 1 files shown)


webui/src/Components/PNGInputField.tsx
```diff
@@ -47,16 +47,21 @@ export function PNGInputField({ min, max, onSelect, onError }: PNGInputFieldProp
...
-		onError(null)
-		if (inputRef.current) {
-			inputRef.current.click()
-		}
-	}, [onError])
+	const onClick = useCallback(
+		(e: React.MouseEvent) => {
+			onError(null)
+			if (inputRef.current) {
+				const form = e.currentTarget.getElementsByClassName('form-control')[0] as HTMLInputElement
```
Julusian: I think form is pointing to the same thing as `inputRef.current`, so it would be nicer to just use that.

webui/src/Components/PNGInputField.tsx
```diff
@@ -47,16 +47,21 @@ export function PNGInputField({ min, max, onSelect, onError }: PNGInputFieldProp
...
+				inputRef.current.click()
+			}
+		},
+		[onError]
+	)
 	const onChange = useCallback(
 		(e: React.ChangeEvent<HTMLInputElement>) => {
 			const newFiles = e.currentTarget.files
-			e.currentTarget.files = null
+			e.currentTarget.files = null // note: this doesn't do anything (files is unaffected)
```
Julusian: This should either be fixed (I assume doing `= ''` instead? or the line should be removed, there is no point having code that does nothing

webui/src/Components/PNGInputField.tsx
arikorn: Interesting! I probably missed this fact because I was using `onChange`, below, as my template. Apparently the difference is that `inputRef.current` can be null -- not an issue here. I'll fix `onChange` to use `inputRef.current` as well since that would simplify it slightly. On the good news side, I figured out how to use the VS Code debugger on React code. It breaks the timing of course, but since my main goal these days is to examine the state at a particular place in the code, it's perfect for me. (Also, good news - Companion recovers gracefully from these timeouts.)

webui/src/Components/PNGInputField.tsx
arikorn: Yeah, it's a bit more complicated, but I can fix it. Do you have any idea what the intention of this line is?  I suspect it's to solve the same problem I'm solving in this PR. On that assumption, I'll move the code up into the `onClick` callback since that's the better place to do it (i.e. *before* React decides whether or not the value has changed). It just adds a tiny layer of safety to reset both values.

To keep things moving, I'll just submit changes to both callbacks, as discussed in this and your other comment, then you can suggest I change things back or just leave or undo it yourself, as you see fit.

webui/src/Components/PNGInputField.tsx
Julusian: well the onChange makes sense to be using e.currentTarget, as in onChange that is already the file input field.
but here in onClick it is searching based on css classes for it, and then uses inputRef to trigger a 'click' on the file input field, so it would make sense for the clear to use the inputRef too.

arikorn: Oh. The search for css class in `onClick` is because there are two items in `e.currentTarget` while in `onClick` and that was the only way I could find to access the one I wanted. Either way works. I'll revert `onChange` now.

arikorn:
OK, it should all be as you requested now. @Julusian .  Thank you, as always for your help and advice.

PR #3569, opened 2025-07-21
feat: Rework connection config layout #3559

### LINKED ISSUE #3559: Rework connection config layout
Since 4.0 we have rearranged the action/feedback options and a few other input fields to be in a fixed 2 column layout:

Image

But connection configs have not been touched and so are at this point inconsistent with the rest.  
I have avoided changing this, because in this one place we are giving control to modules on how to arrange these things, and I know that some modules have put some effort into the layout of things.

But for consistency, I would propose that we remove this ability from modules, and transition to a fixed 2 column layout instead.
Generally, modules are using a 6/6 split, occasionally an 8/4 split for host and port, and even rarer is something else.

I know that for some modules where fields are using isVisible, I have had to create empty static-text fields to show when hiding the real field, to preserve layout.

I expect that doing this will make these configs take up some more vertical height, but as they are usually 2 per row, it will probably only be 33% taller (2x input height instead of 1x input height + 1x label height).  
Additionally, this would allow the 3 companion owned fields at the top to also follow this arrangement, which will save some vertical space.

The one exception to this is the static-text field, I think that will want to stay as-is.

As part of this, maybe we should add a `note: string` field (other ideas on name are welcome) to these config field types, to be able to add a line of text below like:

Image

We already have something for tooltip, but an always visible line below could be useful too, especially in helping ensure names of fields are short

Julusian:
closes #3559 

I still need some input on whether this is something we want to do, but it was quick enough to do so having a screenshot might help.

Atem module (which is a basic 6/6 layout):

Screenshot From 2025-07-22 00-04-31

Screenshot From 2025-07-22 00-04-26

The 'detected module' static-text doesnt look great. I think that in general we want to keep static-text spanning the full width, as they are commonly long blocks of text.  
Which means some might end up odd like this. But that can be mitigated by adding a new `singleLine?: boolean` property to the field type (and maybe guessing this value if a width value is defined and not 12)

Maybe a 'lineAbove' property would be useful too for some modules, to be better able to breakup the list a bit (ie a `<hr />`)

### DIFF
--- a/companion/lib/Instance/ApiVersions.ts
+++ b/companion/lib/Instance/ApiVersions.ts
@@ -3,6 +3,7 @@ import semver, { type SemVer } from 'semver'
 const range1_2_0OrLater = new semver.Range('>=1.2.0-0', { includePrerelease: true })
 const range1_12_0OrLater = new semver.Range('>=1.12.0-0', { includePrerelease: true })
 const range1_13_0OrLater = new semver.Range('>=1.13.0-3', { includePrerelease: true })
+const range1_14_0OrLater = new semver.Range('>=1.14.0-0', { includePrerelease: true })
 
 export function doesModuleExpectLabelUpdates(apiVersion: SemVer | string): boolean {
 	return range1_2_0OrLater.test(apiVersion)
@@ -14,4 +15,8 @@ export function doesModuleSupportPermissionsModel(apiVersion: SemVer | string): 
 
 export function doesModuleUseSeparateUpgradeMethod(apiVersion: SemVer | string): boolean {
 	return range1_13_0OrLater.test(apiVersion)
+}
+
+export function doesModuleUseNewConfigLayout(apiVersion: SemVer | string): boolean {
+	return range1_14_0OrLater.test(apiVersion)
 }
--- a/companion/lib/Instance/Connection/ChildHandler.ts
+++ b/companion/lib/Instance/Connection/ChildHandler.ts
@@ -53,7 +53,11 @@ import {
 import type { ClientEntityDefinition } from '@companion-app/shared/Model/EntityDefinitionModel.js'
 import type { Complete } from '@companion-module/base/dist/util.js'
 import type { RespawnMonitor } from '@companion-app/shared/Respawn.js'
-import { doesModuleExpectLabelUpdates, doesModuleUseSeparateUpgradeMethod } from '../ApiVersions.js'
+import {
+	doesModuleExpectLabelUpdates,
+	doesModuleUseSeparateUpgradeMethod,
+	doesModuleUseNewConfigLayout,
+} from '../ApiVersions.js'
 import { ConnectionEntityManager } from './EntityManager.js'
 import type { ControlEntityInstance } from '../../Controls/Entities/EntityInstance.js'
 import { translateEntityInputFields } from '../ConfigFields.js'
@@ -95,6 +99,7 @@ export class ConnectionChildHandler implements ChildProcessHandlerBase {
 	#hasHttpHandler = false
 
 	hasRecordActionsHandler: boolean = false
+	usesNewConfigLayout: boolean = false
 
 	#expectsLabelUpdates: boolean = false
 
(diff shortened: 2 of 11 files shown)


Julusian:
VLC: 

Screenshot From 2025-07-22 00-10-50

Screenshot From 2025-07-22 00-10-46

Julusian:
google sheets
Screenshot From 2025-07-22 00-11-53

Screenshot From 2025-07-22 00-11-58

thedist:
In the Twitch module, I make use of 3 columns for a large set of checkboxes, I also have a long string of text on an input box that looks quite bad when forced into a column (this could be changed and written as a separate static text perhaps).

Current config
chrome_YM5aM10FQq

This branch
chrome_12VK8JL9B0

Due to this I'm somewhat of two minds, because in some modules it does look nice, but for modules with lots of checkboxes with minimalist titles it creates lots of wasted space (one of my gripes with the Companion UI is how much space stuff takes in so many different areas).

Julusian:
That reminds me, I need to make the save/cancel button be sticky to the bottom of the screen

That channels to monitor field doesnt look great either, and I think is a good argument for supporting this line of text: 
image

And yeah, in general this change does look to make things a bit taller, but my motivator for this change is for consistency, as pretty much everywhere else is now in this 2 column approach, so it would be nice for this to have the same style.   
The other part of my motivation for this is trying to make some things be a bit more responsive with varying screen widths. Various places have had some polishing, and some connections get a bit messed up in narrow screens due to the columns usually designed for 1080p screens

For those checkboxes, maybe this is a field type we are missing. Having a type for checkbox-grid would let companion decide on some breakpoints and do some more dynamic column counts. 

And I suppose I should add that if we really dont want to 'break' modules like this, this is behaviour that could vary depending on the version of the module-api. It would mean being inconsistent across modules for a while, but essentially makes it opt-in, so module devs have a chance to adjust before it gets applied to their module.

> one of my gripes with the Companion UI is how much space stuff takes in so many different areas

Im curious what other things this is (maybe write this somewhere else)

lukescott:
I feel like this sacrifices readability. I use this, for example, to make key bindings for arrow keys more readable:

image

Alternatively I could put them all on the same row.

Putting these on 2 rows, or even 4 rows is terrible looking and a waste of space.

Julusian:
I suppose my argument for this is simply the question: 
**Do we want to change this to be consistent UX with elsewhere, or accept that this is different and even a bit inconsistent within this panel?**

Anything else is either some new functionality that should be added to give modules some new ways to influence the layout (eg a checkbox grid), or is a reasonable loss in the name of consistency.  
Some of this could influence the decision, but either way it is a compromise.

jswalden:
I find the long list of checkboxes easier to understand when it's not in multiple columns, because it's much easier to skim a single alphabetical column to find something (or not find it).

(The up/down/left/right labeling is much clearer when it visually mimics the directions, yes.  But IMO it really more wants direct grid access than to kind of hackily create one.  In fairness this would be a much harder API to design, tho I can imagine something like `{ type: "grid"; width: number; height: number; fields: (SomeCompanionConfigField & { x: number; y: number })[] }`  as a wild sketch.)

My instincts tend sufficiently Pythonic that "special cases aren't special enough to break the rules" would be my guiding star.  My two cents -- and worth around that much.

Julusian:
After discussing this directly with a few people, we have reached an agreement on the following:

There is concern about 'breaking' the layout of modules, making things worse in the short term. To mitigate this, this new layout will only be enforced for modules running the 1.14 (or maybe a patch release of 1.13) module-api, and they will be able to opt-out of the new layout if they feel the need to. This will likely land in Companion 4.2   
At some point in the future, this will become the default layout for modules with older module-api. (probably either in companion 4.3 or 4.4) 
At some point in the future, the old layout will be removed entirely (at the earliest companion 4.4, or could be a few releases later)

For any module opting-out, we encourage them to share with us why they are doing so, as this opt out wont be available forever. We expect that we will need to add some more controls/options/something to allow for some configurability in the new system (eg the grid of checkboxes), the sooner modules tell us what they need/want, the sooner these can be made for them.

krocheck:
I like the proposed approach.

Julusian:
OK, this is done. It will only affect modules using api 1.14.0 or later, and there is an opt out (in the constructor `this.options.disableNewConfigLayout = true`)

The opt out will be removed when we hit api 2.0, and this new layout will probably be forced upon all modules at some point before then

Julusian:
Please open any feature requests for any new field types needed for sensible layouts. I'm considering removing support for the old layout in 5.1 or 5.2 as nothing has been raised.

PR #3575, opened 2025-07-25
Feature: create/delete custom-variable from API

YWatchman:
I was missing an endpoint to create/delete a custom-variable programmatically, this should add it.

### DIFF
--- a/companion/lib/Service/HttpApi.ts
+++ b/companion/lib/Service/HttpApi.ts
@@ -318,6 +318,9 @@ export class ServiceHttpApi {
 		// Add create endpoint for custom variables
 		this.#apiRouter.post('/custom-variable/:name/create', this.#customVariableCreate)
 
+		// Add delete endpoint for custom variables
+		this.#apiRouter.delete('/custom-variable/:name', this.#customVariableDelete)
+
 		// Module variables
 		this.#apiRouter.route('/variable/:label/:name/value').get(this.#moduleVariableGetValue)
 
@@ -644,6 +647,20 @@ export class ServiceHttpApi {
 		} else {
 			res.status(201).send('ok')
 		}
+	}
+
+	/**
+	 * Delete a custom variable
+	 */
+	#customVariableDelete = (req: express.Request, res: express.Response): void => {
+		const variableName = req.params.name
+		// Check if variable exists
+		if (!this.#serviceApi.getCustomVariableValue(variableName)) {
+			res.status(404).send('Not found')
+			return
+		}
+		this.#serviceApi.deleteCustomVariable(variableName)
+		res.status(204).send()
 	}
 
 	/**
--- a/companion/lib/Service/HttpApi.ts
+++ b/companion/lib/Service/HttpApi.ts
@@ -315,6 +315,9 @@ export class ServiceHttpApi {
 			.post(this.#customVariableSetValue)
 			.get(this.#customVariableGetValue)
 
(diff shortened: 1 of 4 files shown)


thedist:
Why the choice of using a `/create` endpoint over just `/custom-variable/:name` and using the request method to determine appropriate CRUD actions to be performed?

I know the API scheme that Companion uses is somewhat janky, but it just adds to that somewhere when the create endpoint requires a `/create` but deletion doesn't require a `/delete`. There should be consistency.

YWatchman:
There's no reason for it, I'll change it up.

Julusian:
I think this needs a discussion from us maintainers to decide if this is something we want to include.

Personally I wonder about how useful this will be, as unless buttons are pre-programmed to be using these variables, then they wont do anything, so what is the purpose of creating them this way.  
Perhaps this is wanted as a substitute for creating a module, allowing an external system to push data that can then be used when building buttons?  
And I worry that this could be a start down a road that I think we don't want to go down of having an api to build buttons/triggers. Of course this doesnt mean we have to go that far, but it would be a good for us to define what the limit is.

YWatchman:
> I think this needs a discussion from us maintainers to decide if this is something we want to include.
> 
> Personally I wonder about how useful this will be, as unless buttons are pre-programmed to be using these variables, then they wont do anything, so what is the purpose of creating them this way. Perhaps this is wanted as a substitute for creating a module, allowing an external system to push data that can then be used when building buttons? And I worry that this could be a start down a road that I think we don't want to go down of having an api to build buttons/triggers. Of course this doesnt mean we have to go that far, but it would be a good for us to define what the limit is.

For a radio station, we're working on making the flow for creating triggers easier as a lot has come into play. We use Artisto, which runs a websocket, you have to subscribe to keys in every websocket session. Thus, I've created a proxy which does that and in the meantime it also creates an Alias in the proxy for Companion to understand. Now, actually, we have to create the alias first in Companion and then add it to the proxy. What I want, is when I create a variable / alias in the proxy, for it to be added automatically to Companion. Otherwise, Companion returns a 400 or 404, not sure.

YWatchman:
We have more than a handful of streamdecks, which handle a few different desks. The studio has one, camera operator, director, the techs for liveroom. They can all control a bit of the On-Hertz Artisto flow, by triggering an alias which in term triggers actions that can for example open a talkback channel.

thedist:
Have you considered simply making a custom module for your use case, as it is sounding like that would be a lot more appropriate of a solution than trying to automate variable creation/deletion through a REST API

YWatchman:
I'm just now thinking about it, not sure yet. Combing back to you, it would mean more required maintenance, keeping up2date with Companion. Now it's http module, and using the variables as feedback.

Julusian:
> Combing back to you, it would mean more required maintenance, keeping up2date with Companion

I disagree with this. Modules written for Companion 3.0 (released just over 2 years ago) are compatible with the recent 4.0 releases (as well as the 4.1 betas, and everything between 3.0 and 4.0) and should run just as well as they did in 3.0.  
At some point there will be some required updates for modules, but the hope is that the changes that modules need to make wont be large

Julusian:
There is a consensus currently that we are not keen on having this included at the moment. We feel that for your use case a dedicated module would be best, and we are not currently sure if there is a good use case for having this api.

So for now we are going to reject this. But if anyone has some different justification for this functionality, we are open to hearing why we should include it.

Issue #3526, opened 2025-06-30
[BUG] Button may remain pressed in awesome new "Surface: Set/Adjust Position" action


arikorn:
### Make sure you're on the latest stable or beta build

- [x] I have tested this on the latest stable or beta release

### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion and not a specific module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

First of all, @Julusian, thank you for this awesome feature that adds an Action to set or adjust the surface offset!  It works great so far with one small exception (for which there's a simple-enough work-around for now).

If you add a Press action to modify the current surface, the button is never released internally (the little yellow bar in the button's Topbar stays yellow, for example). As a result, repeatedly pressing a button to "adjust position by 1" -- as it moves to the left -- will only work every other press, since a second press is needed just to generate the release event.

Workaround: only put these actions in the button-release event (or use a different surface).
     (I'm not sure how this would interact with rotary actions on the current surface of something like a StreamDeck+.)

### Steps To Reproduce

1. Create a new button
2. Add "internal: Surface: Adjust position" (or Set Position) to "Press Actions"
3. Set surface/group to current surface
4. Set X offset to 1
5. Press the button
Notice that the topbar stays yellow 
6. Press the button again (in its new position on the surface)
Notice that no action is done, though the topbar now turns off.

### Expected Behavior

Button should release when the physical button is released.

### Environment (please complete the following information)

```markdown
- OS: Windows 10
- Browser: Chrome
- Companion Version: v4.1.0+8152-main-70113dfcc6
```

dnmeid:
I'd say another reason to revert this feature.

arikorn:
Completed with https://github.com/bitfocus/companion/commit/5621893d45748164a1cd49a07400c1695381be8f noted above.

PR #3576, opened 2025-07-25
Fix issue #3526 - Button may remain pressed if it changes its offset

arikorn:
See issue  #3526 for details.

(Minor note: I would have preferred a Map to a Record for `this.#currentButtonPresses` but didn't change it from the original. Obviously, both ways work...)

Closes issue #3526

### DIFF
--- a/companion/lib/Surface/Handler.ts
+++ b/companion/lib/Surface/Handler.ts
@@ -567,7 +567,9 @@ export class SurfaceHandler extends EventEmitter<SurfaceHandlerEvents> {
 				if (controlId) {
 					this.#controls.pressControl(controlId, pressed, this.surfaceId)
 				}
-				this.#logger.debug(`Button ${thisPage}/${y2 + yOffset}/${x2 + xOffset} ${pressed ? 'pressed' : 'released'}`)
+				this.#logger.debug(
+					`Button ${location.pageNumber}/${location.row}/${location.column} ${pressed ? 'pressed' : 'released'}`
+				)
 			} else if (!this.panel.setLocked) {
 				if (pressed) {
 					const pressCode = this.#pincodeNumberPositions.findIndex((pos) => pos[0] == x && pos[1] == y)
--- a/companion/lib/Surface/Handler.ts
+++ b/companion/lib/Surface/Handler.ts
@@ -96,17 +96,11 @@ interface SurfaceHandlerEvents {
 	configUpdated: [SurfaceConfig | undefined]
 }
 
-type MetaLocation = {
-	thisPage: number
-	xOffset: number
-	yOffset: number
-}
-
 export class SurfaceHandler extends EventEmitter<SurfaceHandlerEvents> {
 	/**
 	 * Currently pressed buttons, and what they are keeping pressed
 	 */
-	readonly #currentButtonPresses: Record<string, MetaLocation> = {}
+	readonly #currentButtonPresses = new Map<string, ControlLocation>()
 
 	/**
 	 * Current page of the surface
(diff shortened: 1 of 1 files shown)


arikorn:
Note: build for OSX failed with error

> Error: /Users/runner/work/companion/companion/launcher not a file

https://github.com/arikorn/companion/actions/runs/16531416319

companion/lib/Surface/Handler.ts
```diff
@@ -566,7 +573,8 @@ export class SurfaceHandler extends EventEmitter<SurfaceHandlerEvents> {
 				if (controlId) {
 					this.#controls.pressControl(controlId, pressed, this.surfaceId)
 				}
-				this.#logger.debug(`Button ${thisPage}/${coordinate} ${pressed ? 'pressed' : 'released'}`)
+				const pageCoordinate = `${y2 + yOffset}/${x2 + xOffset}`
```
Julusian: this is only used in the one log line below, so would be better to inline there

companion/lib/Surface/Handler.ts
```diff
@@ -96,11 +96,17 @@ interface SurfaceHandlerEvents {
...
+	xOffset: number
+	yOffset: number
+}
+
 export class SurfaceHandler extends EventEmitter<SurfaceHandlerEvents> {
 	/**
 	 * Currently pressed buttons, and what they are keeping pressed
 	 */
-	readonly #currentButtonPresses: Record<string, number> = {}
+	readonly #currentButtonPresses: Record<string, MetaLocation> = {}
```
Julusian: feel free to change this into a map. there isnt a reason for it to be a record other than it was written that way originally.

Julusian: Im wondering, perhaps instead of storing the offset with this MetaLocation, perhaps this should be holding the ControlLocation type (the absolute position of the button). I think that will make things a bit simpler than storing the offset.

Instead of `let thisPage, xOffset, yOffset: number`, it could be:
```
let location: ControlLocation = {
	pageNumber: thisPage + (pageOffset ?? 0),
	column: x2 + xOffset,
	row: y2 + yOffset,
}
```
then when reading from the `#currentButtonPresses`, you can simply replace the whole of that location object. (optionally, the definition and value for this 'location' variable could be split across the root level and inside the if statement)  
The `this.#pageStore.getControlIdAt` call can then become `this.#pageStore.getControlIdAt(location)` instead of constructing that type in the argument


Looking at the existing code, I am wondering if `pageOffset` is being handled correctly. It should probably be used in the key for that map too..

Julusian (changes requested):

companion/lib/Surface/Handler.ts
arikorn: Done.

companion/lib/Surface/Handler.ts
arikorn: Done.

arikorn: Thanks! I like the idea of using ControlLocation as the map/record value -- once again, thanks for filling in my lack of depth here.  One could take it a step further and store the `controlId` in the map, but it's another "six of one/half-dozen of the other" situation so I left it alone.  However, I did take the liberty to streamline the rest of the code a bit, given this new approach. 

Good point about `pageOffset`, I didn't pay much attention to it.  FWIW, Here's my understanding regarding what should be in the key: the argument to `#onDeviceClick` is the location of the physical button. If you think about what's happening: the physical button does not move between press and release even if the mapping to the internal page(s) changes, so the best key is the arguments passed in to `#onDeviceClick` -- I updated the code and comment to reflect that.

Julusian: yeah I did wonder that too, but I think location makes sense.

> FWIW, Here's my understanding regarding what should be in the key: the argument to #onDeviceClick is the location of the physical button.

yep, and the pageOffset is to allow the surface to do some trickery to fake certain buttons being placed on a different page. It was originally so support large xkeys when the grid was fixed at 8x4 size, I think the only remaining use of it is the videohub panel which translates the selected destination to be the pageoffset. So it is probably not possible for it to have the same xy pressed with different pageOffset at the same time, but maybe it is, and its a trivial change to ensure it will work fine.

PR #3593, opened 2025-08-09
feat: launcher advanced settings

Julusian:
This splits out an 'advanced' settings page in the launcher, and some layout/ux tweaks.

The launcher looks similar, but the minimise and quit button have been removed. The window close button now triggers the minimise (ie closing the launcher window), with quit now being possible from the tray icon:
image
image
image
On linux (the first screenshot) there are extra buttons that we dont want, but they cant be removed and I dont expect many linux desktop users. the minimise button acts the same as close, so its not too bad.
On macos, the middle button is disabled since taking that screenshot

Also in the tray is an 'advanced settings' option which opens:
image

This is partly in preparation for #3534, which can be added here without adding noise to the launcher window.  
And just generally means we can allow configuring more 'environment' based things (such as the log-level, which wasnt possible to change before).  
Generally these should translate to things which we currently define as command line arguments to the companion process, and should be things tied to the environment and/or must not be configurable from the ui itself.


Julusian:
@arikorn continuing the discussion over here

> except for the color scheme.

I default to dark mode for most things these days, I am open to following the os on that, but I think it needs some additional tweaking to look ok there, so maybe not in this first iteration.

>  Oh, and is it really necessary to "enable" developer modules?

:shrug: 
This was a direct translation of the existing config fields, I havent put much thought into that for now. But maybe it should be simplified now

arikorn:
Good idea to break it out, @Julusian. My comments:

1. Enable/Disable
    > This was a direct translation of the existing config fields, I havent put much thought into that for now. But maybe it should be simplified now

    Yes please! In addition to being unnecessary -- both intrinsically for the reasons described in item 1 here: https://github.com/bitfocus/companion/pull/3534#issuecomment-3166603244, and because the user must select "dev" in the v4.0 "connection" tab in the main GUI -- an enable/disable checkbox will also will probably revive the bug I mentioned back in the beginning in this other comment: https://github.com/bitfocus/companion/pull/3534#issuecomment-3145993057 

2. Window "titlebar" buttons
    It took me too long to realize that you differentiated between "minimize" and "hide = minimize to tray".  I really don't like saving this to tray -- my system tray is way too cluttered, with over 20 tiny icons in it.  That said, I would strongly encourage you to put in the minimize button to minimize to the taskbar, and to have "x" actually quit the app (i.e. follow the standard OS behavior) and/or keep the quit button in the main window -- otherwise, it's very confusing (for Windows users, at least)! Also, making people go through a multistep, and somewhat obscure, sequence to actually quit does not sound like a good UX. At the very least, add options for this in your advanced settings.

3. Getting to advanced settings
Forcing the user to go to the tray for advanced settings seems like another step backwards in usability, since it will be rather confusing, especially for people used to the cog. (I honestly did not even realize that the launcher created a tray icon until yesterday.)

4. And to carry over my other suggestion... I'm still in favor of showing a summary of advanced info in the main launcher window, as mocked up here: https://github.com/bitfocus/companion/pull/3534#issuecomment-3153386119

Oh well, 3 & 4 are not really a bit deal, since it would only affect developers (and even I am capable of adapting... 😀 )

Thanks, as always!

thedist:
> Oh, and is it really necessary to "enable" developer modules?

> 🤷
This was a direct translation of the existing config fields, I havent put much thought into that for now. But maybe it should be simplified now

Consider the security implications. If you don't have to enable developer modules, it means any user on a system with multiple operators could just drop any code they want into the module dev fold and have it executed, potentially with elevated permissions if Companion is running as Admin (I don't know if an Admin Companion would spawn child processes that also have those same permissions?). Some productions may want Companion more locked down to only modules that have gone through our approval to be included.

Someone who really wants to do something nefarious will obviously still be able to, but this just provides at least a small amount of protection from an inexperienced user Google how to do something, and reading a post saying download this and put it in the module dev folder and run it.

> Window "titlebar" buttons
It took me too long to realize that you differentiated between "minimize" and "hide = minimize to tray". I really don't like saving this to tray -- my system tray is way too cluttered, with over 20 tiny icons in it. That said, I would strongly encourage you to put in the minimize button to minimize to the taskbar, and to have "x" actually quit the app (i.e. follow the standard OS behavior) and/or keep the quit button in the main window -- otherwise, it's very confusing (for Windows users, at least)! Also, making people go through a multistep, and somewhat obscure, sequence to actually quit does not sound like a good UX. At the very least, add options for this in your advanced settings.

Depending on the type of app, it's very common for apps that are designed to often run in the background for the X button on windows to minimize to tray, that's perfectly normal. Apps do often have a setting option to control if it should minimize or close, so that the user has the final say in how they want to control it. Personally for the type of app Companion is, I'd expect the close button to minimize to tray by default in Windows, and I can't speak for other OS's.

> And to carry over my other suggestion... I'm still in favor of showing a summary of advanced info in the main launcher window, as mocked up here: https://github.com/bitfocus/companion/pull/3534#issuecomment-3153386119

I'm not a fan of over-complicating the window. What if in the future we wanted to utilize the window for other things, such as include a link to the latest patch notes, or Companion news, or Social links like to Slack or a forum, or important messages. While we can already give users this information through the Web UI, this is just an example of another place which we can utilize to inform end users. Spamming that space with settings (as who knows how many more options we may wish to add in the future) would lead to a bad user experience compared to an advanced settings panel which gives plenty of room for future expansion and leaves the main Window clean.

arikorn:
@thedist wrote
> it means any user on a system with multiple operators could just drop any code they want into the module dev fold and have it executed, potentially with elevated permissions

Jeff, if I understand you correctly, you believe that having an enable/disable button will prevent a malicious actor from running their own code as a dev module. This implies that they are sophisticated enough to develop a module but not sophisticated enough to know how to test and run their code (since if they tested or ran their code previously, they would have had to encounter this button)... For that matter: how did they even know which folder to put it into without seeing the window with the "enable" button?

Just a reminder (@Julusian): if you keep the enable/disable button, be sure to fix the bug it creates.

>Window "titlebar" buttons

It sounds like we agree that the user should have the option to have the "x" button quit the app entirely.  Thanks!

I want to emphasize again, that my main and original request was to include a "minimize to taskbar" option (i.e. "_" in Windows/Linux).

>I'm not a fan of over-complicating the window. ... Spamming that space with settings

I'm pretty sure it's not considered "spam" or even "over complicating" if the user explicitly requested it. Showing current dev status has always been proposed as a user option.

@Julusian, I am concerned that this feature has gotten away from its original intention and is leading to a frustrating UX. Please bear with me here: Judging from the timeline of our comments in #3534 and your commits here, you started this feature in response to my suggestions (starting with: https://github.com/bitfocus/companion/pull/3534#issuecomment-3141848465) that (a) it would be good to have the standard minimize button and (b) that there is a bug introduced in v4.0 when the "cog" is turned off.  Your [initial response](https://github.com/bitfocus/companion/pull/3534#issuecomment-3146428278) addressed both of these issues by replacing the "Hide" and "Quit" buttons with a "_" and "x" where the titlebar might have been, and adding a "Developer" button in their place that would reveal developer options and which you confirmed would resolve the "cog" issue too... Unfortunately, the current version of this PR, reverses these features: keeping a "disable" button ensures that the "cog" issue is not resolved here, and you have removed the "minimize to taskbar" button and redefined the "close" button/action to not quit the application. Speaking as a user and as a module developer, this hurts the UX:  for "regular" users, there is no longer an easy way to quit the launcher/server, instead one has to know to look for a tray icon, click and find the (likely hidden) tray icon, then click again to raise a menu then click once more to select quit. For the module developer, the situation is even worse: in the current launcher a developer knows instantly by looking at the launch window whether the dev interface is active, and it takes a single click to turn it on. With this PR, all of the information is hidden and there is no simple way to get to it, instead the user/developer has to go through multiple clicks in different windows to find out the current status. And if not, and they get tripped up when they start the Web UI, they have to backtrack, go through the multiple-click process, etc., if they have the good fortune of realizing that this is the cause of their problems. I know you agree with me that this level of inconvenience is worth avoiding since in a recent PR of mine you had me make substantial changes to avoid similar circumstances: here I would like to urge you, not so much to make substantial changes, but to go back to the original ideas so that the user base can avoid the confusion and frustration...

thedist:
> This implies that they are sophisticated enough to develop a module but not sophisticated enough to know how to test and run their code

You don't need to be a developer to use dev modules. If a production doesn't want the dev modules folder enabled, why force it on them by removing them the choice? It shouldn't be that big a deal.

> I'm pretty sure it's not considered "spam" or even "over complicating" if the user explicitly requested it. Showing current dev status has always been proposed as a user option.

Your very own picture you linked to in the other thread shows the small launcher window with very little space for the options you're bloating it with, and keep in mind as I previously stated you have to consider the future and what options may be added over time. A separate advanced Settings screen such as what Julian showed seems much more suitable.

I don't get why things are being over complicated here, or trying to reinvent the wheel in app design,

arikorn:
@thedist wrote:
> If a production doesn't want the dev modules folder enabled, why force it on them by removing them the choice? It shouldn't be that big a deal.

Not a big deal, it just introduces a bug. Please read #3534 or follow the links I've posted earlier. To restate what I wrote in my previous comment:  "if you keep the enable/disable button, be sure to fix the bug it creates."

@Julusian, a last thought: it's possible that reinstating the "clear" button would solve both issues, i.e. disable dev modules and avoid the bug, but I suspect not.

>Your very own picture you linked to in the other thread shows the small launcher window with very little space for the options you're bloating it with,

Since the window I show is actually several lines smaller than the current window, I find it difficult to call it "bloat" or "very little space" but more to the point, and I paraphrase your own words, as quoted above: "If a production wants to see the dev modules folder in the launcher, why forcibly prevent them?" This is a feature that currently exists, why insist that it must be removed?

 If you want more detail, Jeff, on why I advocate for either of these issues, please read my previous comment, which ends with an explanation of why it could make life difficult for developers.

OK, I think we've reached the end of constructive conversation here. Please feel free to reply. I'll do my best to let you have the last word 😀 .

Julusian:
> It took me too long to realize that you differentiated between "minimize" and "hide = minimize to tray". I really don't like saving this to tray -- my system tray is way too cluttered, with over 20 tiny icons in it. That said, I would strongly encourage you to put in the minimize button to minimize to the taskbar, and to have "x" actually quit the app

Companion has always had a tray icon, and the minimise behaviour has always been to the tray.  
What I am proposing here is consistent with a lot of other software, such as the elgato strreamdeck software or things like discord or slack. As there isnt really anything to interact with in the application window it doesnt make sense to minimise to the taskbar, I know that I never interact with it other than maybe clicking the 'launch gui' button once at startup. So I think that all that will do is confuse/annoy users as many wont realise it can be minimise to the tray.

Another thought I have had is maybe this window simply shouldnt exist? In the satellite app, there is no equivalent window, it simply runs in the tray and has a settings window that can be opened.  
In reality, I dont think that removing this window is a good idea, that will cause too much user confusion. But I am struggling to give it a purpose.  
Maybe it should be the settings window, and except for the first launch we could make it default to not showing on launch? Or make it the settings window with a 'status' tab to it to replicate the current content.  
My concern is that putting the settings in the same window will make it feel less sleek/polished.

> Forcing the user to go to the tray for advanced settings seems like another step backwards in usability, since it will be rather confusing, especially for people used to the cog.

I dont entirely disagree with this, I am wondering if there should be a button in the window too. But as this is likely to be used by <1% of users, it shouldnt be too prominent. The cog was nice before, but with that space now occupied, I'm not sure if it should be made more prominent by putting it along the bottom or not.

> Personally for the type of app Companion is, I'd expect the close button to minimize to tray by default in Windows, and I can't speak for other OS's.

On macos I believe most apps (certainly many apple ones do, eg safari and textedit) when you close the last window they will remain in the dock (taskbar), and you have to explicitly quit them from a menu. Or doing the same but to the tray I think is also common (I am less certain of this, dont have anything installed which I would expect to do this)



> it means any user on a system with multiple operators could just drop any code they want into the module dev fold and have it executed, potentially with elevated permissions

This is only about whether to have an 'enabled' checkbox as well as choosing a path. Because if a path is defined it should be reasonable for us to use that, and if they want to disable it then clear the path instead. So no different in security than today, just the UX of the settings.

> Just a reminder (@Julusian): if you keep the enable/disable button, be sure to fix the bug it creates.

tbh Im not sure if I would call that a bug. It may be weird and potentially undesirable, but it works as intended.  
When disabling dev modules, then yeah the connections wont run. In the same way that uninstalling a needed module version will do similarly.
And the cog behaviour makes sense for 3.x (because the dev versions would override the builtin ones), and less so now but I think it makes sense so that the field only applies when visible (in the current version). I think that changing this behaviour now will be confusing for people as it has worked as a toggle for years. But moving it entirely gives a clearer indicator that things are different.

> for "regular" users, there is no longer an easy way to quit the launcher/server, instead one has to know to look for a tray icon, click and find the (likely hidden) tray icon, then click again to raise a menu then click once more to select quit.

Again this is no different than many other software, and is something that most users will already need to know as once it is minimised/hidden it is only present in the tray.

> Speaking as a user and as a module developer, this hurts the UX: for "regular" users, there is no longer an easy way to quit the launcher/server, instead one has to know to look for a tray icon, click and find the (likely hidden) tray icon, then click again to raise a menu then click once more to select quit. 

On any machine I have used in production, it always gets minimised to tray because I will end up with too much clutter in the taskbar already. So in my experience, this is no different to today.

> For the module developer, the situation is even worse: in the current launcher a developer knows instantly by looking at the launch window whether the dev interface is active, and it takes a single click to turn it on. With this PR, all of the information is hidden and there is no simple way to get to it, instead the user/developer has to go through multiple clicks in different windows to find out the current status.

Maybe I am unusual in this, but on my machines they are either dev machines and have this enabled with a well known path, or they are production machines and wont have it enabled, so it doesnt matter.  
I am open to adding some info about what is configured in the window, but it needs some thought and design on how to present it so that it isnt cluttered or too noisy. and this feels like something that could be a follow up, or could be part of this if there is a good pitch that inspires me to do it. 

> This is a feature that currently exists, why insist that it must be removed?

The only reason it is there is because it was easier to add there than to make a new window for it. And I intentionally put it behind the cog so that it would only show for those who needed to interact with it.  
So as it wasnt really thought about before, I want to question whether it should be there or if it is just expected to be there because it used to be there

> And to carry over my other suggestion... I'm still in favor of showing a summary of advanced info in the main launcher window, as mocked up here: https://github.com/bitfocus/companion/pull/3534#issuecomment-3153386119

My big problem with adding more config to it is that users __will__ misinterpret it. I cannot count the amount of times users have advised each other that they need to change the 'gui address' field to their lan address to resolve a module not connecting, even though we keep telling them that this has no effect on that (I might be half convincing myself to look at whether these fields should be moved too). So unless the fields have clear descriptions and explanations users will set them thinking they do something they dont.  
So in the mockup you did 'Companion hostname' is way too vague and users will start thinking they need to set it (and who knows what else they may do to the other fields there at the same time).  

This is why I started on this separate window, so that we have the space to explain this next to the fields.  
Fiddling with the minimise/close behaviour was something that makes sense to do (I dont see the window enough to care that much about it, it can go months between times I see it), but I am not set on, so if we cant agree here will likely be broken out into its own PR

thedist:
> What I am proposing here is consistent with a lot of other software, such as the elgato strreamdeck software or things like discord or slack. As there isnt really anything to interact with in the application window it doesnt make sense to minimise to the taskbar, I know that I never interact with it other than maybe clicking the 'launch gui' button once at startup. So I think that all that will do is confuse/annoy users as many wont realise it can be minimise to the tray.

Agreed, it's been standard for apps like this that generally run in the background and stay running (but are not services) to minimize to tray like this for many, many years. Using Slack as an example, there are preferences options to change from the default of close = minimize to tray to close = quit the app, so perhaps that's an option to consider to give user choice but by default minimize to tray.

> In reality, I dont think that removing this window is a good idea, that will cause too much user confusion. But I am struggling to give it a purpose.

My thoughts for the Companion window is more basic stuff, such as version number, and indicator if new versions are available, links to support such as Slack or GitHub, and potentially some visual indicator if there are errors. While some of this may be redundant with what the WebUI shows, if for whatever reason a user can't get the WebUI to work or doesn't know how to access it providing support links and info if there's a new version could be beneficial to have here too. Then settings could be in a separate Advanced Settings window.

> This is only about whether to have an 'enabled' checkbox as well as choosing a path.

While it may be infrequently changing, it can be beneficial to have a checkbox so that it can be enabled/disabled without having to type or copy/paste the path each and every time. Having a field set make it always active, and having it empty make it disabled, is quite odd behaviour. One example, I could work on a production where it's pointed at a network share, and if for whatever reason that destination is unavailable or the route isn't running then disabling a checkbox would keep that location written in the field but I would know Companion wouldn't be trying to access a location that is unavailable. Or if moving between different productions sometimes I'd want it enabled, sometimes disabled, and a checkbox would simplify this over editing a path each time.

This is a niche thing though as most places I've seen running Companion change this infrequently so it's not a huge thing either way.

> This is why I started on this separate window, so that we have the space to explain this next to the fields.

I'm in favour of a separate window too, having more space to be clear about the settings available, and it also give us a better footing if we wanted to add more settings here in the future.

arikorn:
Thank you @Julusian  and @thedist for your considered responses... I agree with everything Jeff wrote in the previous comment and much of what you wrote, Julian. OTOTH, I feel like we are juggling too many issues at the moment. Maybe organizing it will help a bit? And if you see something I missed, feel free to point it out. Maybe we can check off items as we go. Here's the TOC:

0. Color scheme of the settings window should be closer to the system colors. (or at least allow a light scheme?)
1. "Title bar" ornaments
1a. Close button behavior ("x" and also the "close" role), or how to keep the functionality of the current "Quit" button.
1b. Existence and function of a "minimize to taskbar" button ( "_" )

2. How can a user access advanced settings?
2a. Only from the system tray? (Does this reduce the dev's UX?)
2b. A button in the launch window? (But not too prominent.)
2c. new: Maybe reinstate the cog (but give it a slightly different meaning)?

3. Should there be a launch window at all?
3a. Do people interact with it now?
3b. Regardless, what goes in it? Should it just be a settings window (show all settings?)?
3b. ...and how often to show it?

4. Allow an enable/disable button for dev modules folder?
4a. Should we have one?
4b. Does having one create a bug?
4c. Would having one generate confusion, especially if it his "hidden" deep in a click-tree.

5. Should settings-summary be allowed to be shown in the launch window?
5a. Will it lead to confusion/misinterpretation?

6. Underlying it all: what are common/expected use cases as it is relevant to the launcher window?
6a. Julian: different machines; never use dev on production machines; never interact with the launch screen
6b. Jeff: multi-operator gigs where director just wants a "disable dev" button, or where a network drive is missing.
6c. Me: "Occasional" use; always run dev module! Never use the system tray (for launcher).

It probably pays to start with that last issue because it informs all the rest.

**6c**: I use Companion only a few times a week, for development or production. I became a module developer to get the functionality I want (and to share with others, of course), but our release schedule is infrequent enough that I just run my current dev in production. The module is stable enough that I've never had any issues. Regardless of where I use it, the easy access to quit the server, streamlines both productions (when I need to update Companion) and development, where I quit it at the end of a coding session (since I don't use it every day). 

OK Let's go through the five "features" (bold just to highlight the topic, not to shout)
[_Edit_: Also,  just to be clear, for the following: I don't care if launcher puts an icon in the tray while it's open, I just would prefer not to be forced to interact with it.]

**Item 1 "Standard title bar icons**: _allow_ "x" to quit, add "_" to minimize to taskbar:
**1a**. I very much want to be able to **quit the app when I close it**, and note that the "Quit" button does this today and is prominently featured. In other words, many people will expect it too. As Jeff pointed out, popular apps that use the system tray give user the choice. When I click "x" on Slack or Spotify, it quits the app - including the tray. Presumably I set it and forgot it long ago. So both because of precedent in Companion (i.e. the quit button) and in other apps, I think we should allow the "x" to quit (possibly as an option, to satisfy those who want Companion to be up full-time.) And honestly, for my use-case, Companion is not like a Slack or Spotify that have a reason to be running in the background 24/7.
   [_Edit_:   the simplest solution is to just keep the ("hide" and) "quit" buttons as they are now.]
   FWIW, the Companion "launcher" includes the whole server. It uses substantial resources on my computer: more memory than anything but Chrome, which is running 14 windows and dozens of tabs. This screen shot is taken shortly after starting the launcher, after it settled down, and with no surfaces attached, and not connected to my mixer (Osee GoStream).
image

**1b**. Although the precedent in Companion is to hide-to-tray (currently the Hide button), I never minimize to the tray. Instead, I either leave the launcher behind the browser or I minimize to taskbar -- either way it is quickly raised from the taskbar. If we use the popular apps as the standard, all the ones I know **include "_" buttons in the title bar** and when used they **minimizes to the taskbar** (Slack, Spotify, etc.). FWIW, my preference for the taskbar -- even though I have 20 icons on my taskbar too -- is that it is so much bigger than the tray, i.e. icons are more visible, and also simpler to access.

**2. Where to allow access to advanced settings.**
Probably the simplest solution is to **reinstate the cog**. Once again, there's precedent in standard apps such as Slack and Spotify. Julian, you're already using a "custom titlebar" so it should be easy enough to reinstate the cog next to the minimize button in Windows or wherever you had it in MacOS. FWIW, just to be clear, the cog will "mean" something slightly different internally: clicking it will open a settings window and nothing else, no side-effects.

**3. Should there be a launch window at all**
Perhaps we can table this one for now? I think we all agree that current users expect a launch window? FWIW, I agree with Jeff's ideas, particularly that it would be a good place to encourage updating. And in item 5, below, I suggest that other's may value information on it. Also, I actually do interact with it occasionally: for whatever reason it seems as if Bonjour gets lost sometimes and only works after I change the GUI interface and relaunch (and sometimes I do it when running side-by-side with a core-dev version to more clearly differentiate the two?). Also, if I'm using the LAN IP, it occasionally changes due to DHCP, and the launcher needs to be changed too.

**4. Allow an enable/disable button for dev modules folder**
Surprise here: **I'm on Jeff's side** for this one now, too. Yes, it's technically redundant and yes, it's doesn't really "architecturally" fit with the way v4 handles dev modules, but it's just not a big deal and there will be people who just "want" the button. Especially, Julian, if you don't consider the problem when it's off to be a bug. I do still advocate making it easily accessible (from the launcher window) and making status option visible.

**5. Should settings-summary be allowed to be shown in the launch window**
I'm still in favor of this. I agree totally with you, Julian, that having a settings window allows for more explanation, and is a good thing. Yet that doesn't negate allowing the user to request to status on the main window. I don't see how it will add confusion: it will only be seen by people who request it, which means they'll have been educated by the settings window. Also, your specific example of  'Companion hostname' has simple solutions: as you alluded to, it needs a clearer label -- as [I already recommended in the original PR](https://github.com/bitfocus/companion/pull/3534#issuecomment-3156267626). More to the point, the only people who will see that bit of info are ones who are using syslog _and_ have opted to see it on the main screen, so they will know it's a syslog option. (and seeing hostname on the launch screen will probably be very valuable to them, based on the OP's comments there).  My recommendation here is to **preserve, as default, the current behavior**: anyone who has enabled dev modules should see the dev folder by default (and clicking on it takes them to the settings window). Everyone else starts with a _tabula rasa_, as they do now.

arikorn:
Some quick edits (will change it above too)
1a. silly me:  or just keep the ("hide" and) "quit" buttons as they are now!

also, just to be clear: I don't care if launcher puts an icon in the tray while it's open, I just ~~don't want to~~ would prefer not be forced to interact with it.

and.. for completeness: 
0. Color scheme of the settings window should be closer to the system colors. (or at least allow a light scheme?)

Julusian:
Picking this up again after a couple of weeks away;

0) I am not opposed to this, but I think the light mode of the ui toolkit looks bad (it has very little contrast). And this sounds like something that could easily be done as a follow up, if anyone is bothered enough by the lack of light mode to figure it out.
1) I am binning off the titlebar/hide/quit rework as part of this. That can be done later if someone is motivated enough to figure out how it should work and do it
2) That means the old cog button can easily remain, which resolves that
3) This was an idea that I didnt think was something good to do, so we can ignore this
4) Sounds like it should stay then
5) tbh I am hitting the problem here of not seeing much value in this, and it being a bunch of complexity to implement it. So am not particularly keen to do it myself. I am pretty tempted to see if there are many complaints from the wider user base before doing something on this.

> and seeing hostname on the launch screen will probably be very valuable to them, based on the OP's comments there

that is an interesting point. I would not have assumed that to have been the useful bit of information. I would expect that seeing only that syslog is configured at all (and perhaps an ok/error status indicator) would be all that is needed. especially considering other software/hardware sometimes doesnt let you provide the hostname separate to the rest of the system.


-----

So with all that, this is now in a slightly simpler state, and I dont have any outstanding questions that I want answers to before this can be merged.

Unless anyone disagrees and has any remaining objections, I shall merge this soon

arikorn:
Thanks @Julusian, sounds good.  

1. FWIW, it's a bit of a shame to toss the titlebar stuff (item 1) since you've done all the heavy lifting already. But otherwise, I'm happy to go with your decision.

5. Just to clarify: (I agree that hostname shouldn't be an option, but) what I meant here, is that a person using syslog may want to quickly check the hostname in order to confirm which device sent a particular syslog message.  No problem with your decision to put this off for now and/or to take it on myself in the future if it really bothers me.

  On the general topic of the cog, please make sure that clicking the cog no longer enables/disables developer mode. You've probably done it, but I wasn't able to run the launcher in my WSL dev environment -- at first `yarn dev` complained that it couldn't find electron, then I added it with `yarn add electron --dev`, which was perhaps unnecessary and ran `yarn install` both in the parent folder and in the launcher folder but now it complains about a missing "libnspr4.so"  If you have any advice, I'd certainly appreciate it!

Julusian:
1) As there isnt an agreement on how it should behave, I dont want to hold the rest of this back. For anyone wanting to pick this up, I kept the undoing of this as a single commit 149244a8ba8f16b9149ca3abaf15181bb0769e3b which can be your starting point for redoing it

PR #3607, opened 2025-08-29
feat: configure syslog from launcher

phillipivan:
Add syslog configuration to the new launcher `Advanced Settings` window.

I would prefer to allow specifying the syslog server via hostname as well as IP; however when an invalid hostname is supplied, companion will crash when trying to write to the logs, and this is not caught by the `try/catch` block that sets up the syslog transport for winston.

When the `Local Hostname` field is empty, the system defaults to `os.hostname()`

Trigger restart debounce duration extended slightly to make it more tolerant of typing in the text fields.

image

### DIFF
--- a/companion/lib/main.ts
+++ b/companion/lib/main.ts
@@ -21,6 +21,25 @@ import logger from './Log/Controller.js'
 import { ConfigReleaseDirs } from '@companion-app/shared/Paths.js'
 import { type SyslogTransportOptions } from 'winston-syslog'
 
+const isValidIPv4 = (ip: string): boolean => {
+	const parts = ip.split('.')
+	if (parts.length !== 4) {
+		return false
+	}
+	for (const part of parts) {
+		const num = Number(part)
+		// Check if it's a valid number and within range [0, 255]
+		if (Number.isNaN(num) || num < 0 || num > 255) {
+			return false
+		}
+		// Disallow leading zeros for numbers other than '0' itself
+		if (part.length > 1 && part[0] === '0' && num !== 0) {
+			return false
+		}
+	}
+	return true
+}
+
 const program = new Command()
 
 program.command('check-launches', { hidden: true }).action(() => {
@@ -73,13 +92,13 @@ program.command('start', { isDefault: true, hidden: true }).action(() => {
 
 	if (options.syslogEnable) {
 		const opt: SyslogTransportOptions = {}
-		if (options.syslogHost) opt.host = options.syslogHost
+		if (isValidIPv4(options.syslogHost)) opt.host = options.syslogHost
 		if (options.syslogPort) {
 			const port = Number.parseInt(options.syslogPort)
 			if (!Number.isNaN(port) && port > 100 && port <= 65535) opt.port = port
 		}
 		if (options.syslogTcp) opt.protocol = 'tcp4'
-		if (options.syslogLocalhost) opt.localhost = options.syslogLocalhost
+		opt.localhost = options.syslogLocalhost ? options.syslogLocalhost : os.hostname()
 		logger.addSyslogHost(opt)
 	}
 
--- a/companion/lib/main.ts
+++ b/companion/lib/main.ts
@@ -80,7 +80,11 @@ program.command('start', { isDefault: true, hidden: true }).action(() => {
 			if (!Number.isNaN(port) && port > 100 && port <= 65535) opt.port = port
 		}
 		if (options.syslogTcp) opt.protocol = 'tcp4'
-		opt.localhost = options.syslogLocalhost?.trim() || os.hostname()
+		const localhost: string = options.syslogLocalhost
+			?.replaceAll(/[^a-zA-Z0-9-_. ]/gm, '')
+			.substring(0, 254)
+			.trim()
+		opt.localhost = localhost || os.hostname()
 		logger.addSyslogHost(opt)
 	}
 
--- a/companion/lib/main.ts
+++ b/companion/lib/main.ts
@@ -20,25 +20,7 @@ import { nanoid } from 'nanoid'
 import logger from './Log/Controller.js'
 import { ConfigReleaseDirs } from '@companion-app/shared/Paths.js'
 import { type SyslogTransportOptions } from 'winston-syslog'
-
-const isValidIPv4 = (ip: string): boolean => {
-	const parts = ip.split('.')
-	if (parts.length !== 4) {
-		return false
-	}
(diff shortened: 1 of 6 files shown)


companion/lib/main.ts
```diff
@@ -21,6 +21,25 @@ import logger from './Log/Controller.js'
 import { ConfigReleaseDirs } from '@companion-app/shared/Paths.js'
 import { type SyslogTransportOptions } from 'winston-syslog'
 
+const isValidIPv4 = (ip: string): boolean => {
```
Julusian: maybe this could be simplified to use https://nodejs.org/api/net.html#netisipv4input ?

companion/lib/main.ts
```diff
@@ -73,13 +92,13 @@ program.command('start', { isDefault: true, hidden: true }).action(() => {
...
 		const opt: SyslogTransportOptions = {}
-		if (options.syslogHost) opt.host = options.syslogHost
+		if (isValidIPv4(options.syslogHost)) opt.host = options.syslogHost
 		if (options.syslogPort) {
 			const port = Number.parseInt(options.syslogPort)
 			if (!Number.isNaN(port) && port > 100 && port <= 65535) opt.port = port
 		}
 		if (options.syslogTcp) opt.protocol = 'tcp4'
-		if (options.syslogLocalhost) opt.localhost = options.syslogLocalhost
+		opt.localhost = options.syslogLocalhost ? options.syslogLocalhost : os.hostname()
```
Julusian:
```suggestion
		opt.localhost = options.syslogLocalhost?.trim() || os.hostname()
```

I suspect trimming the value would be a good idea, partly to ensure it doesnt get set to just something like ` ` and result in not sending anything.

Feel free to ignore if you think it should remain as you have written

launcher/main.js
```diff
@@ -893,6 +924,13 @@ if (!lock) {
 						`--log-level=${uiConfig.get('log_level')}`,
 						uiConfig.get('enable_developer') ? `--extra-module-path=${uiConfig.get('dev_modules_path')}` : undefined,
 						disableAdminPassword || process.env.DISABLE_ADMIN_PASSWORD ? `--disable-admin-password` : undefined,
+						uiConfig.get('enable_syslog') ? '--syslog-enable' : undefined,
+						uiConfig.get('enable_syslog') ? `--syslog-host=${uiConfig.get('syslog_host')}` : undefined,
+						uiConfig.get('enable_syslog') ? `--syslog-port=${uiConfig.get('syslog_port')}` : undefined,
+						uiConfig.get('enable_syslog') && uiConfig.get('syslog_use_tcp') ? `--syslog-tcp` : undefined,
+						uiConfig.get('enable_syslog') && uiConfig.get('syslog_local_hostname')
+							? `--syslog-localhost=${uiConfig.get('syslog_local_hostname')}`
```
Julusian:
```suggestion
							? `--syslog-localhost="${uiConfig.get('syslog_local_hostname')}"`
```

In case it contains any spaces?

Arguably maybe there should be some more sanitisation of the values, but I suspect it doesnt matter, and it can't exactly do much harm (the user already has local access to the machine, with the same priviledges)

launcher-ui/src/sections/Syslog.tsx
```diff
@@ -0,0 +1,143 @@
...
+								<Input
+									id="syslog_local_hostname"
+									type="text"
+									value={syslogLocalHost}
+									onChange={(e) => handleSyslogLocalHostChange(e.target.value)}
+									className="flex-1 rounded-r-none"
+								/>
+							</div>
+							<p className="text-sm text-muted-foreground mt-1">
+								Change the hostname this companion instance will report
```
Julusian:
```suggestion
								Change the hostname this companion instance will report. If not set the system hostname will be used
```

Julusian (approved):
Looks good, just a couple of minor questions

phillipivan:
@Julusian I was trying to get the `os.hostname()` value to populate the `placeholder` field for the Local Hostname input, but my all my attempts have been thus far unsuccessful.

I thought something like `placeholder={os.hostname()}` would do it, but it doesnt seem to like that. Can you point me in the right direction?

launcher-ui/src/sections/Syslog.tsx
```diff
@@ -0,0 +1,143 @@
...
+						</div>
+					</div>
+					<div className="grid grid-cols-4 gap-4 items-center">
+						<Label htmlFor="syslog-local-host">Local Hostname</Label>
+						<div className="col-span-3">
+							<div className="flex">
+								<Input
+									id="syslog_local_hostname"
+									type="text"
+									value={syslogLocalHost}
```
phillipivan, line 131: @Julusian I was trying to get the os.hostname() value to populate the placeholder field for the Local Hostname input, but my all my attempts have been unsuccessful.

I thought something like `placeholder={os.hostname()}` would do it, but it doesn't seem to like that. Can you point me in the right direction?

launcher/main.js
phillipivan: I'll add some sanitation to limit length and character set. At least with the syslog server I've been testing with (Kiwi) whitespace internal to the hostname has been handled fine, even though that wouldn't be allowable in an actual hostname. So I'm not sure if said sanitation should be strict (`[a-zA-Z0-0_-]`) or rather more permissive.

launcher-ui/src/sections/Syslog.tsx
Julusian: I would be happy with it not showing anything, but if you want to you will need to do the `os.hostname()` query in the `main.js` of the launcher.

I would suggest adding it as part of the 'info' ipc message send to this window

phillipivan: image

phillipivan:
I've sorted the placeholder value for the hostname. I think this is all done now unless you have any other feedback.

launcher/main.js
```diff
@@ -116,6 +116,12 @@ if (!lock) {
 		enable_developer: false,
 		dev_modules_path: '',
 		log_level: 'info',
+
+		enable_syslog: false,
+		syslog_host: '127.0.0.1',
+		syslog_port: 514,
+		syslog_use_tcp: false,
+		syslog_local_hostname: os.hostname(),
```
Julusian:
```suggestion
		syslog_local_hostname: '',
```
It defaults to using `os.hostname()` later on, so I dont think we should be baking the current hostname into the config file. It will just lead to it not following the current hostname unless the user explicitly clears that field

PR #3620, opened 2025-09-07
feat(graphics-overhaul): support 'feedback overrides'

Julusian:
This is an alternate proposal to #3482

**This is an early POC, the UX needs some attention and I have a plan to make it nice.**

I said early on in the graphics overhaul that keeping the override behaviour of feedbacks will not work. I maintain that it won't work in the same way as before, but this is the next best approach I can come up with for handling this.

The plan is a new 'overrides' tab next to the style. In this will be a pretty normal list of feedbacks, but with a different footer below them:
image

For the path column, as this is a POC this is the crudest implementation possible. A final version of this will have a fancy component that will allow selecting an element and property based on the currently defined elements. The intention is to store the elementId and propertyName, as that should make it pretty durable to elements being moved around the tree, and not be tied to user defined names (which will collide at times), but I am open to it referencing an element name if we think that would be better?

The idea is that this will accept boolean or advanced feedbacks, and they will be handled a little differently.

For boolean feedbacks, the 'value' column will be an input field matching the type used by the chosen path, it will get the same 'value or expression' behaviour as the style editor.
For advanced feedbacks, the 'value' column will be a dropdown to choose between the values that we support from advanced feedbacks.

My hope with this, is that while it is a little more work to hookup the overrides, it can behave the same as before and most importantly with no need to touch an expression.  
Advanced feedbacks are particularly clunky in this, and we wont be able to give any guidance on which of the properties the feedback may produce, but I see this as a backwards compatibility thing, while we phase these feedbacks out.


Julusian:
I think this is now fully functional, and is a little more polished, but needs some more polishing still

PR #3622, opened 2025-09-08
feat: rework import pages ux

bryce-seifert:
This reworks the import page UX to make each step more clearly grouped, and explain to the user what the import in each case will do. It also looks to separate the different import buttons to more clearly differentiate them.

Before Full
Screenshot 2025-09-07 at 4 58 42 PM

After Full
Screenshot 2025-09-07 at 4 57 04 PM

Before Buttons
Screenshot 2025-09-07 at 5 02 37 PM

After Buttons
Screenshot 2025-09-07 at 4 57 38 PM

Before Triggers
Screenshot 2025-09-07 at 4 58 52 PM

After 
Screenshot 2025-09-07 at 4 57 17 PM
Triggers

### DIFF
--- a/webui/src/ImportExport/Import/Triggers.tsx
+++ b/webui/src/ImportExport/Import/Triggers.tsx
@@ -85,6 +85,10 @@ export function ImportTriggersTab({
 			<h4>Triggers</h4>
 			<p>Select the triggers you want to import.</p>
 			<table className="table table-responsive-sm mb-3">
+				<colgroup>
+					<col style={{ width: '5rem' }}></col>
+					<col style={{ width: 'auto' }}></col>
+				</colgroup>
 				<thead>
 					<tr>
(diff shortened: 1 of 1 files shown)


Julusian (approved):
Looks great!

My only comment is that I think the checkbox column in the table of triggers should be kept as narrow as possible (like it was before).  
I think that it will make it easier for users to match checkbox with name. And the contents wont ever be wider than a checkbox, so it will make for a more consistent layout/experience if the position of the name column is fixed to be close to it.

bryce-seifert:
Interestingly, that seemed to only be present in Safari. (I guess I took the before in Chrome, and after in Safari as I was testing)

So I tried a workaround to try and force Safari into a more narrow column. Safari on left, Chrome on right
Screenshot 2025-09-08 at 8 33 50 AM

PR #3651, opened 2025-09-24
FIX: Negative value in msToTimestamp

alex-Arc:
**The issue**
`secondsToTimestamp` prepends negative number with a minus sign while `msToTimestamp` dose not

**Other stuff**
add test for the representing the new functionality.
unified the path for timestamp generation in both the millisecond and seconds case 

**Questions**
I could add formatting for days as well to fix the note about more than 24 hours?
I could implement a non padding option, "h:mm:s" -> 1:05:30 ?

### DIFF
--- a/docs/4_secondary_admin_controls/expressions/functions.md
+++ b/docs/4_secondary_admin_controls/expressions/functions.md
@@ -242,7 +242,7 @@ You can do the reverse of this with `secondsToTimestamp(str)`
 
 Convert a number of seconds into a timestamp of format 'HH:mm:ss'.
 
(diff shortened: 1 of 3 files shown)


Julusian:
> I could add formatting for days as well to fix the note about more than 24 hours?

I think this was a semi-intentional choice, Im not sure if it should be changed.

> I could implement a non padding option, "h:mm:s" -> 1:05:30 ?

Yeah, feel free to do this. I do think it would be good to have a better set of formatting options here, perhaps it should use the `dayjs` formatting library we are using elsewhere but I am also concerned that we might want to only allow a subset of that and perhaps with different letters in places.

PR #3658, opened 2025-09-28
fix: very old contour shuttle pro v1 has an XKey vendor ID

arikorn:
See my related PR in: https://github.com/nytamin/contour-shuttle/pull/3

This fix moves the test for contour shuttle before the XKey tests. Otherwise and error is generated saying that the (older-model) Contour Shuttle is an unknown XKey device.


Julusian:
you've not handled the `ProductModelId.ShuttleProV1a` id in this

arikorn:
Odd that VS Code didn't flag it for me last night when I tested it.  Oh well, it's a case that never happens, because `productModelId` is set to ShuttleProV1 inside the `PRODUCTS` variable,  but there's no harm in handling it.

Looks like it all checks out now.  Hopefully you can include it in the next minor update.  Thanks @Julusian !!
And thanks for adding the comment and updating the shuttle-node version!

PR #3659, opened 2025-09-28
feat: run some ui tests on browserstack

Julusian:
We have a sponsored license to browserstack.

As there have been issues every now and then with accidentally breaking the emulator on older browsers, we can utilise browserstack to have some automated tests to try to catch these issues early.

This is a minimal PR that sets up playwright and integrates it with browserstack to run tests in a handful of browsers.

Over time, we could try and add more tests, for general application usage, if someone wants to take on that task.


Julusian:
After fighting with this for a few days, and then a week of trying to talk to support about this, I have lost all motivation for it and am dreading the fighting this entails.

The value of this proposal was for testing older browsers, which is annoyingly the main thing that was not working at all, so this also has lessened in value.

I am closing this as I cant see myself continuing with this anytime soon. If anyone else wants to give it a try I can give some pointers

PR #3668, opened 2025-10-01
Feat: allow the user to define the inclusion og the negative sign

alex-Arc:
continues on from #3651

### DIFF
--- a/docs/4_secondary_admin_controls/expressions/functions.md
+++ b/docs/4_secondary_admin_controls/expressions/functions.md
@@ -240,7 +240,7 @@ You can do the reverse of this with `secondsToTimestamp(str)`
 
 **secondsToTimestamp(seconds, format)**
 
-Convert a number of seconds into a timestamp of format 'HH:mm:ss'.
+Convert a number of seconds into a timestamp of format 'n:HH:mm:ss'.
 
 Note: There is no limit to the number of hours shown, it will display values greater than 24.
 
@@ -248,13 +248,14 @@ By supplying the format parameter, you can choose which components will be inclu
 
 The following components are allowed:
 
+- `N` / `n` - minus sign
 - `HH` / `hh` - hours
 - `mm` - minutes
 - `ss` - seconds
 
 **msToTimestamp(milliseconds, format)**
 
-Convert a number of milliseconds into a timestamp of format 'HH:mm:ss.SSS'.
+Convert a number of milliseconds into a timestamp of format 'n:HH:mm:ss.SSS'.
 
 Note: There is no limit to the number of hours shown, it will display values greater than 24.
 
@@ -262,6 +263,7 @@ By supplying the format parameter, you can choose which components will be inclu
 
 The following components are allowed:
 
+- `N` / `n` - minus sign
 - `HH` / `hh` - hours
 - `mm` - minutes
 - `ss` - seconds
--- a/shared-lib/lib/Expression/ExpressionFunctions.ts
+++ b/shared-lib/lib/Expression/ExpressionFunctions.ts
@@ -167,7 +167,7 @@ export const ExpressionFunctions: Record<string, (...args: any[]) => any> = {
 	secondsToTimestamp: (v, type) => {
 		const negative = v < 0
 		v = Math.abs(v)
-		type = type ? type : 'hh:mm:ss'
+		type = type ? type : 'n:hh:mm:ss'
 
 		const seconds = pad(Math.floor(v) % 60, '0', 2)
 		const minutes = pad(Math.floor(v / 60) % 60, '0', 2)
@@ -177,13 +177,14 @@ export const ExpressionFunctions: Record<string, (...args: any[]) => any> = {
 		if (type.includes('HH') || type.includes('hh')) timestamp.push(hours)
 		if (type.includes('mm')) timestamp.push(minutes)
 		if (type.includes('ss')) timestamp.push(seconds)
+		const showSign = negative && (type.startsWith('N') || type.startsWith('n'))
 
-		return (negative ? '-' : '') + timestamp.join(':')
+		return (showSign ? '-' : '') + timestamp.join(':')
 	},
 	msToTimestamp: (v, type) => {
 		const negative = v < 0
 		v = Math.abs(v)
-		type = type ? type : 'mm:ss.ms'
+		type = type ? type : 'n:mm:ss.ms'
 
 		const ms = v % 1000
 		const seconds = pad(Math.floor(v / 1000) % 60, '0', 2)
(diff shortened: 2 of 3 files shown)


docs/4_secondary_admin_controls/expressions/functions.md
```diff
@@ -240,28 +240,30 @@ You can do the reverse of this with `secondsToTimestamp(str)`
 
 **secondsToTimestamp(seconds, format)**
 
-Convert a number of seconds into a timestamp of format 'HH:mm:ss'.
+Convert a number of seconds into a timestamp of format 'n:HH:mm:ss'.
```
peternewman, line 243: Given this seems to produce -12:34:56 (rather than -:12:34:56), would this be clearer:
```suggestion
Convert a number of seconds into a timestamp of format 'nHH:mm:ss'.
```

docs/4_secondary_admin_controls/expressions/functions.md
```diff
@@ -240,28 +240,30 @@ You can do the reverse of this with `secondsToTimestamp(str)`
...
 
+- `N` / `n` - minus sign
 - `HH` / `hh` - hours
 - `mm` - minutes
 - `ss` - seconds
 
 **msToTimestamp(milliseconds, format)**
 
-Convert a number of milliseconds into a timestamp of format 'HH:mm:ss.SSS'.
+Convert a number of milliseconds into a timestamp of format 'n:HH:mm:ss.SSS'.
```
peternewman, line 258: Again:
```suggestion
Convert a number of milliseconds into a timestamp of format 'nHH:mm:ss.SSS'.
```

shared-lib/lib/Expression/ExpressionFunctions.ts
```diff
@@ -167,7 +167,7 @@ export const ExpressionFunctions: Record<string, (...args: any[]) => any> = {
 	secondsToTimestamp: (v, type) => {
 		const negative = v < 0
 		v = Math.abs(v)
-		type = type ? type : 'hh:mm:ss'
+		type = type ? type : 'n:hh:mm:ss'
```
peternewman: Again:
```suggestion
		type = type ? type : 'nhh:mm:ss'
```

shared-lib/lib/Expression/ExpressionFunctions.ts
```diff
@@ -177,13 +177,14 @@ export const ExpressionFunctions: Record<string, (...args: any[]) => any> = {
 		if (type.includes('HH') || type.includes('hh')) timestamp.push(hours)
 		if (type.includes('mm')) timestamp.push(minutes)
 		if (type.includes('ss')) timestamp.push(seconds)
+		const showSign = negative && (type.startsWith('N') || type.startsWith('n'))
```
peternewman: showNegative, given it won't show +12:34:56

shared-lib/lib/Expression/ExpressionFunctions.ts
```diff
@@ -177,13 +177,14 @@ export const ExpressionFunctions: Record<string, (...args: any[]) => any> = {
...
 
-		return (negative ? '-' : '') + timestamp.join(':')
+		return (showSign ? '-' : '') + timestamp.join(':')
 	},
 	msToTimestamp: (v, type) => {
 		const negative = v < 0
 		v = Math.abs(v)
-		type = type ? type : 'mm:ss.ms'
+		type = type ? type : 'n:mm:ss.ms'
```
peternewman:
```suggestion
		type = type ? type : 'nmm:ss.ms'
```

alex-Arc:
hey @peternewman thanks for taking a look, i'm still trying to find the right shape
a couple of things come up
- `HH` and `hh` seams to be the normal way to indicate 24/12 hour (should 12h clamp or loop and what about 24h? we could add days)
- I've added `a` as the token for AM/PM
- how strict should we be about requiring separators? the regex (/b{token}/b) can force the to be separators
- should we trim of anything bisides the timestamp?

arikorn:
I'm wondering if the conclusion in Issue #3698 , i.e. to add printf-style formatting might not be a more general and possibly simpler approach?  If nothing else, at least these two conversations should be cross-referenced!

alex-Arc:
> printf-style formatting

how would you transform a ms number into a 24-hour string with a printf string?

Julusian:
I wouldn't say that this could use printf, but maybe this should use an off the shelf library to do formatting? 
We are already using https://day.js.org/docs/en/display/format for formatting dates/times in the ui, perhaps the same should be used here?

alex-Arc:
> https://day.js.org/docs/en/display/format for formatting

yes I did look at this, but they do not support negative values

arikorn:
Sorry @alex-Arc, my bad. I should have written strftime (such as Julian's suggestion), but really if all you're doing is seconds to hh:mm:ss it's overkill.

One could ask whether more generalized formatting is desirable, and that's my point in cross-referencing these similar issues.

On that note, (i.e., cross-referencing) there appears to be at least two "generic" timer module, which already do the formatting. Not sure if that's a vote in favor of or against this PR 😀 :  [generic-stopwatch](https://github.com/bitfocus/companion-module-generic-stopwatch), and [highpass-countdown](https://github.com/bitfocus/companion-module-generic-stopwatch)

alex-Arc:
Ah ok.

I made this PR to be nonbreaking, aside having to add the `n` modifier to get there negative sign back
is it an options to break this ?

in regards to modules them self formatting, I'm not sure I understand your point?
Is it that all modules them self should manually generate all time formatting that could be needed as variables?

arikorn:
@alex-Arc  wrote:
>in regards to modules them self formatting, I'm not sure I understand your point?
Is it that all modules them self should manually generate all time formatting that could be needed as variables?

The point was mostly "academic" (i.e., to cross-index relevant information.) I am an information hound! 

To explain how it might be relevant, there are two sides to this coin:
1. At least two modules appear to do something similar to what you're proposing, as part of a timer package. If the purpose of this PR is specifically for timers, then perhaps you'd be better-served using one of these modules, which already provides the formatting and may provide powerful or flexible timers as well...

2. The benefit is really one-way (a user benefits from modules, modules do not benefit from expressions): Modules don't have easy access to expressions, at least not until the module API is updated to allow it, so this PR won't help them much, regardless. (So yes, although it wasn't my original intent to say so, your surmise is correct: module developers will still have to write their own hh:mm formatting. It's not really a big deal compared to the job of writing a module!)

As I wrote at the outset, I'm merely providing information, not making specific recommendations! I definitely see merit in what you're doing.

alex-Arc:
all good 😉

1. I do not believe that modules can expose their functions, so I don't see  how that existence of the timer modules have any overlap with expressions that the users themselves control 
2. here I do not agree, if there are good formatting expressions for the users to use then there is no need for the module to do formatting so it doesn't matter that the module can't use the expressions.
Working on the [ontime](https://github.com/bitfocus/companion-module-getontime-ontime) module i can say that you can't really provide all the ways of formatting that the user might what, so I believe it is better for them to have the option themselves. Also the amount of time values in ontime is quite large if that was multiplied by 6 to cover the most used time formatets then the amount of variables would become quite excessive

I see this as a small improvement to an already existing feature

Julusian (approved):

alex-Arc:
@Julusian happy to see this merged, anything you would like cleaned or changed about it?

Are there any thoughts on the rounding?
In the case where milliseconds are show it makes sense to floor all the other values.
But if the smallest unit is seconds it looks more correct to either do a `round` or `ceil` so that the seconds value ticks over on the actual zero crossing

Issue #3669, opened 2025-10-02
[BUG] Preset `previewStyle` is overridden by feedback


arikorn:
### Make sure you're on the latest stable or beta build

- [x] I have tested this on the latest stable or beta release

### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion and not a specific module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When defining module presets, the `previewStyle` property customizes the display in the Presets tab -- presumably so that current state doesn't change the look of the button and/or so the Preset more clearly describes the button's function.  For example: a toggle button may have text "Play" by default and "Pause" (with a color change) when the feedback "isPlaying" is true. But showing either "Play" or "Pause" in the Presets tab can be confusing, so instead the `previewStyle` can set the text to "Play/Pause". This currently works as long as the feedback isn't true, but if it is, the feedback overrides the  `previewStyle` .

### Steps To Reproduce

1. In a module define a preset with `style`,  `previewStyle` (that overrides `style`) and a feedback.
2. Look in the Presets tab in the Admin UI
3. With the (boolean) feedback false, the  `previewStyle`  values will show
4. Do whatever changes the feedback to true

### Expected Behavior

The feedback should not change the button in the Presets tab. (But it currently does.)

### Environment (please complete the following information)

```markdown
- OS: Windows 10
- Browser: Chrome, current version
- Companion Version: 4.2=8448 beta
```

Julusian:
I dont think this is a bug

The `previewStyle` field was added before presets were capable of drawing with feedbacks, I believe as a workaround for the limited dynamic drawing ability of presets, often resulting in them drawing badly/non-representive of when they are placed onto a button. I believe that even variables wouldnt parse in button text fully (certainly not reactively), so the previewStlye would let you override that with predictable placeholder text instead of an unpredictable static string

So when adding support for feedbacks to be drawn, it felt natural for them to override even the previewStyle, as the feedbacks are an override and will behave that way once the button is placed.  



Im curious if there is an example of why you dont want this behaviour?

arikorn:
[...]

PR #3673, opened 2025-10-04
Fix: Preset previewStyle is overridden by feedback

arikorn:
Fixes Issue #3669

This is what I meant by my "proposal" in the comments and I think it's also what [you are suggesting](https://github.com/bitfocus/companion/issues/3669#issuecomment-3366580776), but if not then it's a point of departure for discussion.

While we could try to be more clever about how much of the feedback one is overriding, it will generally fail with "advanced" feedbacks, since the style information is missing then.

### DIFF
--- a/companion/lib/Instance/Definitions.ts
+++ b/companion/lib/Instance/Definitions.ts
@@ -282,6 +282,10 @@ export class InstanceDefinitions extends EventEmitter<InstanceDefinitionsEvents>
 			style: definition.previewStyle ? convertPresetStyleToDrawStyle(definition.previewStyle) : definition.model.style,
 			steps: {},
 		}
+		if ('previewStyle' in definition && definition.previewStyle !== undefined) {
+			// don't let feedbacks override the previewStyle.
+			result.feedbacks = []
+		}
 
 		// Omit actions, as they can't be executed in the preview. By doing this we avoid bothering the module with lifecycle methods for them
 		for (const [stepId, step] of Object.entries(definition.model.steps)) {


arikorn:
@Julusian, let me think about this more... or better yet, please weigh in with your thoughts! 

Meanwhile, two items, regardless:
1. I noticed that this proposal will affect your Satellite module. Does this make things "worse"?
2. How can I use a dev module with my local core development builds? I've tried running the launcher using the Wiki instructions but in v4.1 it doesn't draw the settings-window contents and even in v4.0, though I can set the dev module path, the "Launch" button doesn't do anything! (all this in WSL, and for this particular case, using the latest core version would be best, of course).  

Thanks!

Julusian:
> I noticed that this proposal will affect your Satellite module. Does this make things "worse"?

Yeah, I was thinking that something like this would occur, but couldnt think where (I couldnt think of many uses of previewStyle).

For that module, it really needs the 0/0 it adds in the previewStyle so that you can tell how the buttons are supposed to be arranged, and the feedback is also really helpful so if you are looking for a specific button then you can find it easily.

So I think it would be better to:

>  I imagine that it shouldnt disable feedbacks in your proposal, but just let the previewStyle overrule it

Meaning, keep the presets, but merge the previewStyle on top of the style, after the presets.  
I now realise that this isnt a simple thing to do...

Perhaps this will have to be done by adding an internal feedback to the end of the feedbacks array, which uses the internal 'check expression' feedback, with a hardcoded `true` expression, and the style values from previewStyle






----

>  it doesn't draw the settings-window contents 

There is another dev command needed for that, to run the server for that ui. I think its `yarn dev` inside the launcher-ui folder.

> the "Launch" button doesn't do anything! 

I am not surprised; electron will be doing a linux system thing to open a browser, and presumably it doesnt find a browser installed inside wsl.

> How can I use a dev module with my local core development builds? 

Personally I wouldnt use the launcher for this, I dont think that is setup to restart whenever core code changes.  
If running headless, the `module-local-dev` folder will be used for loading dev modules from, or you can override that with either the `--extra-module-path` argument (I think), or by setting the env var COMPANION_DEV_MODULES to a path (this is what I do, I set it up in my bash profile once and it then works for every clone/checkout of comapnion on my machine)

arikorn:
Thanks for the developer tips Julian, that will be super helpful!

I had a similar idea regarding advanced feedbacks, but maybe simpler: Advanced feedbacks don't have a `style:` property (or more precisely it's set to `undefined`), so I can stick the `previewStyle` in the `style` property. I just checked it (thanks again to your "dev module" advice) and it seems to work as expected with no further code changes. What do you think? ... I'll push soon after a little more testing.

arikorn:
OK, this should be "correct" now.  

Note: I didn't use `convertPresetStyleToDrawStyle` to copy `previewStyle` because it adds properties that the user didn't specify, so didn't seem appropriate for merging fields here.  (Also, without `Object.assign()` Typescript wasn't very happy...)

Julusian:
hmm.. I havent test it but I don't think that will work how you want it to.

The style property on advanced feedbacks wont be used, so this wont work with them.

> Perhaps this will have to be done by adding an internal feedback to the end of the feedbacks array, which uses the internal 'check expression' feedback, with a hardcoded true expression, and the style values from previewStyle

I still think that this is the way to go. should be as simple as pushing a new object to the array of feedbacks, with fields setup appropriately. Not sure how well the style aligns, but shouldnt be too hard to change that.

Then the merge onto the base style can also be removed, as the feedback will always override it

arikorn:
> hmm.. I havent test it but I don't think that will work how you want it to.

But it does! I tested it before pushing.  And if you inspect at `getUnparsedFeedbackStyle` or the like, you can confirm that it's returning the merged value.

Your way is less intrusive but for some reason it works differently (it's technically more correct). The problem arises from the fact that I don't want `previewStyle` to set the text, since my advanced feedback sets the text, but `CompanionButtonStyleProps` requires me to put in a text property. So... with your way the added feedback blanks out the text (as would be expected from the definition). With my way, for reasons I can't explain, it does not.  I suppose a proper solution would be to make text (and all properties, really) optional in `previewStyle`.

And, yes, I tested it and ~can push once we~ would like to resolve this minor issue...

Could I just change `PresetDefinitionButton` in *Presets.ts* from
~~~TypeScript
	previewStyle: CompanionButtonStyleProps | undefined
~~~
to
~~~TypeScript
	previewStyle: Partial<CompanionButtonStyleProps> | undefined
~~~

this actually makes it more consistent with `FeedbackEntityModel`, FWIW.

So  bottom line: I'd prefer your suggestion with this one change to the previewStyle definition. Maybe I'll just push it so you can see what I'm doing in context...

arikorn:
Purely academic: FWIW, I suspect that in my second version the previewStyles added to the "advanced" feedback are applied before the advanced feedback is applied. That would explain why the colors changed (not in the advanced feedback) but the previewStyle text was overwritten by the advanced feedback.  Again: purely academic; take 3 (pushed my yesterday) should be doing things the way you suggested (and I agree it's better).

Julusian:
> And if you inspect at getUnparsedFeedbackStyle or the like, you can confirm that it's returning the merged value.

but inside there, the only place I can see it is reading from the `style` property of the feedback is inside of:
```
switch (definition.feedbackType) {
			case FeedbackEntitySubType.Boolean:
```
so I don't see how setting the style property on an advanced feedback would be doing anything.
If it is doing something, then there is a bug somewhere as it should not be.

arikorn:
Well, FWIW, I couldn't stop myself from delving, so here's my analysis;

In [`applyComplexStyle`](https://github.com/bitfocus/companion/blob/de3df0a5afd50d4b39e2f546feb3db297d31446a/companion/lib/Controls/Entities/FeedbackStyleBuilder.ts#L44)
~~~TypeScript
			this.#combinedStyle = {
				...this.#combinedStyle,
				...prunedValue,
			}
~~~
does what I deduced above from the behavior: `this.#combinedStyle` contains the merged style, including the parts injected from `previewStyle` in my ["take 2" commit, above](https://github.com/bitfocus/companion/pull/3673/commits/41a37276d77705deb75d4de63bc4813926ebef27). , Notice that`prunedValue` (essentially the output of the advanced feedback), is last in the list, so properties explicitly returned by the advance feedback override the combinedStyle.

If you trace the origins of `this.#combinedStyle` (for presets), I think it's coming from `ControlButtonPreset.#applyPresetModel()` in Presets.ts.
~~~TypeScript
	#applyPresetModel(storage: PresetButtonModel): void {
		this.#baseStyle = Object.assign(this.#baseStyle, storage.style || {})
~~~

Anyway, I wouldn't call it a bug, since normally style *is* undefined for advance feedbacks. (My second commit here being the exception.)

Thanks, as always, for the merge!

PR #3683, opened 2025-10-07
fix: Page up button/Page down button don't set page-history

### LINKED ISSUE #3678: [BUG] Page up button/Page down button don't set page-history

### Describe the bug

The action "internal: Surface: Set to page" has an option for "back" and "forward" which references `class InternalSurface` property `#pageHistory`.  However, the built-in Page up button/Page down button buttons don't call `changeSurfacePage`, so are not included in the page history.

This results in the possibility of getting "stuck" on a page if the user got to the page using a Page up/Page down button, but the target page doesn't have those buttons, but instead is relying on "Set to page" back to get out...

### Steps To Reproduce

1. Create two pages.
2. On one of them add a button with action "internal: Surface: Set to page" set to "back"
3. On a surface, use the "built-in" page up/page down arrows to go away from and back to the page
4. Press the button made in step 2. Nothing happens

### Expected Behavior

Expected behavior: it should go back to the previous page.

```markdown
- OS:Windows 10
- Browser: Chrome
- Companion Version: 4.1.1
```

### Additional context

One could add a call to `this.#surfaceController.on('surface_page',` (all in *companion/lib/Internal/Surface.ts*) but when I tried it, it led to an infinite loop. One could resolve the infinite loop by adding an optional "alreadyDone" arg to `changeSurfacePage` but no doubt there's a more elegant way to do.  As always, I'm happy to either submit a fix this way or follow a hint, if you'd like me to take a stab at it.

Note that history is already limited to 100 entries, so there's no harm in recording the up/down buttons.

arikorn:
Resolves #3678

This was a bit of a puzzle that I couldn't put down but put simply: `ControlButtonPageDown` and `ControlButtonPageUp` take very different paths from normal buttons (`ControlButtonNormal`) starting from `ControlsController.pressControl`.  The paths don't merge until `SurfaceGroup.setCurrentPage` so getting PageUp/Down buttons to register their history would have to take place over there (i.e. in `SurfaceGroup`).   Upon reflection, it actually make sense to keep history in the SurfaceGroup object, so it's hopefully a happy "coincidence."

Bottom line: I moved the function that maintains page history into *Groups.ts* ~`SurfaceGroup.changePage`~ `SurfaceGroup.setCurrentPage` from `InternalSurface#changeSurfacePage` and then made some fairly minor changes to make it the main point of entry for changing pages.

Just to keep the mods clear, I left some original code in comments in ~`SurfaceGroup.changePage`~ `SurfaceGroup.setCurrentPage` These comments can be deleted before merging, if you prefer.

I hope you agree with my choices!  It works correctly in a basic test of mixing presses of "back" and PageDown/PageUp...

### DIFF
--- a/companion/lib/Internal/Surface.ts
+++ b/companion/lib/Internal/Surface.ts
@@ -545,13 +545,6 @@ export class InternalSurface extends EventEmitter<InternalModuleFragmentEvents> 
 		}
 	}
 
-	incrPage(surfaceId: string, forward: boolean): boolean {
-		//const surfaceId = this.#fetchSurfaceId({controller: surfaceId}, undefined as unknown as RunActionExtras,  false)
-		if (!surfaceId) return true
-		this.#changeSurfacePage(surfaceId, forward ? '+1' : '-1')
-		return true
-	}
-
 	executeAction(action: ControlEntityInstance, extras: RunActionExtras): boolean {
 		if (action.definitionId === 'set_brightness') {
 			const surfaceId = this.#fetchSurfaceId(action.rawOptions, extras, true)
--- a/companion/lib/Surface/Group.ts
+++ b/companion/lib/Surface/Group.ts
@@ -42,6 +42,11 @@ export class SurfaceGroup {
 	#currentPageId: string
 
 	/**
+	 * Page history for surfaces
+	 */
+	readonly #pageHistory: { history: string[]; index: number }
+
+	/**
 	 * The surfaces belonging to this group
 	 */
 	surfaceHandlers: SurfaceHandler[] = []
@@ -149,6 +154,8 @@ export class SurfaceGroup {
 			this.groupConfig.startup_page_id = this.#currentPageId
 		}
 
+		this.#pageHistory = { history: [this.#currentPageId], index: 0 }
+
 		// Now attach and setup the surface
 		if (soleHandler) this.attachSurface(soleHandler)
 
@@ -214,17 +221,6 @@ export class SurfaceGroup {
 	}
 
 	/**
-	 * Perform page-down for this surface group
-	 */
-	doPageDown(): void {
-		if (this.#userconfig.getKey('page_direction_flipped') === true) {
-			this.#increasePage()
-		} else {
-			this.#decreasePage()
-		}
-	}
-
-	/**
 	 * Set the current page of this surface group
 	 */
 	#setCurrentPageInternal(newPageId: string, defer = false): void {
@@ -238,6 +234,17 @@ export class SurfaceGroup {
 	 */
 	getCurrentPageId(): string {
 		return this.#currentPageId
+	}
+
+	/**
+	 * Perform page-down for this surface group
+	 */
+	doPageDown(): void {
+		if (this.#userconfig.getKey('page_direction_flipped') === true) {
+			this.#increasePage()
+		} else {
+			this.#decreasePage()
+		}
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


arikorn:
Not sure if this will help illustrate the issue.  Here are the two call-trees side-by-side and aligned to show commonalities and differences. (The original location of the page-history code is highlighted, showing how it happens before the two call trees re-merge.)
image

arikorn:
I just noticed another inconsistency that may need to be resolved: the old `setCurrentPage` function resulted in a noop if a non-existent page was specified. The old  `#changeSurfacePage` (now named `setCurrentPage` and move to SurfaceGroup) would instead redirect to the first page under the same conditions. We should probably choose one or the other regardless of this PR (by default in this PR, we're choosing to go to page 1 in all cases).

companion/lib/Internal/Surface.ts
```diff
@@ -550,6 +545,13 @@ export class InternalSurface extends EventEmitter<InternalModuleFragmentEvents>
 		}
 	}
 
+	incrPage(surfaceId: string, forward: boolean): boolean {
```
Julusian: It seems that this method is unused?

companion/lib/Surface/Group.ts
```diff
@@ -252,15 +252,80 @@ export class SurfaceGroup {
...
-		const newPageId = this.#pageStore.getOffsetPageId(this.#currentPageId, -1)
-		if (!newPageId) return
-		this.setCurrentPage(newPageId)
+		this.setCurrentPage('-1')
+	}
+
+	/**
+	 * Page history for surfaces
+	 */
+	readonly #pageHistory = new Map<string, { history: string[]; index: number }>()
```
Julusian: This could be simplified, it doesn't need to be a map anymore now that it lives inside of the group
```suggestion
	readonly #pageHistory: { history: string[]; index: number }
```

companion/lib/Surface/Group.ts
```diff
@@ -252,15 +252,80 @@ export class SurfaceGroup {
...
+	/**
+	 * Change the page of a surface, keeping a history of previous pages
+	 */
+	setCurrentPage(toPage: string | 'back' | 'forward' | '+1' | '-1', defer = false): void {
+		const groupId = this.groupId //this.#surfaceController.getGroupIdFromDeviceId(surfaceId)
+		if (!groupId) return
+
+		const currentPage = this.#currentPageId //this.#surfaceController.devicePageGet(groupId, true)
+		if (currentPage === undefined) {
+			// Bad groupId
```
Julusian: I think it would be good to remove this case now. In most cases (all except +1 or -1) an invalid/missing currentPageId shouldnt cause problems.
The behaviour of getOffsetPageId might even make this fine. If it does change to page1 when currentPage is bad I think that is ok. better that than being stuck in a broken/invalid state

companion/lib/Surface/Group.ts
```diff
@@ -252,15 +252,80 @@ export class SurfaceGroup {
...
+		} else {
+			// no history yet
+			// start with the current (from) page
+			let pageHistory = this.#pageHistory.get(groupId)
+			if (!pageHistory) {
+				pageHistory = {
+					history: [currentPage],
+					index: 0,
+				}
```
Julusian: this should move to the constructor now, there is no need to lazily define it anymore

arikorn:
Thanks for your review Julian!  Yeah, it's true that in PR's like this I aim for the absolute minimum change, and so don't consider bigger-picture items...  so I'm glad that you did!

Two minor changes:
1. I moved `doPageDown()` to be next to `doPageUp()` for improved readability. The code is otherwise untouched.

2. I deleted `groupId` assignment and test, since it's no longer needed. Pretty sure you'll agree with that one, but to be safe I added it to the review for your approval.

I think that should do it!

Julusian:
I have a couple more thoughts on bits that could be tidied/simplified now, but I won't burden you with doing them as they are largely stylistic and partially changing pre-existing style/flow choices
2234efd18f52abe906380c8b9be9c4b9ca296da8

arikorn:
Thanks Julian.  At least I can address the question you put in the subsequent commit:

[// TODO - should this search for the first valid pageId?](https://github.com/bitfocus/companion/commit/2234efd18f52abe906380c8b9be9c4b9ca296da8#diff-2c8cc586e1e7654939861ab7a7f151fbdf3dbb47c1678f3d8a6213fbfa0ea3b1R263)

My suggestion is: "not as written," i.e. it's good as is, because the typical reason `pageTarget` is undefined is that we ran out of history, so `pageIndex` is oob, for example, if going 'back' from the first item in the history: `pageTarget = pageHistory.history[-1]`. As currently written, this leaves history at index 0 (or length - 1, if going forward) and that makes sense (at least to me!). If  you wanted to be very precise, I suppose you could differentiate between "ran out of history" and "a page on the history somehow disappeared"?, but that seems pretty rare. Then again, for clearer documentation, maybe that is the way to go? or just leave a comment to this effect?

PR #3685, opened 2025-10-08
Fix: Surface Groups should always show "Current Page"

arikorn:
Currently a standalone surface will show and allow the user to adjust the current page but a group won't unless "Use Last Page At Startup" is true. This PR fixes it so groups behave similarly to standalone surfaces. 

(It's  useful in an emergency case - such as user accidentally went to a page with no back arrows -- typically during page-layout development...)

### DIFF
--- a/webui/src/Surfaces/EditPanel.tsx
+++ b/webui/src/Surfaces/EditPanel.tsx
@@ -362,7 +362,7 @@ const SurfaceEditPanelContent = observer<SurfaceEditPanelOldProps>(function Surf
 								</CFormLabel>
 								<CCol sm={8}>
 									<InternalPageIdDropdown
-										disabled={surfaceId !== null && !surfaceInfo?.isConnected && !groupConfig.config.use_last_page}
+										disabled={false}
 										includeDirection={false}
 										includeStartup={false}
 										value={groupConfig.config.last_page_id}


webui/src/Surfaces/EditPanel.tsx
```diff
@@ -355,14 +355,14 @@ const SurfaceEditPanelContent = observer<SurfaceEditPanelOldProps>(function Surf
...
+						{(surfaceId === null || !!surfaceInfo?.isConnected || !!groupConfig.config.use_last_page) && (
 							<>
 								<CFormLabel htmlFor="colFormCurrentPage" className="col-sm-4 col-form-label col-form-label-sm">
-									{surfaceInfo?.isConnected ? 'Current Page' : 'Last Page'}
+									{!groupConfig.config.use_last_page ? 'Current Page' : 'Last Page'}
 								</CFormLabel>
 								<CCol sm={8}>
 									<InternalPageIdDropdown
-										disabled={false}
+										disabled={surfaceId !== null && !surfaceInfo?.isConnected && !groupConfig.config.use_last_page}
```
Julusian: I can't see when this is ever true. when have you found?

Julusian: I asked a second opinion :shrug: 

image

I'm going to undo this so that I can merge this into 4.1.2

arikorn: Yeah, I probably went through several variations of hiding and disabling.  Looks good and thank you, as always, for the fixes and the merge!

PR #3692, opened 2025-10-09
Add MQTT remote API

cgarwood:
Adds an MQTT remote API to Companion providing live updates of all Companion variables and allows pressing buttons and setting custom variables.

I used the Ember+ module as a reference, and tried to align syntax for commands with the other modules as closely as possible.

Uses the same `mqtt` dependency as the Generic MQTT Companion Module

This is my first PR against Companion core. I wasn't 100% sure how to handle the yarn.lock file for the PR.

### DIFF
--- a/companion/lib/Service/Mqtt.ts
+++ b/companion/lib/Service/Mqtt.ts
@@ -6,7 +6,7 @@ import mqtt from 'mqtt'
 
 const DEBOUNCE_REINIT_DELAY = 5000
 
-export class MqttService extends ServiceBase {
+export class ServiceMqtt extends ServiceBase {
 	readonly #serviceApi: ServiceApi
 
 	readonly #releaseTime = 20
--- a/companion/lib/Service/Controller.ts
+++ b/companion/lib/Service/Controller.ts
@@ -11,7 +11,7 @@ import { ServiceSatelliteTcp } from './SatelliteTcp.js'
 import { ServiceSurfaceDiscovery } from './SurfaceDiscovery.js'
 import { ServiceTcp } from './Tcp.js'
 import { ServiceUdp } from './Udp.js'
-import { MqttService } from './Mqtt.js'
+import { ServiceMqtt } from './Mqtt.js'
 import { ServiceVideohubPanel } from './VideohubPanel.js'
 import type { UIHandler } from '../UI/Handler.js'
 import { ServiceSatelliteWebsocket } from './SatelliteWebsocket.js'
@@ -47,7 +47,7 @@ export class ServiceController {
 	readonly tcp: ServiceTcp
 	readonly udp: ServiceUdp
 	readonly emberplus: ServiceEmberPlus
-	readonly mqtt: MqttService
+	readonly mqtt: ServiceMqtt
 	readonly artnet: ServiceArtnet
 	readonly rosstalk: ServiceRosstalk
 	readonly satelliteTcp: ServiceSatelliteTcp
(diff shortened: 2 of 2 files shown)


companion/lib/Service/Mqtt.ts
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	private startEventListeners(): void {
+		// Listen for variable changes
+		this.#serviceApi.on('variables_changed', this.variablesChanged)
+	}
+
+	private stopEventListeners(): void {
+		this.#serviceApi.removeListener('variables_changed', this.variablesChanged)
+	}
```
Julusian, line 105: I'm not keen on where these are being called from, as it makes it hard to tell when they are being called and to ensure that they are called the correct number of times.

I think it would be easier to follow if they were setup in just the constructor of the class. And teardown might not be necessary (ember+ doesnt do it), as we currently instantiate these at startup and keep them until exit.  
The cost of the methods firing when there is no client connected is not enough to worry about I expect, perhaps the `variablesChanged` method needs a 'is connected' check too?

companion/lib/Service/Mqtt.ts
```diff
@@ -0,0 +1,201 @@
...
+	private variablesChanged = (variables: Set<string>): void => {
+		if (this.#client === null) return
+
+		for (const variable of variables) {
+			let [label, name] = variable.split(':')
+
+			if (label == 'internal' && name.startsWith('custom')) {
+				label = 'custom'
+				name = name.substring(7)
+			}
```
Julusian, line 116: I don't think this block will do anything, while there are some internal aliases of `internal:custom_*` to `custom:*`, that is all handled at the point of reading the values, they should never be reported as being changed (I think)

companion/lib/Service/Mqtt.ts
```diff
@@ -0,0 +1,201 @@
...
+
+			let value = undefined
+
+			if (label === 'custom') {
+				value = this.#serviceApi.getCustomVariableValue(name)?.toString()
+				if (value === undefined) return
```
Julusian, line 122: You can skip this block. While technically more correct, its a few levels of indirection that result in the same thing as calling `this.#serviceApi.getConnectionVariableValue('custom', name)`.

It exists primarily to allow the caller to not need to know that the label is 'custom', but you dont need to worry about that as you already have the label and name pair

companion/lib/Service/Mqtt.ts
Julusian: I now see that the ember+ service was doing this already, I shall look into removing it from there too

companion/lib/Service/Mqtt.ts
```diff
@@ -0,0 +1,201 @@
...
+			} else {
+				value = this.#serviceApi.getConnectionVariableValue(label, name)?.toString()
+				if (value === undefined) return
```
Julusian, line 125: I suspect that you don't want to do this check, as when a variable gets removes/unset then you wont be reporting that to clients. 

In ember+, this would usually manifest as the node going away, but I dont think you can do the same, so should probably report the variable becomign empty

companion/lib/Service/Mqtt.ts
```diff
@@ -0,0 +1,201 @@
...
+			}
+
+			const topic = `${this.topic}/variables/${label}/${name}`
```
Julusian, line 128: Just thinking about future proofing, it might be better to do `${this.topic}/variables/${label}/${name}/value`, so that the description or type or something could be added as separate properties later?

Or maybe the message should be a small json blob:
```
{
  "value": "abc"
}
```
to allow for other details to be added with it?

companion/lib/Service/Mqtt.ts
```diff
@@ -0,0 +1,201 @@
...
+						this.logger.error(`Failed to subscribe to MQTT topic ${actionsTopic}: ${err.message}`)
+					}
+				})
+
+				// Handle incoming messages
+				this.#client!.on('message', (topic, message) => {
+					this.handleMessage(topic, message)
+					this.logger.debug(`Received MQTT message on topic ${topic}: ${message.toString()}`)
+				})
+			})
```
Julusian, line 70: should any current values be published? having to wait for them to change could be problematic at times?

cgarwood:
Thanks for the review! May be a week or so but I'll take a look and make adjustments 🙂

cgarwood:
Hate to do this after you took the time to review, but I may have bitten off a little more than I can chew at the moment. Going to close this for now so it's not just sitting in limbo, and will submit a new PR if/when I get more free time to dedicate to it again.

Issue #3698, opened 2025-10-10
Feature: Number formatting for variables


arikorn:
### Is this a feature relevant to companion itself, and not a specific module?

- [x] I believe this to be a feature for companion, and is not specific to a module

### Is there an existing issue for this?

- [x] I have searched for similiar existing issues

### Describe the feature

It would be nice to be able to specify a format for numeric variables.

The simplest one would be to specify the number of decimal places: For example: `$(module:SomeNumber:2.1)` would specify 2 places before and one place after the decimal (other options: .1 or 0.1 to let left float free; -.1 to add space for negative sign;  +.1 to always show + or -)

Fanciest would be full `prinft`-style formatting

This is similar to #3664, which asked specifically about time variables, but more general (and the "simple" solution may not solve that one.)

### Usecases

When changing, for example, audio fader settings in increments of 0.5, with text centered, the number jumps back and forth depending on whether it is round ('1.0') or  not ('1.5')

Current behavior:

![Image](https://github.com/user-attachments/assets/4c405ccf-bbde-4686-8458-be587effd533)

Here's how it should look (done with an awkward expression that would need to be rewritten for each variable!)

![Image](https://github.com/user-attachments/assets/099d8772-7764-4154-a86d-1e8d890f40a2)

arikorn:
OK, I see now that the "simplest version", mentioned above, can be done with `toFixed()` in an expression, but that still requires expressions and a fair bit of skill. So for example (is 4.1 when buttons started supporting expressions for text?):

\`mic1:\n${toFixed($(gostream:AudioFader_mic1), 1)}\`  -- ("Expression Mode")
[edit: ]
or 
concat('mic1:\n', toFixed($(gostream:AudioFader_mic1), 1)) -- slightly simpler to write, though longer
{end edit]

Instead of:
mic1:\n$(gostream:AudioFader_mic1:0.1) -- ("Normal" mode, proposed enhancement)

Still probably worth doing this proposed enhancement -- I had a difficult time getting the expression just right. (Maybe we should add parenthesis matching in expression inputs???)

thedist:
Number formatting does need an cleaner way to do it, but I think there could be much cleaner and more intuitive ways to go about it.

For example, while `toFixed` is somewhat verbose in the need for lots of extra syntax in the whole button string, it's clear what it is and for many it's intuitive as to how it works.

[...]

PR #3699, opened 2025-10-12
Feat: Add a string-concatenation operator to expressions

arikorn:
Title was: Allow the `+` and `+=` expression operators to act on strings

So, for example: `'Fader: ' + toFixed($(module:faderValue),1)` returns 'Fader: 1.0' rather than N/A

Note: This implementation will ignore trimmable spaces, so the user cannot force it to be a string-op by starting with space or empty string, i.e.

   `' ' + 7 + 5` equals 12 (number) and not '75' (string)

It could be fixed with a bit of effort, but is probably rare enough to ignore?

This PR was inspired by discussions in Issue #3698 
(Also, I'm assuming that the idea was merely overlooked rather than intentionally omitted... though if the later, I'd encourage reconsidering..)


Julusian:
My immediate thought is that I am a bit concerned about this potentially being a wide reaching breaking change.

Although you may have mitigated that by trying it as a number first, but that will make this quirky and a little hard to explain/predict. (I might be thinking pretty edge case, but there will be bug reports/questions eventually)

Perhaps there is another currently unused operator that we could use instead? Im not sure if that is a better or worse idea

arikorn:
@Julusian , you know it's funny: I started by looking for a [different operator for string concatenation](https://en.wikipedia.org/wiki/Comparison_of_programming_languages_(strings)). You could choose one from there, if you insisted, but it seems to me that by far the most common operator is `+`. Expression operators leave very few unused options.

More to the point: **here's the current behavior in Companion v4.1.3** compared to this PR, only showing where they differ for ease of analysis: The only differences I could find were: (1) the intended effect of concatenating numbers and strings and (2) if a variable is set to NaN. In this case, I actually prefer what this PR does, since it makes it easier to see where the NaN arose in the expression.

So bottom line: I don't understand how this would break anything, (In modules, it would still evaluate to NaN, i.e. `parseVariablesInString` always returns a string, so `Number('1a1')` is NaN.)

image

[edit: if  you want to get crazy:
* 1.2 + true = 22 in both cases
* 1.2 + {a: 1} = NaN currently, '1.2[objectObject]' in this PR
* 1.2 + [3,4,5] = NaN currently, '1.23,4,5' in this PR
 I imagine some of those can be addressed, if needed.]

That said, my testing just now reveals somewhat buggy behavior of expressions (corresponding to asterisks in the image).
1. Typing NaN as a literal in the button text input is interpreted as an identifier, which is `undefined`, rather than `NaN`. (That's why this PR is the same as v4.1.3 in the case of `1.2 + NaN` -- the 6th example -- rather than returning "1.2$NA", i.e. `1.2 + undefined` = `NaN`). Note that it works correctly in the action `internal: Custom Variable: Set with expression` (but not in the Custom Variable page, which is yet another issue).
2. If Companion variable or literal is null it resolves to 0.
3. When a variable value is, in fact, NaN, the expression sees it as the string "$NA" (no doubt coming from some JSONify). Not sure whether this is a problem, though, but easily fixed. (In my mind $NA means "not available" rather than "not a number".)
4. Custom variables in "JSON" mode doesn't accept expressions. On "invalid" input sometimes it reverts to the previous value, sometimes to null. Examples: type NaN, it goes to null, not $NA, type 1/2 it becomes 1, (I can't get it to revert to previous value now.)
5. (Maybe another PR for me: typing Enter/Return while in the input for a custom variable doesn't update its value, nor is it updated interactively as is done with button text. I'd probably prefer needing Enter in either case but would mess with button text.)

arikorn:
In looking over my code, I realized that the above can be stated both more "logically" and succinctly:
The current code is: 
~~~TypeScript
    return Number(left) + Number(right)
~~~
This PR changes the return value  ***only*** if that value would have been NaN. i.e.
~~~TypeScript
    const result = Number(left) + Number(right)  // <--- what the current code returns
    return isNaN(result) ? left + right : result // <--- only change the return value if current code returns NaN
~~~

By definition, therefore, (a) currently "valid" numeric operations guaranteed to be identical and (b) anything that tries to convert the return value to a number will still get NaN.

(And because of the bug mentioned in note 1, above, there is even an edge-case in which even this PR returns NaN, still identically to the current behavior.)

Julusian:
A case that I am afraid of is someone doing `$(custom:a) + '\n' + $(custom:b)`. These both contain strings, and always strings. So this works fine. Then at some point, those strings happen to be two strings which are just numbers, perhaps `0` and `1`.
Then it suddenly does not do what they want at all, and the only way to fix it is to go back to using `concat($(custom:a), '\n', $(custom:b))`.  
Which means that this change can tidy up some expressions, it can also leave lurking bugs that require resorting to the old ways to resolve.

Maybe this case could be handled too, but I am worried we will end up continually patching these edge cases and make a complex mountain of stuff trying to replicate almost how js behaves, but slightly different.

arikorn:
@Julusian wrote:
>A case that I am afraid of is someone doing `$(custom:a) + '\n' + $(custom:b)`. These both contain strings, and always strings... Then at some point, those strings happen to be two strings which are just numbers, perhaps 0 and 1.

I think a question to ask is: do the UX benefits of simplifying string concatenation outweigh the rare case it might introduce new confusion? I would argue that your edge case is "very unlikely" because:

(A) The common usage is to start labels with a string, such as: `"Mic " + $(custom:a) + '\n' + $(custom:b)`, which will always resolve as expected. and

(B) Looking at images shared on the Facebook group, I could not find a single button that just had two numbers separated by whitespace, whereas many of them mixed text and numbers such as "Cam 1". My personal experience with the current options for string concatenation, (shared in 3698) found it to be frustrating. So from my perspective the benefits of this PR solving a very common issues outweigh it's apparently rare potential for confusion.

And conversely, of course, your case of a string variable changing to a number variable would be just as confusing today if you reversed it: a person writes  `$(custom:a) + $(custom:b)` expecting "a" and "b" to be "numerical" variables but then at some point, one of those variables happens to be alphabetic, they'll get NaN and the only way to fix it is to add some conditional logic, etc. (Note that this is technically more likely than your case in which *both* of the variables have to change!) Put simply: complex situations will invariably require complex solutions, regardless.

That said, if you really want to go with another operator, maybe `..` (or give up bitwise operators and use `&` -- probably too confusing though)?

Or, we can address the special case of whitespace separators. Or keep it as is, and wait for someone to complain? At the very worst, I see any special cases being addressed inside `resolve()`, so maybe a molehill rather than a mountain? 😁  But I do agree with you that had Companion started off with just the  Javascript `+` behavior this exercise would have been unnecessary (though maybe other more common issues would have arisen, if numbers usually show up as strings...)

arikorn:
~~OK, another idea: use no operator (there is precedent in the Wikipedia page), but this will require an initial operator to differentiate it from "compound" operations (since the parser doesn't distinguish between space and newline, for example), so, for example, if the first token is an ampersand (or plus sign? or empty string?...), concatenate the rest:~~

~~`& $(module:mic) '\n' $(module:volume)`~~

~~It's actually pretty clean... (and '+' could be optional and ignored... or make it "either-or": plus works with ampersand and if your special edge-case comes up, the user can fix it by prepending the ampersand.)~~

~~Hmm... in fact, Companion is already very close to this (undocumented?):~~

~~`concat( $(module:mic) '\n' $(module:volume))`~~

~~is a valid expression...~~

[Edit] Nevermind. It's very enticing, but in the end will likely be very confusing without a fairly major effort: for example, something like: `& $(module:mic) + $(module:volume)` becomes ambiguous and/or confusing. (Default would be to add the two if they're numbers,) and the parser can't distinguish  `& 'a' (1 + 2)`  from a function call, without additional punctuation, i.e. `& 'a'; (1 + 2)`. So it's back to the current version of the PR...

Julusian:
I have a bunch which are currently not expressions and like `$(x32:name_ch_05)\n$(x32:fader_ch_05)`, which isnt an expression, but also isnt starting with a prefix.

In fact with `$(custom:a) + '\n' + $(custom:b)`, just has to become empty for this to go wrong. as that will be executed as `($(custom:a) + '\n') + $(custom:b)`, so if a='' and b='word', you would end up with `0word`, with the newline becoming a 0.  
Variables becoming empty is I imagine a fairly common scenario that would trip users up.  


I suppose that the main that makes me unsure on this proposal is that users will still need to know the other ways to do this for when they are working with numbers that they want to concatenate, or strings which cast to numbers. There isnt a way to say 'perform this as a string operation' (something which could maybe be expressed as `String($(custom:a)) + $(custom:b)`, but then we need to preserve knowledge of that explicit to these operators)

> And conversely, of course, your case of a string variable changing to a number variable would be just as confusing today if you reversed it: a person writes $(custom:a) + $(custom:b) expecting "a" and "b" to be "numerical" variables but then at some point, one of those variables happens to be alphabetic, they'll get NaN 

I dont think this would be as confusing, as the expected outcome is different. If being fed two strings, you expect them to be concatenated, not merged into a single value. I think it becoming NaN is only weird if you expect a and b to be combined as strings instead of numbers.

> I do agree with you that had Companion started off with just the Javascript + behavior this exercise would have been unnecessary (though maybe other more common issues would have arisen, if numbers usually show up as strings...)

Back when this started, custom variables (and a lot of variables really) were mostly only ever strings, even if they contained a number. so this would have meant needing explicit type casts inside a majority of expressions.  
I think that some modules still do this so that they can give pre-formatted numbers at fixed precisions

arikorn:
Hmm... well then, the only "ironclad" options that currently comes mind are:

1. A new operator. My suggestion above was `..`,  (Or remove the bitwise operators and use `&`, or maybe make up our own like `+&`? or `$+` is mentioned in Wikipedia)
     Ex. ` 'a' .. (1 + 2) .. toFixed($(myvar))`, or `$(custom:a) .. '\n' .. $(custom:b)` for example
2. Resurrect the part I crossed out above (start expression with a concat operator) but document it as needing semicolons or commas as separators, 
    Ex1: empty-string is the operator: `''; 'a'; (1 + 2); toFixed($(myvar))` or
    Ex2:  any currently-not-unary operator (`&` here): `& 'a'; (1 + 2); toFixed($(myvar))`  `& $(custom:a); '\n'; $(custom:b)`

Either option should be very simple to code (I've already done option 2 ex1, for testing locally, for example.)

or..
3. Don't bother and we can either keep as "pending" or close the PR.

Julusian:
1) I am tempted by `..`. I dont think removing the bitwise operators is a good idea (breaking change, and it is useful at times)
2) I think that is going to be hard to document/explain and parse (within jsep currently), so I am not keen on that.
3) This is also a little tempting.

I think I want to think on this for a bit longer, and see if anyone else has any input.
I think my only objection to .. is that it is a little non standard (but is used in some languages) and so feels a little odd. not really a good enough reason to block this though

arikorn:
Quick replies, @Julusian:
1. Another option is `$+`. which might be more popular: it's used in mIRC. Or make up a new operator? `@` and `#` are available...
2. Actually these parse just fine in jsep right now: `& 'a'; (1 + 2); toFixed($(myvar))` returns a 'Compound' result and is fairly simple to concatenate (just map and join). It does require that the concatenation operator be at the top level (i.e. not inside a function call). You don't even need the semicolons in many cases. I agree though that it's a bit unusual.
3. Yeah. No problem waiting & ruminating...

dnmeid:
Here's my opinion. I think that '+' being able to concat strings is a long overdue. It is something that many users expect to work out of the box as it is the most intuitve version. Typically only after they realise that it is not working they use template strings or concat().

In the past we didn't care much for variable types, originally a Companion variable was meant to be always text and that is the reason why currently '1' + '2' = 3 and not '12'. I consider this to be a questionable decision in the past but nevertheless some old modules export all variables as strings, even if it is a numeric value and some users do still use this.  
So if we change the behaviour of + it will definitely break some configurations.

During some rewrites and advancements most JS types became usable in Companion variables and got used. Now with the readout of variable types the users are much better aware of the concept of types. I think changing the + to the most intuitive operation should be the way to go, even if there could be some trouble.

The most intuitive way doesn't necessarily be 100% JS compatible, I think especially + has some very unintuitive behaviour in JS.
To start how about that:
- all operands are numbers (and not NaN): arithmetic addition, existing functionality
-  all operands are either number or string, that can be parsed as number (not ''): arithmetic addition, existing functionality
- at least one operand is a string that can not be parsed a number: concatenation of the parsed strings ( 1+2+'dB' would be '12bB' and not '3dB', (1+2)+'dB' would be '3dB', ''+ 1+2 would be '12'), new/changed functionality
Tbd:
what about things like `true + false`, `true + 1`, `[1,2,3] + 42`

Now the only awkward behaviour would be the existing functionality of number parsing. From a programmers standpoint this may look wrong (and it looks wrong to me) but it is the same behaviour as e.g. Excel has and so I think we can get along with it.

So the main change would be handling of empty strings or spaces around numbers. We would have to announce this quite good and maybe also notice the module developers once more to maybe review the types of the variables they expose.

Julusian:
> In the past we didn't care much for variable types, originally a Companion variable was meant to be always text and that is the reason why currently '1' + '2' = 3 and not '12'.

I would argue that we still largely dont care much about variable types. When defining a custom variable you can opt into it not being a string, but that require the user to do that. And the action to set a custom variable doesnt support that (it does if setting from an epression).
So I would guess that today a majority of custom variables containing a number are doing so as a string.  
If I check my admittedly small list of custom variables, the only ones which are numeric are the ones set with expressions. and their startup value is a string 0.

>  consider this to be a questionable decision in the past but nevertheless some old modules export all variables as strings, even if it is a numeric value and some users do still use this.

Checking over some popular modules:
* bmd-atem - most numbers are numbers, some are forced to string for some reason
* studiocoast-vmix - many numbers are forced to string to give a predictable format

>  So if we change the behaviour of + it will definitely break some configurations.

This is what worries me.  
If we do this, I wonder if this justifies being a 5.0.

I dont disagree that using '+' is intuitive, but I worry about the cost and frustration of supporting that.

> all operands are either number or string, that can be parsed as number (not ''): arithmetic addition, existing functionality

How do we define parsable as a number?  
`1 ` and `1.` both return `1` when fed through `Number(v)` or `parseInt(v)`. I dont think they should be determined to be numbers, so we would need to define some rules on what is and isnt parsable as a number.  
This worries me, as we will miss things, and have to refine it with patch releases


> at least one operand is a string that can not be parsed a number: concatenation of the parsed strings ( 1+2+'dB' would be '12bB' and not '3dB', (1+2)+'dB' would be '3dB', ''+ 1+2 would be '12'), new/changed functionality
Tbd:

This one will be a hard one to achieve. The parser converts `$(internal:a) + $(internal:b) + $(internal:c)` into: 
```
{
	type: 'BinaryExpression',
	operator: '+',
	left: {
		type: 'BinaryExpression',
		operator: '+',
		left: {
			name: 'internal:a',
			type: 'CompanionVariable',
		},
		right: {
			name: 'internal:b',
			type: 'CompanionVariable',
		},
	},
	right: {
		name: 'internal:c',
		type: 'CompanionVariable',
	},
},
```

Then at the point of execution, we are iterating through the tree a node at a time. So when executing the + we only have the left and right values. The execution is custom logic, so we probably could pass around some extra state/tacking, and maybe it wouldnt be all that bad as its only the one operator and if we ignore performance we could try number mode and if that fails try again in string mode..

> So the main change would be handling of empty strings or spaces around numbers.

We probably also need to think some more on how to let the users force something to be in string mode. ie you want to `'1'+'2'` to be `12`.
Perhaps that could be 'solved' by a `string('1')` function to force something into a string. combined with some more reworking of the execution to track whether the type of a return value was explicit or implicit, which would influence how it is used. eg `'1'` is explictly declared as a string and should not be coerced as a number, but a `$(custom:a)` containing a string is not explicit so can be tried as a number

----

Personally, the breaking nature of changing '+' and the high level of risk of fiddling with the expression parsing/execution that deeply is worrying to me (risk of both breaking things and annoying users).  
And at the end of it I worry will be a syntax that while is more intuitive, is more frustrating because of the auto number parsing.

If we didn't have the auto number conversion then I would be more open to this, as js proves it is clearly viable. But as I believe that a high proportion is relying on our current behaviour, changing it now feels like making an even bigger mess.

Of course, that is the other approach to take; simply dropping the auto converting strings to numbers and forcing users to do that themselves.  
If we are going to break things, we may as well solve the root problem at the same time

arikorn:
I totally agree, Dorian, that users who are familiar with languages that let the `+` operator do double-duty *do* find the current behavior confusing, in fact the topic came up again [in FB just this weekend](https://www.facebook.com/groups/companion/permalink/4116795851872067/). I also think it's great that you're recommending solutions that go against both convention and Companion practices! With that in mind, maybe it's time to reconsider your objections to Stream Deck+ swipe #3721...

Just to compare these two PRs: here you advocate introducing a breaking change that will confuse many users (by your own admission), break existing modules, and that breaks conventional operator precedence rules (operators of same precedence are evaluated left to right). By contrast the swipe PR (a) introduces no new functionality (button-independent operations to change a page already exist both in Triggers and the TCP API; surface actions not tied to a specific button exist in Loupedeck Live code, maybe xKeys and others), (b) introduces no breaking changes, and (c) actually implements a _conventional_ behavior (as Julian has pointed out). I have even addressed all of your stated concerns in that PR. More importantly users have been requesting SD+ swipe support for years, it has been an unmet milestone since Companion v3.5, and over 50 FB users have liked/loved/commented on my demo of the swipe feature.

Back to this PR: Julian already said most of what I was thinking, so to summarize (a) while your proposed rules makes sense to you, I suspect they will be mystifying to people with different backgrounds, so that (b) Really the only way to avoid confusion is with a separate operator, which is why I ultimately agreed with Julian's arguments.

In terms of priority, TBH, I consider this PR to be more-or-less a "throwaway" exercise, and have no objection to closing it if we can't find consensus, even though I've ended up putting in even more time on this PR than on the SD+ swipe PR.

OTOH, the SD+ PR #3721 should really be merged: I gain nor lose nothing either way, but SD+ swipe functionality is something your users have been requesting for three years and something that you yourselves have prioritized as a milestone for over a year.

arikorn:
reopening at Julian's request and for further or final review. The current version of this PR uses `..` as the string-concatenation operator and will convert numerical operands to strings. It has lower precedence than mathematical operators, so for example
   'b' .. 1 + 3  = 'b4' rather than producing an error.

thedist:
Why is the plan to use `..`? In some languages that's a range operator (eg, array[1..5] for an array of ints from 1 to 5, or "A" .. "Z" for an array of A to Z characters). Even withing JS there are uses for `..`, such as `123.toString()` throws a syntax error, where as `123..toString()` returns the string `"123"`. This is because within JS a `.` character following a number is taken as a decimal point rather than a period. Similarly, `.123` is interpreted as `0.123`, even within Companion Expressions where `.1 + 3` would be `3.1`, so when doing stuff like `'b' .. 1 + 3 = 'b4'` it can be problematic as you're now also requiring whitespace to be exact.

Syntax aside, even `'b' .. 1 + 3 = 'b4'` is somewhat debatable as in JS that would be `"b13"`, not `"b4"` so there would be some bad UX with users expecting something different to what you're returning.

I'd rather not introduce yet more non-standard syntax that's something unique to Companion and differs from industry standards. Why not consider something such as a settings toggle that by default leaves everything as it is so there's no breaking changes, and toggled on the `+` would act more like it does in JS? It'd be non-breaking, it'd give users the choice how they wish Expressions to act, it wouldn't add random syntax to how we handle expressions,

Julusian:
> I'd rather not introduce yet more non-standard syntax that's something unique to Companion and differs from industry standards. Why not consider something such as a settings toggle that by default leaves everything as it is so there's no breaking changes, and toggled on the + would act more like it does in JS? It'd be non-breaking, it'd give users the choice how they wish Expressions to act, it wouldn't add random syntax to how we handle expressions,

If we have that option, it would need to be as part of deprecating the old behaviour and a push to move everyone to the new one.   
But then I worry that it will leave users in confusing states where some presets do and some dont as they expect differing modes. And which ones work will depend on what expression setting they have.
We don't have anywhere that expressions can come from presets currently, but both the graphics overhaul and #2345 will bring many.

Alternatively, we just change it with no option and try to communicate that well and hope the impact is minimal. But would that have to wait for 5.0, or are we ok with that being in a 4.x?

arikorn:
Jeff, (@thedist), I don't entirely disagree with you but want to address some of your claims:

>Even withing JS there are uses for .., such as 123.toString() throws a syntax error, where as 123..toString() returns the string "123". 

Wow! That's pretty amazing and arcane, thanks for sharing that!  Of course in Companion `123..toString()` and even `'a'.toString()` is an error, which does highlight the point that Companion expressions are not, in fact, Javascript expressions.  (FWIW with this PR `123..toString() = '123{object Undefined}'`)

>  when doing stuff like 'b' .. 1 + 3 = 'b4' it can be problematic as you're now also requiring whitespace to be exact

No, this does not depend on whitespace. In this PR,  `'b'..1+3 = 'b4'` (and `'b'...1 + 3 = 'b3.1'` analogous to your Javascript example above). Of course `'b'. .1 + 3` -- space between the dots -- is an error both in current behavior and this PR, but so is `true & & false`

> Syntax aside, even 'b' .. 1 + 3 = 'b4' is somewhat debatable as in JS that would be "b13", not "b4" 

This is exactly why this PR uses a different operator:  ` 'b' .. 1 + 3` is _not_  ` 'b' + 1 + 3`.  In Javascript ` 'b' .. 1 + 3` is an error whereas in this PR `1 .. 3` is '13', not 4. There's no ambiguity about what `+` does and what `..`  does. (And regarding operator precedence, that's a setting in jsep, we can make it anything we want.)

OTOH by using Javascript rules, as you suggest at the end, the behavior of `$(myvar) + 1` can change if the contents of $(myvar) changes from characters to digits, for example, even if it's always a string, which is different from Javascript behavior in which `'1' + 1 = '11'`

> differs from industry standards

I don't think we can claim any industry standards on [string concatenation operators](https://en.wikipedia.org/wiki/Concatenation), it's really all over the map, and there is even [precedent for `..`](https://en.wikipedia.org/wiki/Comparison_of_programming_languages_(strings)), though your point about range operators in other languages is a fair one.

---
So in the end, I think this issue is fraught no matter what we do because of Companion's history of treating numerical strings as numbers. Having an independent operator still seems to me to be the least confusing, but I think doing nothing would be my second choice, if the alternative is incorrectly mimicking Javascript.

dnmeid:
> With that in mind, maybe it's time to reconsider your objections to Stream Deck+ swipe...

@arikorn I don't think that I like the way how you are arguing. You are often using manipulative patterns and I would be happy if you could stop that.

Going back to the concatenation operator, my opinion is still the same. I think + is the best choice and the breaks we would introduce feel minimal to me. Given that the often used '42' + '123' would still work, breaks would only occur at cases that are quite exotic. Of course it depends how strict we are at parsing a string to a number. That would need to be a Companion usage decision like e.g. our extended range of line breaking characters. 

I'm fine with going that route without a configuration option in major version 5. 
I'd be more concerned with a + behavior setting that this would lead to more frustration or confusion in the long run. Many users are searching for help with expressions or asking chatbots. I'd prefer to have only one meaning of + in expressions.

On the other hand I think any other operator would be very unintuitive and so the only advantage would be to save you from typing some more chars because we have the concat() function. Wise programmers know that you write code only once, but read it often. I think readability is more important than writability and `'a' + 'z'` or `concat('a', 'z')` are more readable than `'a'..'z'`.

arikorn:
@dnmeid wrote:
> You are often using manipulative patterns 

My sincerest apologies, Dorian!  I have only the highest regard and respect for you and your many and long-term contributions to Companion. I do not know what you mean by "manipulative patterns" especially when you say that I "often" use them, -- it is true that I let some exasperation show in this one reply, but I believe I have been very respectful to you in all our other interactions. If you feel otherwise, I'd be happy to discuss it with you -- there is always risk of misunderstanding  with written communications -- please reach out to me here or on slack. Again please accept my apologies.

As for this PR, we appear to be unanimous in believing that it is fraught, at best, so I'm closing it. (Sorry Julian!)

This is, of course open-source code so if someone wants to reuse any parts of it they are more than welcome to it!

PR #3708, opened 2025-10-15
Fix: partial imports resetting unselected items.

### LINKED ISSUE #3639: [BUG] 4.1 Import of just Custom Expressions reset everything

### Describe the bug

When I import just custom expression variables, it is resetting my whole configurations.  All of the pages, connections, etc are being removed.

### Steps To Reproduce

1. Create a configuration that has multiple pages, custom variables, triggers, connections, custom expressions
2. Export the config with just expression variables
3. Import the just exported configuration and it should only allow expression variables to be imported.
4. Click the Import Selected Components.
5. You will now see that everything except for the connections that the expression uses are left.  The pages, connections, triggers, variables are all reset.

### Expected Behavior

I expected it to just import the custom expression variables and remove any that already existed.

```markdown
- OS: MAC 15
- Browser: Chrome
- Companion Version: 4.1.0 stable
```

### Additional context

[My Expression Export](https://github.com/user-attachments/files/22371595/Justins-Air-4.ts.net.lan_2025-09-16-13.json)

arikorn:
Fixes #3639

Notes:
1. *Full.tsx* was not setting `config` to reflect the lack of available data fields.

2. *Controller.ts* `importFull` did not check that `config` had "valid" values, i.e. false if data is missing, and was therefore allowing these categories to be replaced with [] or {} (not in `#reset()`, but in `importFull` itself).

Note: either 1 or 2, above, are enough to fix the issue, but I fixed both as a failsafe.

In addition, some code tidy-up:

3. In `importFull` don't allow `conifg` to be null. It never is, and allowing it to be null just makes the code harder to read. (We've discussed this before, Julian, but this time it really was confusing things near the top, so I hope you're OK with keeping it cleaner. It can always be reinstated at a later time if needed.)

4. In *ImportExport.ts* rearrange fields (properties) so they are in the same order in the different interface definitions, for ease of comparison.

### DIFF
--- a/companion/lib/ImportExport/Controller.ts
+++ b/companion/lib/ImportExport/Controller.ts
@@ -511,7 +511,7 @@ export class ImportExportController {
 						if (data.connectionCollections && data.connectionCollections.length > 0) {
 							this.#instancesController.connectionCollections.replaceCollections(
 								data.connectionCollections || [],
-								fullReset
+								!fullReset
 							)
 						}
 
--- a/companion/lib/Resources/CollectionsBase.ts
+++ b/companion/lib/Resources/CollectionsBase.ts
@@ -49,20 +49,20 @@ export abstract class CollectionsBaseController<TCollectionMetadata> {
 	}
 
 	/**
-	 * ~Replace~ Update all collections with imported collections
-	 * if `reset` is false, preserve existing entries with the same ID. (This could lead to sortOrder clashes, but it shouldn't matter)
+	 * Replace Update all collections with imported collections
+	 * if `merge` is true, preserve existing entries with the same ID. (This could lead to sortOrder clashes, but it shouldn't matter)
 	 * TODO: add merge options? deal with sortOrder clashes?
 	 */
-	replaceCollections(collections: CollectionBase<TCollectionMetadata>[], reset = true): void {
+	replaceCollections(collections: CollectionBase<TCollectionMetadata>[], merge = false): void {
 		// Clear existing collections
-		if (reset) {
+		if (!merge) {
 			this.#dbTable.clear()
 			this.data = []
 		}
 
 		// Import new collections
 		for (const collection of collections) {
-			if (reset || !this.#dbTable.get(collection.id)) {
+			if (!merge || !this.#dbTable.get(collection.id)) {
 				this.#sortCollectionRecursively(collection)
 				this.data.push(collection)
 				this.#dbTable.set(collection.id, collection)


companion/lib/ImportExport/Controller.ts
```diff
@@ -488,22 +488,23 @@ export class ImportExportController {
...
-							if (!data.custom_variables) resetArg.customVariables = false
-							if (!data.expressionVariables) resetArg.expressionVariables = false
-							if (!data.surfaces) resetArg.surfaces = false
-							if (!data.triggers) resetArg.triggers = false
+						// `config` tells what to load. Ensure that config is false for missing sections:
+						// note: this is failsafe is not strictly necessary, since Full.tsx does it right, now.
+						// note that config doesn't have a entries for 'connections' or 'userconfig'...
+						for (const key in config) {
+							let dataKey = key.replace(/^customVariables$/, 'custom_variables')
+							dataKey = dataKey.replace(/^buttons$/, 'pages')
```
Julusian, line 496: I'm not the biggest fan of this, because I know I will forget to update it when renaming things inside the exports in the future.
Equally, that was a problem with what I had before and being easy to forget to add new things, sooo 🤷

companion/lib/ImportExport/Controller.ts
```diff
@@ -488,22 +488,23 @@ export class ImportExportController {
...
+							config[key as keyof typeof config] &&= dataKey in data
 						}
 
+						// Add fields missing from config
+						// If partial import, dont ever reset connections (or userconfig until added to UI)
+						const resetArg: ClientResetSelection | null = fullReset
```
Julusian, line 502: I'm wondering if what we need here is a function which takes in `config` and `data`, and constructs a new `config` and `resetArgs` from scratch, so help with the sanitising.

Or perhaps even instead of returning a `config`, it could return a mangled `data`, containing just the values which should be imported. optionally with a different structure, to pair things like the collections and items together

companion/lib/ImportExport/Controller.ts
```diff
@@ -488,22 +488,23 @@ export class ImportExportController {
...
+							? null
+							: { ...config, connections: false, userconfig: false }
+
 						// Destroy old stuff
-						await this.#reset(resetArg, !config || config.buttons)
+						await this.#reset(resetArg, config.buttons)
 
 						// Import connection collections if provided
 						this.#instancesController.connectionCollections.replaceCollections(data.connectionCollections || [])
```
Julusian: This line is one that still concerns me as a lurking bug

companion/lib/ImportExport/Controller.ts
```diff
@@ -531,7 +532,8 @@ export class ImportExportController {
 							}
 						}
 
-						if (data.pages && (!config || config.buttons)) {
+						// note data.pages is needed only to satisfy TypeScript, since config.buttons is false if pages is missing.
+						if (data.pages && config.buttons) {
 							// Import pages
 							for (const [pageNumber0, pageInfo] of Object.entries(data.pages)) {
```
Julusian: alternatively, do like we have elsewhere?

```suggestion
						if (config.buttons) {
// Import pages
for (const [pageNumber0, pageInfo] of Object.entries(data.pages || {})) {
```

webui/src/ImportExport/Import/Full.tsx
```diff
@@ -159,6 +159,17 @@ function FullImportTab({ snapshot }: FullImportTabProps) {
...
 	}))
 
+	// make sure config matches the available data. This is resilient, but it might be
+	// cleaner/simpler to just manually add snapshotKeys.includes('keyname') in useState?
+	const availableConfig = { ...config }
+	for (const k in availableConfig) {
+		const key = k as keyof typeof config // get around TypeScript errors
+		availableConfig[key] &&= snapshotKeys.includes(k) // check-types doesn't like `key` as arg to `includes())`
+		if (availableConfig[key] !== config[key]) {
+			setConfig(availableConfig) // note: this is effectively a noop after the first change is detected.
```
Julusian: This is going to make react very grumpy, it doesnt like state being updated during the render.

You might be able to wrap this whole block in a 
```
useEffect(() => {
...
}, [config, snapshotKeys])
```

Semi-related: I should probably have a look at some of this and see if it can be converted to mobx things, to simplify some reactivity.. Not sure how worth it that is though

companion/lib/ImportExport/Controller.ts
arikorn: I completely agree with your sentiment, but don't see any "correct" way to fix it without breaking either the export file or the internal database. The current code does make it more compact, and a bit easier to see the "problem" (i.e. a little harder to forget...).

That said, the problem is not related to this PR, and has existed since at least the mid v3's, if memory serves, so perhaps you could post an Issue, and either suggest or we can discuss a solution outside this PR?  (I'm happy for you to assign it to me!)

companion/lib/ImportExport/Controller.ts
arikorn: I don't see how it would change anything: the current code essentially "reconstructs" `config` from `data` and then uses it to create `resetArgs` , which seems the correct sequence to me. 

With the exception of your next comment, collections seem to be handled safely for now, i.e. are paired in the code with their categories...  (And why are collections not named "groups" as in surfaces?)

If your concern is the mismatch of properties as in your previous comment then, perhaps we could make this part of that separate issue?

companion/lib/ImportExport/Controller.ts
arikorn: Good call! I've "protected" it for now.

Do connection collections need to be defined before instances are imported (i.e. not the other way around)? I assume the answer is yes, since the code works, but just asking...

companion/lib/ImportExport/Controller.ts
arikorn: Done.  (Though I use `??`)

webui/src/ImportExport/Import/Full.tsx
arikorn: I don't think that `useEffect` is appropriate here, since it will cause the same number of rerenders (one, actually, either way, i.e. once it changes config the first time it doesn't change it again unless `snapshotKeys` changes, which I suspect it never does). `useEffect` just postpones execution until after the render, but the outcome is otherwise identical in this case.

After going around in more circles in my head I realized that this snippet doesn't need to call setConfig at all since it doesn't affect the UI (the only properties to be changed are disabled ones), and even if it did, we're modifying the value  before it's passed to the subcomponents, so it's perfectly valid React (sort-of...read on). The need for the snippet is to ensure that the correct values are returned to the server, and it works because we're not changing the value in React's "mind", i.e. we're preserving object identity.  I suppose if you wanted to be real picky, I could put it in `doImport` before `importFullMutation`.  Or in the `snapshotKeys` `useMemo`... and swap the order with config. Hmm... as much as I hate to say it, it's probably most correct to put the code into `doImport`, on the very improbable (impossible?) chance that `snapshot` changes midstream to *add* new data fields...  Alright, I'll let you weigh in on this one first...

(mobx, is like Immer? OK. I looked it up. I was thinking the same thoughts as you...i.e., not sure how worth it that is.)

arikorn:
Another thing I hate to say is that without checking for broken "links" between categories (like loading pages w/o variables and missing a variable), partial imports will always have the potential to produce unwanted effects. Since my intent with the original PR was mostly to skip resetting user-config, if you're still not comfortable with all this, perhaps rather then enabling "partial imports" I could just add a checkbox "Don't reset Settings" Or if we want to keep partial imports we could add a "buyer-beware" warning...?

Just to be clear: (a) I believe the code fixes here should be merged regardless, as they do address real issues and (b) even without partial imports a person can cause similar problems by importing partial exports.

Anyway, since @digitaldrummerj appears represent someone who actually wants to import selectively, perhaps Justin you could chime in on your use-case and thoughts on this question?

companion/lib/ImportExport/Controller.ts
Julusian: > And why are collections not named "groups" as in surfaces?

Initially they were called them groups, then that changed because of behavioural differences. Collections are intended for organisational purposes, but surface groups link the surfaces to be like one surface so are a functional purpose.  
So I wanted to differentiate them in the terminology used.

> I don't see how it would change anything: the current code essentially "reconstructs" config from data and then uses it to create resetArgs , which seems the correct sequence to me.
> If your concern is the mismatch of properties as in your previous comment then, perhaps we could make this part of that separate issue?

Yeah so this does kinda tie to the previous comment. While that is what it is doing, by doing it more explicitly it would be clearer, and while a bit wordier it could also be done in a typesafe way, by more easily explicitly building new objects from scratch.  
The intention being to make it a more explicit multi-step transform so that the final step of performing the import isnt trying to match stuff up across 2/3 objects, but can just look at the one.

But yeah probably not required to do as part of this, it is preexisting mess

webui/src/ImportExport/Import/Full.tsx
Julusian: well that mutation during the render could also make react grumpy, because its a change to an object react is storing that it is unaware of.. tbh Ive never looked into if that is allowed or not

oh, actually they dont object to conditionally calling setState during render, so maybe what was there was fine https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-render

but actually doing this inside doImport probably makes the most sense. That avoids any extra renders, and the extra dependency in that callback wont be any worse than it already is with its dependency on `config`.
yeah the snapshot changing should be impossible

Julusian:
> Another thing I hate to say is that without checking for broken "links" between categories (like loading pages w/o variables and missing a variable), partial imports will always have the potential to produce unwanted effects. 

While that would be annoying, that is their fault for not importing everything that they need.  
We don't offer all the import functionality they might need to handle that though, which is where my proposed overhaul will step up to allow for this more complex stuff.

Showing that an import will leave broken references may not be too bad to do actually, the challenge becomes how to show that in the ui, and how to help them fix it. Again, that overhaul would give some more scope/space to have an interactive process to be warned about these things and give some ability to resolve them as part of the import.
(I might be dumping too much hope and wishful features into that overhaul though. It is still just that sketch in my mind)

webui/src/ImportExport/Import/Full.tsx
arikorn: Done.

arikorn:
OK, I pushed all the changes I believe we've agreed are needed in this PR (let me know if I missed something), and now raise a related issue, which may or may not be worth doing here:

I'm struggling a bit with understanding the `connections` behavior...currently the user is forced to import the connections but if doing a partial import we never reset them but instead merge the two, this is a bit non-intuitive: the standard behavior here is that a feature is either reset and reloaded, or not reset and not reloaded. Why has the checkbox for connections is removed from the UI? and why this "intermediate" behavior?  Since the checkbox is removed regardless of  "partial reset" (what I incorrectly call "partial import") or "full reset & import", maybe adding the choice back to the UI is not part of this PR, but OTOH, the inconsistency occurs only in a "partial" import (i.e. connections are imported but not reset).

A somewhat trivial example of inconsistency:
1. make a connection collection containing two items. 
2. Export the full config
3. Delete the collection (incidentally, this doesn't delete its members, which seems wrong to me)
4. Delete one of the "liberated" module instances.
5. Do a "partial-reset" import of the file, leaving all boxes checked (which doesn't matter, since connections isn't a choice)
6. Now the collection has been restored but it contains only one item (the one you deleted), the other item remains unaffiliated.
7. Now do a "reset & import" 
8. The collection is properly restored with both items.

Since this is due to the "merge", the related question is: why you're doing a merge only for connections?

 If we reinstate the connections checkbox, then worst-case if user deselects connections couldn't we use the same feature that "import page" uses to map connections?

Julusian:
> Why has the checkbox for connections is removed from the UI? and why this "intermediate" behavior?

Quite simply because resetting the connections will remove any actions/feedbacks that reference those connections. So if you were to do a reset which reset and imported connections and triggers, all your buttons would now be broken (but might look ok, as only the actions/feedbacks have been stripped out).

> A somewhat trivial example of inconsistency:

Yeah, I dont disagree with this. Im pretty sure I wrote somewhere that handling a merge of these things is complex and needs some thought on how to do in a good way.

> If we reinstate the connections checkbox, then worst-case if user deselects connections couldn't we use the same feature that "import page" uses to map connections?

Maybe, I need to think on if that would work fine or cause alignment/id issues.  
That sounds like one of the stages I proposed in that overhaul. The problem with the current system is that it wasnt designed to do this kind of granularity

arikorn:
> Resetting the connections will remove any actions/feedbacks that reference those connections. So if you were to do a reset which reset and imported connections and triggers, all your buttons would now be broken (but might look ok, as only the actions/feedbacks have been stripped out).

So it would only be a problem if someone includes connections but not pages and triggers? (So for example, the code could act more cautiously depending on whether this condition is true...)

Since it may be a bigger issue, perhaps we just merge this PR and open an issue with all our remaining comments here?
i.e. my concern probably doesn't result in more than minor annoyances...

Julusian:
> So it would only be a problem if someone includes connections but not pages and triggers? (So for example, the code could act more cautiously depending on whether this condition is true...)

yeah I guess it could. It would need some thought on how to make it clear to the user what will happen too.  
considering that until #3557 it always did a full reset before the import, so it made sense to always import the connections. (ok, not 100% of the time, but almost all the time)

companion/lib/ImportExport/Controller.ts
```diff
@@ -488,39 +488,43 @@ export class ImportExportController {
...
+							: { ...config, connections: false, userconfig: false }
+
 						// Destroy old stuff
-						await this.#reset(resetArg, !config || config.buttons)
+						await this.#reset(resetArg, config.buttons)
 
+						// Always Import instances
 						// Import connection collections if provided
-						this.#instancesController.connectionCollections.replaceCollections(data.connectionCollections || [])
+						if (data.connectionCollections && data.connectionCollections.length > 0) {
```
Julusian, line 511: I think this needs to be tied to whether a fullReset is being done.

As is, importing just the custom-variables from an full-export will replace all the connectionCollections which could result in deleting almost all of the collections

arikorn: Yes, under fairly specific conditions though, i.e. the file has a different set of collections than are currently stored.

However, there are not conditional statements that will change this behavior: connections (including conn. coll.) are always imported -- there isn't even a field for connections/instances in the `config` object.  Then it follows that:

- If `fullReset`: the previous line wiped out collections; this line will add from the file, if any
- If _not_ `fullReset`: previous line preserved collections (reset connection is _always_ false), but this line will *always* replace collections from the file if they are present (import connection is _always_ true).

A slightly different way of viewing it:
- file has _no_ collections: `fullReset` removes existing collections in the reset line; partial reset leaves collections alone.
- file _has_ collections: existing collections are replaced regardless of fullReset or not (because `config.instances`  is implicitly true)

The best quick-fix I can think of is to make `replaceCollections` behave more like `#importInstances`, i.e. don't replace existing objects. Since you seem to be preserving id's across imports, it shouldn't be to bad. Have done. will push shortly.

arikorn:
@Julusian, I think the various components for re-enabling "connections" as an option can be found in other parts of the code, but I still would like to encourage merging this one first as a bug-fix to v4.1.1, and then we can open another PR (I can consolidate our conversation here into an Issue if you'd like) for the remaining items we've discussed, which seem more like features to me.

companion/lib/ImportExport/Controller.ts
Julusian: tbh I was hoping for `if (fullReset && data.connectionCollections && data.connectionCollections.length > 0) {`, but what you have done is better and more like what it would eventually want to  become

PR #3709, opened 2025-10-15
Add missing `solid-js` package to webui

arikorn:
vite crashed (`yarn dev` in webui or `yarn dev:webui`) on startup, complaining that it couldn't find solid-js, so I added it to devdependencies in webui (`yarn add --dev solid-js`)

(It doesn't seem necessary to build the webui.)

### DIFF
--- a/webui/package.json
+++ b/webui/package.json
@@ -89,6 +89,7 @@
     "serve": "vite preview"
   },
   "devDependencies": {
-    "jsdom": "^27.0.0"
+    "jsdom": "^27.0.0",
+    "solid-js": "^1.9.9"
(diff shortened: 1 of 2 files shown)


Julusian:
I noticed this too on a branch the other day after a random change. 

It looks like it was https://github.com/TanStack/router/issues/5420, and so it should be resolved by https://github.com/bitfocus/companion/commit/4fe8ea63e5dfc7a4f4fc9835821bbd2ed27bf993

arikorn:
Wow, that was very close timing!
OK. All good now. Thanks.

PR #3736, opened 2025-10-29
feat: Add options to to restrict page access

arikorn:
- Options to restrict pages and which ones to allow are added to `SurfaceGroupConfig`
- Modify *EditPanel.tsx* to add the new options
- Modify `SurfaceGroup.setCurrentPage` to handle page restrictions
- Import/Export new fields (pages are converted similarly to `startup_page_id` and `last_page_id`
- (Modify `InternalPageIdDropdown` to allow multiple values)
- Tighten up page-restriction enforcement in Groups.ts and ensure as best as possible that page restrictions are enforced (a) at startup and (b) even if enabled after visiting now-restricted pages.
- Update documentation ("Getting Started Guide")

Note: this PR was split off from PR #3721 since it provides independent functionality


companion/lib/Surface/Group.ts
Julusian, line 1: I'm not sure I agree on some of the behaviour here.

I was imagining this restriction logic to only affect the +1/-1 behaviour, so that surfaces could use the next/previous page buttons and get a smaller/customised loop.
Because I see value in being able to jump to other pages from a button, even when wanting the default cycling to be limited to a few other pages.

But I can see an argument for this way too...

arikorn: I'm happy to go either way but the current behavior just seems more consistent. Otherwise, what do you do when the person is out of the "allowed zone"? Does "next page" then go to the next page or jump to an arbitrary page in the allowed zone? Likewise, on startup, if the designer left _startup page_ to something outside the allowed zone, should it start there or, as is the current behavior, go to a page in the allowed zone? In all cases, it seems to me to be most consistent to have "strict" enforcement.  FWIW, the way I've currently implemented it, a designer could put a "jump to page x" action in a page and have it "active" only if permissions allow. My design decision was along the lines of a belief that it's bad design to intentionally get into any of these edge cases, so the current implementation protects the designer from some accidental oversights. (And the name of the feature _is_ "Restrict Page Access" 😀 .)

Anyway, just say the word, if you want this changed.

companion/lib/Surface/Group.ts
```diff
@@ -248,39 +250,85 @@ export class SurfaceGroup {
 		this.#pageHistory.index = 0
 	}
 
+	/**
+	 * Get first page ID, taking into account `groupConfig.allowed_page_ids`
+	 */
+	#getFirstPageId(): string {
+		const validPages = this.groupConfig.restrict_pages ? (this.groupConfig.allowed_page_ids ?? []) : []
+		return validPages.length > 0 ? validPages[0] : this.#pageStore.getFirstPageId()
```
Julusian: Perhaps this should be checking that the contents of `validPages` are valid according to `this.#pageStore.isPageIdValid(pageID)`?

companion/lib/Surface/Group.ts
```diff
@@ -248,39 +250,85 @@ export class SurfaceGroup {
...
+	}
+
+	/**
+	 * Check if a page ID is valid, taking into account `groupConfig.allowed_page_ids`
+	 */
+	#isPageIdValid(pageID: string): boolean {
```
Julusian, line 277: I personally think the order of the if-elseif-else is a bit hard to follow/sanity check from a skim read.

I think I would have gone with:
```
// Fallback when no restrictions
if (allowPages.length === 0) {
  return this.#pageStore.isPageIdValid(pageID)
}

// Make sure page exists
if (!this.#pageStore.isPageIdValid(pageID)) {
  return false
}

// Check is valid within selection
...
```

In other words, having a few small guards at the top, and perhaps a nice and important thing for minimising risk of bugs, it starts with a fallback for the old/default behaviour, making it easy to skim and see that this does nothing when the setting is disabled.  

Not essential to change, just my thoughts on code style

companion/lib/Surface/Group.ts
```diff
@@ -248,39 +250,85 @@ export class SurfaceGroup {
...
+		const allowedPages = this.groupConfig.restrict_pages ? (this.groupConfig.allowed_page_ids ?? []) : []
+		if (!this.#pageStore.isPageIdValid(pageID)) {
+			return false
+		} else if (allowedPages.length > 0) {
+			const isAllowed = allowedPages.includes(pageID)
+			if (!isAllowed && this.#pageHistory) {
+				const pageHistory = this.#pageHistory
```
Julusian: having said that, it might just be this history cleanup which makes it harder to skim as it makes this case look very complicated.

But I'm not sure that this should be here, it seems like it would be enough to do it when trying to pop from the history?

docs/3_config/surfaces/pagepermissions.md
```diff
@@ -0,0 +1,11 @@
+## Page Permissions: Restricting access to pages by device or group
+
+Starting with v4.2, individual surfaces or surface-groups provide the option to restrict which pages a device -- or devices in a group -- may access. For example:
+
+---
+
+## ![Restrict Pages Setting](images/restrict_pages.png?raw=true 'Discovery')
```
Julusian:
```suggestion
## ![Restrict Pages Setting](images/restrict_pages.png?raw=true 'Restrict Pages Setting')
```

docs/3_config/surfaces/groups.md
```diff
@@ -1,10 +1,32 @@
...
+
+A second use case for groups, is that you can use a group as a stable id for a surface.  
 This helps setups where you need to be able to swap out a Stream Deck quickly, and also need actions to reference specific Stream Decks.  
 You can create your groups and program the system ahead of time, then later add or remove surfaces to the groups without needing to update any buttons.
+
+## Use Case 3: Establishing "Permission Groups" with restricted access
+
+Starting with v4.2, groups -- as well as ungrouped surfaces -- now provide the option to restrict which pages a device in that group may access. Using this in a surface-group configuration can be very useful if:
+
+1. Each surface is controlled independently (for example, each surface is controlled by its own computer or Raspberry Pi)
```
Julusian: I dont understand what this means, or perhaps what the relevance is to groups

docs/3_config/surfaces/groups.md
```diff
@@ -1,10 +1,32 @@
...
+Elsewhere in Companion, the actions that can change the page of a surface will no longer see the individual surfaces, and can instead pick the group.
 
-Another use case for this, is that you can use a group as a stable id for a surface.  
+## Use Case 2: Addressing a surface without being tied to its serial number
+
+A second use case for groups, is that you can use a group as a stable id for a surface.  
 This helps setups where you need to be able to swap out a Stream Deck quickly, and also need actions to reference specific Stream Decks.  
 You can create your groups and program the system ahead of time, then later add or remove surfaces to the groups without needing to update any buttons.
+
+## Use Case 3: Establishing "Permission Groups" with restricted access
```
Julusian: I'm still not keen on this being a use case of groups.

While what you have written is true, it is only a valid in combination with either 1 or 2.  
So at the very least it would need a clear warning of 'all surfaces added to this permission group will be linked to show the same page as each other, changing the page on one will change all members of the group'. 
And the permissions also work on the surface individually, so I think that promoting doing permissions via groups is going to be more confusing.

When I see 'permission groups', I expect that to mean that I could assign the same restrictions to multiple surfaces, but they will otherwise be their operated as any other surface (ie, each member of the group can be showing a different page)

I suspect the linking property of 'permission groups' will make them annoying to use, as then when you realise that you need a second page because you ran out of space on the first, now you have to redo it without the permission group.

----

I think a large part of my problem with this is that the groups are designed solely to link multiple surfaces to follow each other on the same page. And while it could be useful to share these permissions properties, I expect that someone will find that page behaviour annoying and it will push us to start morphing these groups to not link the pages and just link other behaviours (partly because that is what I read this documentation to be saying is possible) 

And that is not a path I want to go down; doing so will break core assumptions about how the groups work

companion/lib/Surface/Group.ts
Julusian: tbh I dont have a strong preference, so lets go with what you've done

if/when someone asks maybe that checkbox could become a dropdown to choose between 'disabled', 'strict' and 'loose' enforcement (with better naming).

companion/lib/Surface/Group.ts
arikorn: Done. Added a new function to retrieve and sanitize the allowed pages list (renamed validPages to allowedPages)

companion/lib/Surface/Group.ts
arikorn: Makes sense. I took your advice, but not exactly as you suggested. After resolving your next comment, the "best" order seemed more apparent...

companion/lib/Surface/Group.ts
arikorn: OK. I think it actually works better that way. Not sure what I was thinking at the time.
Done.

docs/3_config/surfaces/pagepermissions.md
arikorn: Oops!  Done.

docs/3_config/surfaces/groups.md
arikorn: Ok, removing it all...

docs/3_config/surfaces/groups.md
arikorn: Gone!

arikorn:
Phew!  The merge at the end was a real mess, and I probably broke my local branch by doing a "partial" merge, but I did test it before merging, so it should all be good.  Hopefully you don't need any more changes before merging this one!!!

Thanks as usual, Julian, for your guidance and support on this PR.

docs/3_config/surfaces/images/restrict_pages.png
Julusian, line 1: looks like this became duplicated during the merge

docs/3_config/surfaces/pagepermissions.md
Julusian, line 1: looks like this became duplicated during the merge

PR #3741, opened 2025-10-30
feat: convert docs to docusaurus

Julusian:
This rebuilds the getting-started docs to use docusaurus. This brings a light and dark mode, easier styling, and fixes some quirks we have with urls and scroll positions.

As a follow up, I intend to build an alternate version of these docs that will be published online. This will give the exact same docs but publicly accessible, with no extra dev cost for us to update, and with the exact same styling.

The docs could do with a bit of polishing, which can happen as part of a regular docs update near to the release.

By using docusaurus for the online docs, we should be able to utilise the builtin versioning functionality of docusaurus, so that we can serve docs for not only the latest version. This will only be possible for future releases, unless someone wants to port the old docs across in the same way as these docs have been.

My intention is to bring the other documentation into a single public instance of this, so that we have a single site with all the companion related docs (including module dev)

Screenshot From 2025-10-30 18-26-56

Screenshot From 2025-10-30 18-27-09

### DIFF
--- a/companion/lib/UI/Express.ts
+++ b/companion/lib/UI/Express.ts
@@ -114,7 +114,7 @@ export class UIExpress {
 			getResourcePath('static'),
 			getResourcePath('webui/build'),
 		])
-		const docsServer = createServeStatic(getResourcePath('docs.zip'), [getResourcePath('docs/placeholder')])
+		const docsServer = createServeStatic(getResourcePath('docs.zip'), [getResourcePath('docs/build')])
 		const webuiServerWithRewriter: Express.RequestHandler = async (req, res, next) => {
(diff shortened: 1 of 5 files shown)


arikorn:
Might it make sense to merge the developer wikis into this as well, so others can help update those too?

Julusian:
@arikorn My intention is that the docs here will be mirrored to https://github.com/bitfocus/website, which in turn will publish them through an alternate docusaurus config (Demo: https://bitfocus.github.io/website/)

I think the module dev docs would be better to keep out of this repository, and to instead publish them directly on that web version. Such as under this section: https://bitfocus.github.io/website/for-developers/. Doing this is part of my overarching plan for this, as I agree the wiki makes editing the docs very tedious

PR #3742, opened 2025-10-31
Fix inconsistencies in handling of SurfacePanel.configFields

arikorn:
* Ensure all `configFields` properties are added to to panelConfig
* Ensure that defaults specified in `configFields` are honored by the WebUI
* Ensure that missing fields are added when importing a preexisting export.
* Deal with TypeScript oddities


arikorn:
It failed lint in something that is not part of this PR. Looks like something you recently pushed @Julusian.

Issue #2424, opened 2023-04-22
Prometheus endpoint


elliotmatson:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

Having a Prometheus endpoint that exposes module variables and connection status would be really helpful, that way I could log the state of a production system over time, and more effectively troubleshoot issues that happened while I was away, and would let people use tools like Grafana to visualize system status and monitor key metrics. 

### Usecases

- Getting alerts on a projector lamp nearing end-of-life, or a Hyperdeck running out of storage
- Being able to follow a video signal through different routers, switchers and devices at some point in the past for troubleshooting
- Monitoring companion connections

Julusian:
I am thinking that a MVP would be to dump connection status, and all 'defined' variables (ignoring the ones without definitions) into this endpoint.
It should be disabled by default, to reduce the risk of impact to users who will not make use of it.

As a follow up, there should be a way to reduce the amount of variables included here. That could be through defining some regexes in the settings tab, or perhaps something in the modules-api to assist with filtering

Q) In the future would it also be good to have a method in the module-api to allow for modules to add some custom fields here? Or will that be add another level of almost variables, but not really variables and make things more confusing/inconsistent?

elliotmatson:
I think the MVP sounds great! 

Maybe we could distinguish things by "variables" and "statistics"? That would help clarify the purpose of each data type. I personally can't think of any values in the modules I regularly use that would work best as a statistic and not a variable, but that's just me

voxxit:
Hi! Just found this ticket after looking for such an endpoint. However, I have to agree generally with @elliotmatson that Prometheus (or really any timeseries database) doesn't seem like the best place to query available variables, unless variables are currently being used to hold timeseries data :-) 

For instance: I know that the ZoomOSC integration stores *current* values of number of hands raised, number of users in-meeting, etc. It would be quite useful to do both:

1. Store the current value for use within the app/show it on a button/provide varying feedback, etc.
2. Expose it via Prometheus (either via a gauge as a current value, or a counter) to show, for example, "how many people showed up to my webinar today?"

Happy to give more of my use case if need be!

thedist:
One thing I'd suggest is rather than requiring Prometheus or similar, is just expose more data of the current point in time and letting users who want time-series data to collect it themselves.

[...]

PR #3743, opened 2025-11-01
feat: extended http-api  for providing all module variables as json …

JohannesRegneri:
…and prometheus metrics #2424

## Summary by CodeRabbit

* **New Features**
  * Added GET endpoints to retrieve variable definitions and values in JSON format for custom, expression, and connection variables
  * Added GET endpoints to expose variables as Prometheus metrics with automatic metric formatting and labeling

* **Documentation**
  * Updated HTTP API documentation with new variable endpoint details and Prometheus metric specifications

### DIFF
--- a/companion/lib/Service/HttpApi.ts
+++ b/companion/lib/Service/HttpApi.ts
@@ -738,7 +738,15 @@ export class ServiceHttpApi {
 		// Create a new registry for this request
 		const register = new Registry()
 
-		for (const [variableName, obj] of Object.entries(variableDefinitions)) {
+		// Create a single gauge metric for all variables
+		const gauge = new Gauge({
+			name: 'companion_variable',
+			help: 'Companion module variables',
+			labelNames: ['connection', 'variable_name', 'variable_type', 'value'],
+			registers: [register],
+		})
+
+		for (const [variableName, _obj] of Object.entries(variableDefinitions)) {
 			let value: any
 			if (isCustomVariable) {
 				value = this.#serviceApi.getCustomVariableValue(variableName)
@@ -748,53 +756,31 @@ export class ServiceHttpApi {
 				value = this.#serviceApi.getConnectionVariableValue(connectionLabel, variableName)
 			}
 
-			// Create sanitized metric name
-			const sanitizedVariableName = variableName.replace(/[^a-zA-Z0-9_]/g, '_')
-			const metricName = isCustomVariable
-				? `companion_custom_variable_${sanitizedVariableName}`
-				: isExpressionVariable
-					? `companion_expression_variable_${sanitizedVariableName}`
-					: `companion_connection_variable_${sanitizedVariableName}`
-
-			const help = isCustomVariable
-				? obj.description || variableName
-				: isExpressionVariable
-					? obj.description || variableName
-					: obj.label || variableName
+			// Convert value to string for label consistency
+			const stringValue = value !== null && value !== undefined ? String(value) : ''
 
 			// Create labels object
-			const labels: Record<string, string> = {
+			const labels = {
+				connection: connectionLabel,
 				variable_name: variableName,
+				value: stringValue,
 			}
 
-			if (!isCustomVariable && !isExpressionVariable) {
-				labels.connection_label = connectionLabel
-			}
-
-			// Handle different value types
+			// Determine gauge value: use actual number/boolean value, or 1 for info pattern
+			let gaugeValue = 1
 			if (typeof value === 'number') {
-				// Numeric values as gauges
-				const gauge = new Gauge({
-					name: metricName,
-					help: help,
-					labelNames: Object.keys(labels),
-					registers: [register],
-				})
-				gauge.set(labels, value)
-			} else {
-				// String and other values as info metrics (gauge with value 1)
-				const stringLabels = {
-					...labels,
-					value: value !== null && value !== undefined ? String(value) : '',
+				gaugeValue = value
+			} else if (typeof value === 'boolean') {
+				gaugeValue = value ? 1 : 0
+			} else if (typeof value === 'string' && value.trim() !== '') {
+				const trimmed = value.trim()
+				if (/^-?\d+([.,]\d+)?([eE][+-]?\d+)?$/.test(trimmed)) {
+					const normalized = trimmed.replace(',', '.')
+					gaugeValue = Number(normalized)
 				}
-				const gauge = new Gauge({
-					name: `${metricName}_info`,
-					help: help,
-					labelNames: Object.keys(stringLabels),
-					registers: [register],
-				})
-				gauge.set(stringLabels, 1)
 			}
+
+			gauge.set(labels, gaugeValue)
 		}
 
 		// Return metrics in Prometheus format


companion/lib/Service/HttpApi.ts
```diff
@@ -320,6 +320,14 @@ export class ServiceHttpApi {
 		// surfaces
 		this.#apiRouter.post('/surfaces/rescan', this.#surfacesRescan)
 
+		// JSON
+		this.#apiRouter.get('/appinfo/json', this.#appInfoGetJson)
+		this.#apiRouter.get('/connections/json', this.#connectionGetJson)
+		this.#apiRouter.get('/variables/:label/json', this.#variablesGetJson)
+
+		// Prometheus metrics
+		this.#apiRouter.get('/variables/:label/metrics', this.#variablesGetMetrics)
```
peternewman: I'd argue this should be Prometheus so other formats have a sensible namespace too (or maybe they should be further nested under metrics:
```suggestion
		this.#apiRouter.get('/variables/:label/prometheus', this.#variablesGetPrometheus)
```

E.g. if someone wanted to implement InfluxDB line protocol ( https://docs.influxdata.com/influxdb3/cloud-serverless/reference/syntax/line-protocol/ ) they could do it at:
/variables/:label/influx

companion/lib/Service/HttpApi.ts
```diff
@@ -320,6 +320,14 @@ export class ServiceHttpApi {
 		// surfaces
 		this.#apiRouter.post('/surfaces/rescan', this.#surfacesRescan)
 
+		// JSON
+		this.#apiRouter.get('/appinfo/json', this.#appInfoGetJson)
```
peternewman: Should this be app-info to match custom-variable above?
```suggestion
		this.#apiRouter.get('/app-info/json', this.#appInfoGetJson)
```

companion/lib/Service/HttpApi.ts
```diff
@@ -641,4 +649,164 @@ export class ServiceHttpApi {
...
+			if (value !== undefined) {
+				result[variableName] = {
+					value,
+					...obj,
+				}
+			}
+		}
+
+		if (Object.keys(result).length == 0) {
+			res.status(404).json({
```
peternewman: Should no variables be a 404? I guess the real question is can we differentiate between a module with no variables and an invalid name for a module? If we can, I'd say it shouldn't be a 404 (and maybe it's just an empty response). Especially given module variables can be dynamic.

companion/lib/Service/HttpApi.ts
```diff
@@ -320,6 +320,14 @@ export class ServiceHttpApi {
 		// surfaces
 		this.#apiRouter.post('/surfaces/rescan', this.#surfacesRescan)
```
peternewman, line 322: Adding reporting of surface state (e.g. offline) would also be really nice, but maybe that's a separate issue/PR?

companion/lib/Service/HttpApi.ts
```diff
@@ -641,4 +649,164 @@ export class ServiceHttpApi {
...
+			: this.#serviceApi.getConnectionVariableDefinitions(connectionLabel)
+
+		const result: Record<string, any> = {}
+
+		for (const [variableName, obj] of Object.entries(variableDefinitions)) {
+			const value = isCustomVariable
+				? this.#serviceApi.getCustomVariableValue(variableName)
+				: this.#serviceApi.getConnectionVariableValue(connectionLabel, variableName)
+
+			if (value !== undefined) {
```
peternewman: Aren't undefined ones also valuable? I think some of my module ones are set to undefined (rather than the empty string) when they've got no data? I don't know if there's a recommended Companion policy there?

companion/lib/Service/HttpApi.ts
```diff
@@ -641,4 +649,164 @@ export class ServiceHttpApi {
...
+				result[variableName] = {
```
peternewman: I don't know the right answer here, but in the Prometheus endpoint, you've included the connection name too.

If I've got multiple connections, talking to the same type of device (e.g. 3 Hyperdecks from the original issue), without the "namespace" being embedded in the returned JSON, I've got to do some processing before I can store them (or store by "namespace"), whereas I could just chuck all the Prometheus datapoints in a file and they're still unique. Although I appreciate it's more data being transmitted that's duplicated if I'm using them separately.

companion/lib/Service/HttpApi.ts
peternewman: I see the existing code does it for a single variable:
https://github.com/bitfocus/companion/blob/d829a131771859692928803b1d1df256ab11c456/companion/lib/Service/HttpApi.ts#L634-L635

Although maybe the distinction there is you know you've tried and it's either undefined or you've used the wrong name, here's it's just randomly missing (as you can't tell you've tried it).

companion/lib/Service/HttpApi.ts
peternewman: The existing does return a 404, but only for a single variable, I'd argue here the status should be representing the module now (and whether it exists):
https://github.com/bitfocus/companion/blob/d829a131771859692928803b1d1df256ab11c456/companion/lib/Service/HttpApi.ts#L634-L635

companion/lib/Resources/Util.ts
```diff
@@ -139,6 +139,16 @@ export const getTimestamp = (): string => {
 	return out
 }
 
+/**
+ * @returns Seconds since Jan 01 1970. (UTC)
```
peternewman: Out of interest, what's the thinking behind this rather than say ISO format?

peternewman (commented):
Exciting, a few opinion points as an API consumer/trying to think of the future a bit...

companion/lib/Service/HttpApi.ts
```diff
@@ -641,4 +649,164 @@ export class ServiceHttpApi {
...
+			},
+		})
+		return
+	}
+
+	/**
+	 * Provides Prometheus metrics for module or custom variables
+	 *
+	 */
+	#variablesGetMetrics = (req: Express.Request, res: Express.Response): void => {
```
Julusian: I'm not keen on the structure of these metrics. As an occasional prometheus user, I would find these hard to work with.

Them being called something like `custom_myvar_string` will make them very hard to find when in a prometheus with other data. At the very least that should be prefixed with `companion_`.  

But also I want to question if they should be called simply `companion_variable`, with the connection label and variable name being supplied as attributes/properties/whatever they are called.  
My reason for this is that if you want to do any graphing based on a pattern (perhaps list them all in a table or just all with the name starting `abc`), that will be very hard to do with them as separate metrics, but if they are the same metric with different properties, then they can be easily queried together while still being easy to target down to the ones you want.

companion/lib/Service/HttpApi.ts
```diff
@@ -641,4 +649,164 @@ export class ServiceHttpApi {
...
+			const value = isCustomVariable
+				? this.#serviceApi.getCustomVariableValue(variableName)
+				: this.#serviceApi.getConnectionVariableValue(connectionLabel, variableName)
+
+			const labelPrint = (isCustomVariable ? obj.description : obj.label)
+				.replace(/\\/g, '\\\\')
+				.replace(/"/g, '\\"')
+				.replace(/\n/g, '\\n')
+
+			result += `# HELP ${connectionLabel}:${variableNamePrint} ${labelPrint}\n`
```
Julusian: Personally, I would be tempted to use https://github.com/siimon/prom-client to build the response of this request, to avoid needing to manually handle this encoding and syntax structure.  
I don't know if it can do what we need here though, I have only used it for an application wide endpoint.

companion/lib/Service/HttpApi.ts
Julusian: I think that in general it is reasonable to not include the connection prefix, as that will make writing generic processing/parsing code simpler without having to make that parsing aware of the connection label.  
In your case of working with multiple, it sounds reasonable to have to do a little bit of namespacing in your app, you may want to use different labels/ids too.

Prometheus is different, as it is expected they will be put into a global store, where it is likely that other connections/data will exist, so they need to be scoped. (although maybe that could be added by the scraper rule, so maybe not necessary to be builtin?)

companion/lib/Service/HttpApi.ts
```diff
@@ -641,4 +649,164 @@ export class ServiceHttpApi {
...
+	 *
+	 */
+	#appInfoGetJson = (_req: Express.Request, res: Express.Response): void => {
+		this.logger.debug(`Got HTTP /api/appinfo/json`)
+		res.json({
+			success: true,
+			message: 'App information retrieved successfully',
+			timestamp: getTimestampUnix(),
+			data: {
+				appinfo: this.#serviceApi.appInfo,
```
Julusian: I don't like this.  
This `appInfo` object is an internal object that has no guarantees of being at all consistent between versions.  
A majority of its contents I can't see as being of any value to a caller.

I am not against exposing some of this data, but I think it needs to be done in a more controlled fashion, with only the useful bits being exposed

companion/lib/Service/HttpApi.ts
```diff
@@ -641,4 +649,164 @@ export class ServiceHttpApi {
...
+			},
+		})
+		return
+	}
+
+	/**
+	 * Provides JSON for connection status
+	 *
+	 */
+	#connectionGetJson = (_req: Express.Request, res: Express.Response): void => {
```
Julusian: I don't like how this method works.
Searching for variables like this feels rather brittle and messy.

It would be much cleaner to expose direct access to some methods/events/classes needed to drive this.

I can say that some of this will for sure break a bit with the surface module work which is ongoing, as they are 'instances' too, so will appear in some of these variables

companion/lib/Service/HttpApi.ts
peternewman: Fair enough, could we compromise on at least including the connection prefix once in a "header/request" type space, then consumers don't need to track the request/response pair to know what bit of data it was. Then even Companion's Generic HTTP module could process it! 😆 

> Prometheus is different, as it is expected they will be put into a global store, where it is likely that other connections/data will exist, so they need to be scoped. (although maybe that could be added by the scraper rule, so maybe not necessary to be builtin?)

I don't know enough about Prometheus, is the store per host being monitored, or global to the Prometheus instance? Obviously if it's the latter you've still got to add context as soon as you're checking two different Companion instances with the same type of connection configured on both (unless you've already made the connection names globally unique there).

companion/lib/Service/HttpApi.ts
peternewman: Again I don't know enough about Prometheus, but thinking of something I'm aiming to do with a different monitoring solution is extracting health info like temperature from connection variables (across multiple devices of the same type) and graphing that. I assume a use case like that would still work with @Julusian suggestion?

Would you want to add the connection type too to the attributes, so you can graph all the FooBar 5000 temperatures together (without needing to know how each connection is labelled)?

companion/lib/Service/HttpApi.ts
Julusian: I think I'm agreeing with you on this endpoint not including connection label. Prometheus wont understand json though, so is a parallel discussion

----

Its a global store. As an example, if I query my prometheus with `node_boot_time_seconds{}`, I get back:
```
node_boot_time_seconds{instance="10.0.0.1:9100", job="node-a"} = 1763253353.944076
node_boot_time_seconds{instance="10.0.0.2:9100", job="node-b"} = 1763252538
node_boot_time_seconds{instance="10.0.0.3:9100", job="node-c"} = 1763237998
node_boot_time_seconds{instance="10.0.0.4:9100", job="node-d"} = 1762635123.979674
node_boot_time_seconds{instance="10.0.0.5:9100", job="node-e"} = 1762635247
```

I believe that in this case, both the `instance` and `job` are being injected by the scraping and are not in the source. But the source file is also able to specify other fields, eg `node_cpu_seconds_total{}` returns values looking like: `node_cpu_seconds_total{cpu="0", instance="10.0.0.1:9100", job="node-a", mode="idle"} = 0`

But the bit that worries me is a module producing a metric called simply `ip` pollutes the global store quite a bit, so they want to have better names in some way. Those two metrics I showed are from the `node` exporter (linux system), general convention appears to be to at least prefix it with `companion_`, but even that will still make quite a mush. Maybe the best thing would be `companion_connection_${moduleType}_variable_${variableName}` as that is probabyl closest to the typical structure.  
That would avoid embedding the label in the name, while also allowing the metrics to have accurate descriptions (if we ignore that different versions of a module could vary)

companion/lib/Service/HttpApi.ts
Julusian: I think my answer https://github.com/bitfocus/companion/pull/3743#discussion_r2531980911 covers some of this.  
I wrote that before reading this, but yeah I think my proposal on naming will achieve what you want as then:
* the variables will be grouped as metrics based on what they are (ie, how the module defines them), with the connection label/id/whatever being available as a label on the metric.
* These labels on metrics are commonly used to filter, or you can query and ignore the label.
* There is no guarantee that a `temperature` variable will be at all consistent between module types. some will be C some F, some strings, some numbers, so having them grouped by module will both force and allow you to handle that explicitly

For example, a snippet from a grafana graph I have:
image
That first query returns ~10 sensors of data, the second query matches one or two). The `device` label is being used to know how to group values within the metric

JohannesRegneri:
@coderabbitai review

PR #3744, opened 2025-11-02
fix: Build Companion for Windows in WSL including General Webpack Issues

arikorn:
1. electron-builder requires wine to build Windows from Linux. I will add an issue with link to my suggested changes in the Wiki once this is merged
2. @napi-rs needed to load _canvas-win32-x64-msvc_ prebuild.  This is solved by changing the os to `os: platformInfo.nodePlatform,` in .yarnrc.yml
3. retrieval of prebuilds/better_sqlite3.node needs to specifify platform as well as arch. (Also, the script needed to reset the node_modules version of the prebuild back to the Linux version so that `yarn dev` continues to work.)
4. Two source files reference `import.meta.url` in a way that webpack converts to a hard-coded string. One is in Companion source code; the other is in @sentry/node-core. I wrote ~~a yarn patch for the latter and will see if I can get it added in Sentry~~ workaround, for now, and will see if I can get it fixed at the source (sentry or webpack).
5. Add execute permissions to node.exe so that launch check can work.

~~note: in the process of applying the patch, yarn appears to have "upgraded" 10.17.0 to 10.22.0; it doesn't appear to hurt anything.~~

### DIFF
--- a/companion/lib/Graphics/Controller.ts
+++ b/companion/lib/Graphics/Controller.ts
@@ -50,11 +50,6 @@ function generateFontUrl(fontFilename: string): string {
 	const fontPath = isPackaged() ? 'assets/Fonts' : '../../../assets/Fonts'
 	// we could simplify by using import.meta.dirname
 	return fileURLToPath(new URL(path.join(fontPath, fontFilename), import.meta.url))
-	// if (isPackaged()) {
-	// 	return path.join(__dirname, 'assets/Fonts', fontFilename)
-	// } else {
-	// 	return fileURLToPath(new URL(path.join('../../../assets/Fonts', fontFilename), import.meta.url))
-	// }
 }
 
 interface GraphicsControllerEvents {
--- a/companion/webpack.config.js
+++ b/companion/webpack.config.js
@@ -64,7 +64,7 @@ export default {
 	},
 	externalsPresets: { node: true },
 	externals: {
-		// Native libs that are needed (note: the commonjs2 modifier is incompatible with `output.module: true` )
+		// Native libs that are needed
 		usb: 'usb',
 		bufferutil: 'bufferutil',
 		'@serialport/bindings-cpp': '@serialport/bindings-cpp',


companion/lib/Instance/ProcessManager.ts
```diff
@@ -18,7 +18,10 @@ import { ModuleInstanceType, type InstanceConfig } from '@companion-app/shared/M
 import { assertNever } from '@companion-app/shared/Util.js'
 import type { SomeModuleVersionInfo } from './Types.js'
 
-const require = createRequire(import.meta.url)
+// `import.meta.url` gets hardcoded to the build time file by webpack (or so Google's AI Overview insists... regardless, it _is_ getting hardcoded)
+//  so it puts the build-time path into main.js.  The following "trick" fixes it.
+// (Unfortunately, I haven't found a way that skips hardcoding the filename. Using '.' or '' in new URL() fails; the latter fails after build. And fileURLToPath() doesn't help either.)
+const require = createRequire(new URL('ProcessManager.js', import.meta.url))
```
Julusian: All we use this for is some `require.resolve` calls (and for loading in some json files, but that could easily be done with fs.readFile and JSON.parse)
so perhaps there is a better way to do this that will avoid needing to use import.meta.url entirely?

Or maybe some other 'trick' can be done to this. 

This is not the first patch of fighting paths once run through webpack, so it might be that I need to do a larger task of either replacing webpack, or figuring out how to propelry counter webpacks mangling.  
If you search around, there are many places that use one path when `isPackaged()` and another when in dev

tools/build/dist.mts
```diff
@@ -107,12 +107,15 @@ await fs.writeFile(
 )
 await fs.copyFile('yarn.lock', 'dist/yarn.lock') // use the same yarn.lock file, to keep deps as similar as possible
 await fs.copyFile('.node-version', 'dist/.node-version')
+// note: `os: ... platformInfo.nodePlatform` is needed to load @napi-rs/canvas-win32-x64-msvc when building win32 from WSL, perhaps for any cross-platform build?
+//     os: ['current', ...] does not appear to be needed, but if used, perhaps remove unneeded prebuild in package.mts?
+//     Do we need to force arm64?? (if so, how? add to cpu?)
```
Julusian:
```suggestion
```

I dont think this comment adds anything, the `cpu: platformInfo.nodeArch,` line is already handling the x64/arm64 switch, so this change just fills the bit I overlooked/neglected on platform

tools/build/package.mts
```diff
@@ -67,7 +67,10 @@ if (platformInfo.runtimePlatform === 'win') {
 		}
 	}
 
-	const prebuildDirs = await glob('dist/**/prebuilds', { onlyDirectories: true, expandDirectories: false })
+	let prebuildDirs = await glob('dist/**/prebuilds', { onlyDirectories: true, expandDirectories: false })
+	// Note: adding "dist/node_modules/@napi-rs" to this list should never hurt
+	// but it only does anything when os: 'current' is included in 'dist/.yarnrc.yml' (see dist.mts)
+	prebuildDirs.push('dist/node_modules/@napi-rs')
```
Julusian, line 70: Is this needed/doing anything useful?

In the current builds (based on the last one I ran locally a week or so back), no napi-rs prebuilds are copied into `dist/prebuilds`.  
`@napi-rs/canvas` is being added as a dependency in the package.json, and installs the correct binaries through whatever we force in `.yarnrc.yml`

package.json
```diff
@@ -86,6 +86,8 @@
     "@xkeys-lib/core@npm:3.3.0": "patch:@xkeys-lib/core@npm%3A3.3.0#~/.yarn/patches/@xkeys-lib-core-npm-3.3.0-1a65b6423b.patch",
     "unix-dgram": "npm:empty-npm-package@1.0.0",
     "bare-fs": "npm:empty-npm-package@1.0.0",
-    "bare-path": "npm:empty-npm-package@1.0.0"
+    "bare-path": "npm:empty-npm-package@1.0.0",
+    "@sentry/node-core@npm:10.17.0": "patch:@sentry/node-core@npm%3A10.22.0#~/.yarn/patches/@sentry-node-core-npm-10.22.0-e586fad107.patch",
+    "@sentry/node-core@npm:10.22.0": "patch:@sentry/node-core@npm%3A10.22.0#~/.yarn/patches/@sentry-node-core-npm-10.22.0-e586fad107.patch"
```
Julusian: These overrides worry me, as sentry is one of the dependencies that updates most frequently, and I will forget to update this patch, which result in it silently being dropped without me noticing.

I don't have a good solution to this though.. Might be a case of needing to fight webpack config?  
Just thinking out loud of some things I might try; Perhaps webpack can do similar mangling of a dependency as it bundles it?

companion/lib/Instance/ProcessManager.ts
arikorn: Yeah, I think the problem was that `require.resolve` is actually adding _node_modules/_ to the path, so it's doing a little more than just joining paths.  However, if your objection is to hardcoding the filename in the file, I do have another solution that works: move `createRequire` to the place it's used and use `moduleInfo.basePath` as the argument. In some sense this might even be more appropriate. The only tradeoff is that `createRequire()` is now run once for each module, rather than once per session. OTOH, a benefit is that the variable doesn't hang around once it's no longer needed.
Pushed in its own commit, so you can see and judge...

tools/build/dist.mts
arikorn: Done. Yes, the comment was mainly there for review purposes, particularly in case you needed `current` for some other configuration..

tools/build/package.mts
arikorn: Correct! I'm not sure if the fact that it's not put into _prebuilds_  is a bug or not, but os-specific code for @napi-rs canvas _is_ loaded into subfolders of `dist/node_modules/@napi-rs` so for Windows there's a folder called  `dist/node_modules/@napi-rs/canvas-win32-x64-msvc` and for linux there's a folder called `dist/node_modules/@napi-rs/canvas-linux-x64-gnu`
I can remove these lines if you're sure you'll never need to add 'current' to the `os:` list in dist/.yarnrc.yml , but it's harmless as it is, so may just keep it? (It's harmless to remove too, of course, just leaves some unused files.)

companion/lib/Instance/ProcessManager.ts
Julusian: yeah I think that moving createRequire like that makes sense to do.
This is only used for dev modules, so I am not worried about the cost of the call

tools/build/dist.mts
Julusian: I cant see why we would need current for anything; this yarnrc is being used solely to install the node_modules to be shipped with companion.

package.json
arikorn: Working on this, but if all else fails, what do you think about just running main.js through `sed` to replace the offending string? (It would go in package.mts right before the launch check.)

arikorn:
OK, I've spent enough time trying to fix the webpack config to no avail... So far, replacing the bad file-string in main.js is the only workaround I've found that works without patching sentry. There is a glimmer of possibility below, and maybe I can abstract the problem enough to present to the webpack or sentry developers. They both appear to welcome to feedback.

The ["correct"](https://webpack.js.org/configuration/module/#moduleparserjavascriptimportmeta) way to deal with webpack's handling of `import.meta.url` is to [add the following](https://github.com/morganney/import-meta/blob/main/webpack.config.js) to companion/webpack.config.js:

~~~javascript
		parser: {
			javascript: {
				importMeta: false,
			},
		},
~~~
and maybe also (from the example linked above):
~~~javascript
  plugins: [
    new HtmlWebpackPlugin({
      scriptLoading: 'module'
    })
  ]
~~~
`importMeta: false` does, in fact, prevent the conversion of `import.meta.url` to a compile-time string. However, the code still breaks at runtime with a spurious error message of the first encountered import.meta... (this one happens to be in the sentry module, original code):

~~~
<anonymous_script>:31
      module__WEBPACK_IMPORTED_MODULE_3__.register('import-in-the-middle/hook.mjs', import.meta.url, {
                                                                                           ^^^^

SyntaxError: Cannot use 'import.meta' outside a module
    at 32062 (\\wsl.localhost\Ubuntu\home\ari\Git-projects\companion\dist\main.js:1:10160412)

~~~

The file is clearly a proper module.  "Fixing" this one by patching as I did previously just uncovers the next place it complains about -- this time for `import.meta.dirname` which is apparently in @julusian/image-rs/bindings.js... (Adding the recommended plugin shown above didn't help either)

Several issues related to this in vercel/ncc (which uses webpack under the hood) are still open, though not necessarily the same: [here](https://github.com/vercel/ncc/issues/791) and [here](https://github.com/vercel/ncc/issues/1123), and[ a closed one at webpack itself](https://github.com/webpack/webpack/issues/16029), though it's not entirely clear that it should have been closed... and an [open one](https://github.com/webpack/webpack/issues/16878) that may be slightly different. (The advice, there and [here](https://github.com/webpack/webpack/issues/16802) to set `module.parser.javascript.url: false or 'relative'` doesn't help either, FWIW.)

Other webpack config ideas that didn't help make `importMeta: false` work or substitute for it:
- Adding `/* webpackIgnore: true */` ([see here](https://webpack.js.org/api/module-methods/#webpackignore)), which would still require a patch.  
-  removing `experiments.topLevelAwait` (suggested somewhere) didn't help.
- open-wc's `webpack-import-meta-loader`, crashed on undefined `window.location.protocol`. Might be patchable? Note that this is [legacy and 5 years old](https://open-wc.org/docs/legacy/legacy-projects/).
- `import-meta-loader`, a slightly newer package didn't do anything (and to be fair, its docs implied that it wouldn't).
- using string-replace to manually replace the hardcoded filename with import.meta.url behaves the same as if `importMeta: false` were set

 Some other advice that might help fix `importMeta: false` includes
-  doing something in a babel config (do we use babel directly?). or use babel-plugin-bundled-import-meta or babel-plugin-transform-import-meta? (the latter seems more popular). If these work in webpack.config.js directly, I couldn't figure it out.
- Google AI Overview: "Custom Babel Transforms or Loaders: If more granular control is required or if the above methods are insufficient, a custom Babel transform or a Webpack loader can be developed. These tools can specifically target import.meta.url"

~~FWIW, adding `/* webpackIgnore: true */` _does_ work as long as the config does not set `importMeta: false,` as above. So at least it may be more palatable than hardcoding the sourcefile name. (And presumably can be used in ProcessManager.ts, too, if you want to minimize changes.)~~ (could not reproduce)

Ideally I'd either find a minimal example to present to webpack (an/dor sentry) developers, or I can convince sentry to put in the innocuous comment.

_A testing note_: when I run the packaged code on Windows it says "Sentry error reporting is disabled". This is correct and unrelated behavior because the SENTRY file is empty (I gather something on your end fills it in with a web address.) If I copy the SENTRY file from an "official" build then it correctly says "Sentry error reporting configured".

package.json
arikorn: See long comment, in the PR. So far the only thing that works without patching Sentry is to do a string-replace on main.js. The latest commit does it.

arikorn:
Copying commit comment, since it says a bunch of things:

Replace sentry patch with manual code replacement, update various files to ESM

- Update several files to ESM as part of a futile attempt to fix webpack behavior when `module.parser.javascript.importMeta` is set to false. It didn't help but is probably a good thing to keep, regardless... I also fixed several typescript complaints.
- Fix ProcessManager.ts by removing the reference to `import.meta.url` (undo previous commit's undo)
- Fix @sentry/node-core code by replacing the incorrect file path in the bundled _main.js_ file.

companion/lib/Instance/ProcessManager.ts
arikorn: >If you search around, there are many places that use one path when isPackaged() and another when in dev

I suspect that some of those have been resolved in webpack. I can look through them all once this PR is finalized, If you'd like.

package.json
arikorn: Just to be clear: the Sentry patch has been removed in the latest commit, so the original concern is resolved..

arikorn:
1. The latest commit removes the need for the search-and-replace fix to @sentry/node-core code
2. The code currently runs correctly with webpack set to 'development' mode - hopefully the webpack developers will be able to help unravel the issue with 'production' mode.
3. Although this started out as a fix for WSL->Windows compilation, much of the recent work fixes long-standing issues with Companion's interactions with webpack for all platforms. This was necessary to remove the search-and-replace hack, above!

arikorn:
For cross-reference:
[My issue in webpack](https://github.com/webpack/webpack/issues/20088#issuecomment-3488554224)
and
[a PR from them trying to fix the remaining bugs](https://github.com/webpack/webpack/pull/20107)

arikorn:
Thanks to assistance from the webpack developers I have a working solution (pushed): disable a single optimization. As far as I can tell, removing the `concatenateModules` optimization makes almost no difference at all (file sizes are essentially the same).

From the [webpack docs](https://webpack.js.org/configuration/optimization/#optimizationconcatenatemodules):
>Tells webpack to find segments of the module graph which can be safely concatenated into a single module.

(Github found no errors in this PR -- it's only complaining about docs/3_config/buttons/button_indicators.md )

companion/lib/Graphics/Controller.ts
```diff
@@ -96,7 +99,8 @@ export class GraphicsController extends EventEmitter<GraphicsControllerEvents> {
 	readonly #renderQueue: ImageWriteQueue<string, [RenderArguments, boolean]>
 
 	#pool = workerPool.pool(
-		isPackaged() ? path.join(__dirname, './RenderThread.js') : fileURLToPath(new URL('./Thread.js', import.meta.url)),
+		// note: import.meta.url can be replaced with import.meta.directory as long as we use node v22.16 and later
```
Julusian, line 97: our current minimum nodejs version is 22.21.1, so I guess this change could be made.

I havent been paying attention to the development of `import.meta.dirname` or `import.meta.filename`, so maybe most/all of these uses of `import.meta.url` and `fileURLToPath` can be simplified to be something like: `path.join(import.meta.dirname, isPackaged() ? './RenderThread.js' : './Thread.js')`

Be aware that for some we do want to avoid any if/ternary logic within the construction of the paths, as we need webpack to build and copy the file to the output directory. And for others we explicitly don't want it to do that, so are forcing a conditional into it to make it unparsable

arikorn: FWIW, `import.meta` preprocessing had to be disabled to get other things working. (Even the webpack folks agreed that this was a reasonable approach to our situation.) So in regard to the last paragraph, I don't think the ternary logic should make a difference any more one way or the other anymore.

Just to copy one thing I wrote in slack, for the record: based on webpack-developers' comments, webpack is "optimized" for web applications, and it is therefore perfectly reasonable that we need to adjust settings when running as a distributable node application. Hopefully it's all done for now, or at least, future issues should be easier to resolve! 

Since we've agreed to do the `url` to `dirname` in a separate PR, I believe all issues in this conversation have been addressed. As always, please reopen if you disagree.

tools/build/package.mts
arikorn: I'm resolving this one, because what it does is harmless at worst, and it is slightly useful in that it "documents" the existence of a platform-specific binary that is not in the prebuild directory. 

Please feel free to delete those lines if you disagree. (Or tell me to...)

arikorn:
OK, I've resolved all outstanding comments and fixed the merge conflict so if you agree, this PR should be good to go.

Thanks, as always, Julian!

companion/lib/Graphics/Controller.ts
```diff
@@ -47,11 +47,14 @@ export interface GraphicsOptions {
...
-		return fileURLToPath(new URL(path.join('../../../assets/Fonts', fontFilename), import.meta.url))
-	}
+	const fontPath = isPackaged() ? 'assets/Fonts' : '../../../assets/Fonts'
+	// we could simplify by using import.meta.dirname
+	return fileURLToPath(new URL(path.join(fontPath, fontFilename), import.meta.url))
+	// if (isPackaged()) {
+	// 	return path.join(__dirname, 'assets/Fonts', fontFilename)
+	// } else {
+	// 	return fileURLToPath(new URL(path.join('../../../assets/Fonts', fontFilename), import.meta.url))
+	// }
```
Julusian:
```suggestion
	const fontPath = isPackaged() ? 'assets/Fonts' : '../../../assets/Fonts'
	// we could simplify by using import.meta.dirname
	return fileURLToPath(new URL(path.join(fontPath, fontFilename), import.meta.url))
```

companion/webpack.config.js
```diff
@@ -0,0 +1,141 @@
...
+		// 		stream: false,
+		// 		zlib: false,
+		// 		timers: false,
+		// 		http: false,
+		// 		https: false,
+		// 	},
+	},
+	externalsPresets: { node: true },
+	externals: {
+		// Native libs that are needed (note: the commonjs2 modifier is incompatible with `output.module: true` )
```
Julusian:
```suggestion
		// Native libs that are needed
```

companion/webpack.config.js
```diff
@@ -0,0 +1,141 @@
+import path from 'path'
+import fs from 'fs'
+import { sentryWebpackPlugin } from '@sentry/webpack-plugin'
+import { fileURLToPath } from 'url'
+// import { createRequire } from 'module'
+// const require = createRequire(import.meta.url)
```
Julusian, line 6:
```suggestion
```

arikorn:
Hooray!!! Thanks for seeing this one through Julian!

arikorn:
1. For the record, I compared a build from just before to just after merging this PR.  After fixing the difference in the `devtool` value, file sizes are essentially unchanged (6 KB larger for main.js + RenderThread.js out of 5.6 MB). However the old webpack process also creates five files:

~~~
  assets by info 3.61 KiB [immutable]
    asset 91e42a4e979510d2957b.json 3.43 KiB [emitted] [immutable] [from: package.json] (auxiliary name: main)
    asset 802c72f5d48da0e10e46.js 130 bytes [emitted] [immutable] [from: dist/Graphics/Thread.js] [minimized] (auxiliary name: main)
    asset 4134ae6556cf5bfb22c5.json 47 bytes [emitted] [immutable] [from: ../nodejs-versions.json] (auxiliary name: main)
  asset BUILD 60 bytes [emitted] [from: ../BUILD] (auxiliary name: main)
  asset SENTRY 0 bytes [emitted] [from: ../SENTRY] (auxiliary name: RenderThread, main)
~~~

This PR creates the last two in _build_writefile.mts_. The first three, I'm pretty sure are never referenced in the packaged code:

- "_package.json_" presumably from [companion/lib/Registry](https://github.com/bitfocus/companion/blob/47ae09fd85e3ef18730bcef0b7a788cc401d1c8e/companion/lib/Registry.ts#L31) -- except that somewhere in v4.2 beta it was changed to point to the package.json generated by dist.mst during the package-build process.
- (note: a package.json is also referenced in companion/lib/Instance/ProcessManager.ts, but this refers to a file in node_modules, which is not the file copied in by webpack, above.) 
- " _dist/Graphics/Thread.js_", presumably referenced from [companion/lib/Graphics/Controller.ts](https://github.com/bitfocus/companion/blob/47ae09fd85e3ef18730bcef0b7a788cc401d1c8e/companion/lib/Graphics/Controller.ts#L96), is conditioned on isPackaged() false
- "_../nodejs-versions.json_",   presumably referenced from c[ompanion/lib/Instance/NodePaths.ts](https://github.com/bitfocus/companion/blob/47ae09fd85e3ef18730bcef0b7a788cc401d1c8e/companion/lib/Instance/NodePath.ts#L9), states in the comments that it's not used when packaged.

2. Another note: I was probably mistaken in my comments on `output.clean` that webpack is called twice. It appears to be only called for the "server" files, not the webui files. This comment is purely academic, though: what we're currently doing is probably safer for our process.

3. Finally, Alexander at webpack says that [they now have plans to add an option](https://github.com/webpack/webpack/issues/16878#issuecomment-3528069219) to leave `import.meta.url` alone when not in `new URL()` calls. This might be a best-of-both worlds compromise, i.e. do some packaging magic when possible; just work correctly otherwise.  He says, however, it will not process `import.meta.pathname` code, since they only do it for resolving modules and not in `path.join()`.

Issue #3733, opened 2025-10-28
Improvements for Full Imports


arikorn:
### Is this a feature relevant to companion itself, and not a specific module?

- [x] I believe this to be a feature for companion, and is not specific to a module

### Is there an existing issue for this?

- [x] I have searched for similiar existing issues

### Describe the feature

The main purpose of this issue is to consolidate and/or expand upon some issues with "Full Import" (mostly discussed in the recently merged PR #3708) with a goal of making sure I address them all in a future PR or two! Note that although these are largely motivated by the newer partial-reset functionality, it affects full reset functionality as well.

1. **Allow the user to select connections**.  Currently connections _must_ be imported with a full import whether-or-not a full reset is being performed (and despite the fact that a user could create an export file with no connections in it, which is effectively the same as the ability to deselect connections here.) The problem is that clearing the current connections before importing them will delete all of the module-related actions in all pages & triggers. To avoid this unwanted effect when using partial reset, Companion will not import any connections (or connection collections) that are currently in the active configuration/database

   **Proposed solution part 1a**: warn the user if they select pages or triggers but not connections (this is already done in the "Reset Configuration" button workflow.)  ...or pages but not variables & triggers?... etc. A more advanced method might scan the pages and triggers to see if they reference connections... (@Julusian wrote: "Showing that an import will leave broken references may not be too bad to do actually, the challenge becomes how to show that in the ui, and how to help them fix it.")

   **Proposed solution part 1b**: When loading connections with partial reset, use the "merge" functionality from the single-page import flow (aka Buttons tab).  Note that if the user is importing all of connections, pages, expression variables and triggers, (or if they contain no module-references) there is no need to request merge preferences. 

2. **Address internal object mismatches** currently the database, import file and WebUI options each have different fields or different field names for the various components. This can make changing the code more brittle. (In particular in `importFull` we use three variables: `config`, `data`, and `resetArgs` with possibly overlapping function)  Separately, perhaps `data` should be reorganized to keep collections with their components? @Julusian writes "The intention being to make it a more explicit multi-step transform so that the final step of performing the import isn't trying to match stuff up across 2/3 objects, but can just look at the one."

   **Potential solution to issue 2**: Perhaps write a "matching function" that resides in "ImportExport.ts" so it lives together with the "mismatched" objects. The function could also do some basic validation to ensure no new fields have been added (`assertNever`??)...

[...]

PR #3747, opened 2025-11-02
feat: rework import and reset types

Julusian:
This reworks the api for doing a full import or full reset. Partially in preparation for a larger rework of the import, and also in preparation of handling more chunks with some granularity things which want to describe 

This is also rebuilding some of the import and reset ui to use tanstack form instead of the manual state tracking, as this makes some parts of the forms simpler to handle correctly (in particular updating the heirarchy state).

### DIFF
--- a/companion/lib/ImportExport/Import.ts
+++ b/companion/lib/ImportExport/Import.ts
@@ -6,7 +6,11 @@ import type {
 } from '@companion-app/shared/Model/ExportModel.js'
 import type { ControlsController } from '../Controls/Controller.js'
 import { CreateExpressionVariableControlId, CreateTriggerControlId } from '@companion-app/shared/ControlId.js'
-import type { ClientImportOrResetSelection, ConnectionRemappings } from '@companion-app/shared/Model/ImportExport.js'
+import type {
+	ClientImportOrResetSelection,
+	ConnectionRemappings,
+	ImportOrResetType,
+} from '@companion-app/shared/Model/ImportExport.js'
 import {
 	fixupControl,
 	fixupExpressionVariableControl,
@@ -131,6 +135,8 @@ export class ImportController {
 	}
 
 	importFull(data: ExportFullv6, config: ClientImportOrResetSelection): void {
+		const isImporting = (value: ImportOrResetType): boolean => value === 'reset-and-import'
+
 		const mergeConnections = config.connections === 'unchanged'
 
 		// Always Import instances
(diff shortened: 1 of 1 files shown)


Julusian:
@arikorn This has led me down the path of thinking that you might have been right about trusting the ui to provide sane values.  
This is largely because the ui must now provide either `'unchanged', 'reset-and-import', 'reset'` for each thing instead of a boolean, and I dont see a sane way of correcting 'bad' values in this (while respecting that a future ui will allow choosing between all 3 modes; today it will be limited to unchanged/reset-and-import)

arikorn:
@Julusian wrote:
>@arikorn This has led me down the path of thinking that you might have been right about trusting the ui to provide sane values.

😊 Thanks for the shout-out Julian!

> the ui must now provide either `'unchanged', 'reset-and-import', 'reset'` for each thing 

Hmm... if you think about the 2x2 matrix for import vs reset, should there be a fourth value in `zodImportOrResetType`?

|      |      | import|   |
|---|---: | :---: | :---: |
|          |   |  **_no_** | **_yes_** |
|**reset**|_**no**_ |unchanged| ?? |
|        |**_yes_**   | reset-only | reset&import|
               
I would call the missing value (import w/o reset) "merge", (as you are currently doing with connections).  I think in your notes you call it "custom."  And of course there are 2-3 types of merges: merge preferring existing, merge preferring imported. (Similar to the import single page options.), or if one wanted to get totally crazy: merge like git does...  Even if you're not intending to implement it in this PR, the vlaues might as well go in. Though, as much as I hate to ask it: wouldn't it be simpler to just have two variables -- "import" and "reset"? (Even in the ui, then you just have two columns of checkboxes; "reset" is always enabled, "import" is controlled as is done today...)

BTW, for the sake of cross-reference, this PR/branch appears to be addressing item 2 in our issue #3733 

Also, I'm wondering if this would be a good opportunity to rename some of the WebUI properties to match the data, so `pages` instead of `buttons` in what is now `zodClientImportOrResetSelection `. Ideally custom_variables instead of customVariables, but that, admittedly, is unfortunate. I wonder how bad it would be to rename the data object property instead?

Julusian:
> I would call the missing value (import w/o reset) "merge", (as you are currently doing with connections). I think in your notes you call it "custom." And of course there are 2-3 types of merges: merge preferring existing, merge preferring imported. (Similar to the import single page options.), or if one wanted to get totally crazy: merge like git does... Even if you're not intending to implement it in this PR, the vlaues might as well go in. Though, as much as I hate to ask it: wouldn't it be simpler to just have two variables -- "import" and "reset"? (Even in the ui, then you just have two columns of checkboxes; "reset" is always enabled, "import" is controlled as is done today...)

Yes, there should be a 4th option, but I think what that looks like will want to vary depending on the property it is for.  
For example, for `connections` it would want to contain something to replace `connectionIdRemapping: Record<string, string | undefined>`. Everything else will also want some kind of object to describe what to do in a merge, whether that it providing id/name aliases, or specifying what to import and what not to.  

That is all very tied to that mutli-step ui overhaul I proposed though, and there isnt any point doing a backend for it until there is a ui to drive it (which will also help define what it should be)

arikorn:
Incidentally, my new template for the Contour Shuttle (in docs) is a good example of a use-case for non-resetting (i.e. merging) imports: the companionconfig file defines the variables, the surface association (though sadly, only for the v1 Pro), as well as the page. So the user may want to add these to their database rather than make them the only thing. (I suppose it also, or maybe more closely, relates to single-page import, to add the variables references on that page, which is an issue already, if I recall correctly...)

Issue #3737, opened 2025-10-29
[BUG] Event On Sunrise/Sunset wrong time after DST switch (Companion)


ParasolDigitalOperations:
### Make sure you're on the latest stable or beta build

- [x] I have tested this on the latest stable or beta release

### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion and not a specific module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Since the recent Daylight Saving Time change, the Sunset trigger in Companion fires one hour off.
My computer’s system time and timezone are correct (CET/CEST), and coordinates for the location are set properly. Before the DST change everything worked as expected.

### Environment (please complete the following information)

```markdown
- OS: Debian Linux
- Browser: Brave 1.83
- Companion Version: v4.1.4
```

peternewman:
To ask the obvious question, have you restarted Companion since the clock change? I.e. is this a special case with Companion not updating the sunrise/sunset over a clock change, or they are permanently calculated incorrectly in winter or summer?

ParasolDigitalOperations:
> To ask the obvious question, have you restarted Companion since the clock change? I.e. is this a special case with Companion not updating the sunrise/sunset over a clock change, or they are permanently calculated incorrectly in winter or summer?

Yes, both the system (OS) and the application have been restarted.
In my case, the trigger still does not work. Even a newly created trigger with similar parameters shows the same behavior.

coolsmarty22:
Alternatively you can set the offset to -60 for workaround. This works for me.

ParasolDigitalOperations:
> Alternatively you can set the offset to -60 for workaround. This works for me.
that's what I did

Julusian:
There is a fix for this in v5.0.2. Hopefully it works, I don't know how to go about testing it

PR #3754, opened 2025-11-04
fix: resolve DST bug in sunrise/sunset trigger timing (#3737)

darinpope:
The original implementation of sunrise/sunset event timing had a critical DST bug where event times would jump 1+ hours across DST transitions (issue #3737). The root cause was using getTimezoneOffset() in day-of-year calculations, which changes during DST transitions.

### DST Fix Implementation (Timer.ts)
- Replaced timezone-offset-based day calculation with dayjs for robust calendar math
- Changed time creation from local timezone to UTC-based Date.UTC() approach
- Applied offset in UTC minutes instead of local minutes to avoid DST issues
- This ensures sunrise/sunset times remain stable across DST boundaries

### Testing
- Added comprehensive test suite with 536 tests covering:
  - 20 geographic locations across 6 continents (20 locations × 10 years × 2 event types = 400 baseline tests)
  - DST transitions in both spring and fall for multiple hemispheres
  - Non-DST timezones (Hawaii, Arizona, Iceland, Moscow, India, Nepal)
  - Edge cases: leap years, year boundaries, polar regions, equinoxes/solstices
  - Offset precision: positive/negative offsets, boundary values, extreme values
  - Multiple Timer instances, idempotency, concurrent updates
  - Invalid inputs: NaN, Infinity, extreme coordinates (±90°, ±180°)

### Debug Logging
- Added informative debug logging on setSun() with:
  - Event type (Sunrise/Sunset)
  - Location coordinates (latitude, longitude)
  - Offset in minutes
  - Next scheduled time in both local and UTC formats

### New Test Categories
- Offset Precision Near DST Boundaries
- Multiple Independent Timer Instances
- Large Negative Offset Edge Cases
- Absolute Idempotency with State Verification
- Maximum Valid Coordinate Precision

In addition to automated tests, also tested interactively with numerous lat/long combinations.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed sunrise/sunset timer scheduling to eliminate daylight saving time-related drift, ensuring more reliable and accurate event timing.


peternewman:
I'll admit I've not looked through the 3.5k lines of tests, and I'm not sure I could tell you what the correct behaviour should actually be, but have you got a test for when the offset to sunset/sunrise takes it into the magic DST hour that doesn't exist/exists twice depending on which way the clock is going?

darinpope:
Assuming I understand your question, the tests you are looking for are in the "DST Boundary Jump Verification" test at line 1226.

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+					type: 'sunset',
+					latitude: 52.52,
+					longitude: 13.41, // Berlin
+					offset: 0,
+				}
+
+				// Day before DST (March 29, 2025) - CET (UTC+1)
+				vi.setSystemTime(new Date('2025-03-29T12:00:00Z'))
+				timer.setSun('pre-dst-berlin', params)
+				const preTime = timer.getSunNextExecuteTime('pre-dst-berlin')
```
peternewman, line 1283: According to https://www.timeanddate.com/sun/germany/berlin?month=3&year=2025 this should be 18:35

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+
+				// Day after DST (March 30, 2025) - CEST (UTC+2)
+				vi.setSystemTime(new Date('2025-03-30T12:00:00Z'))
+				timer.setSun('post-dst-berlin', params)
+				const postTime = timer.getSunNextExecuteTime('post-dst-berlin')
```
peternewman, line 1288: Likewise this should be 19:37 (as the CET to CEST offset is 1 hour).

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+
+				// Convert to local time strings for comparison
+				const preDate = new Date(preTime!)
+				const postDate = new Date(postTime!)
+				const preLocalHour = preDate.getHours()
+				const postLocalHour = postDate.getHours()
+
+				// Local time should be approximately the same
+				const hourDifference = Math.abs(postLocalHour - preLocalHour)
+				expect(hourDifference).toBeLessThan(3)
```
peternewman, line 1301: Shouldn't this therefore be checking its only changed by the DST offset?:
```suggestion
				expect(hourDifference).toBeLessThan(2)
```

Or are you also trying to cover the edge case place where the sunset was say 18:59 beforehand and hence would be 20:01 afterwards (~2 minutes of movement for the day, and then an hour of clock change)?

Wouldn't it be better to avoid that edge case by not picking test cases which sit near an hour boundary, or (and this perhaps brings it's own issues) doing maths on the hours and minutes combined, where the total move would be the clock change and ~5 minutes?

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
```
peternewman, line 1338: Again 18:45-17:43

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+			try {
+				const params = {
+					type: 'sunset',
+					latitude: -33.87,
+					longitude: 151.21, // Sydney
+					offset: 0,
+				}
+
+				// Day before DST ends (April 6, 2025) - AEDT (UTC+11)
+				vi.setSystemTime(new Date('2025-04-06T12:00:00Z'))
```
peternewman, line 1318: This has the clock change happening 5th into 6th April in Sydney:
https://www.timeanddate.com/sun/australia/sydney?month=4&year=2025

Assuming your offset is due to picking UTC rather than local, you ought to add a big here be dragons comment warning.

companion/test/Controls/Triggers/Events/Timer.test.ts
peternewman: Should you also add another test to confirm you've caught the right day too (again assuming it hasn't also tipped over an hour change during the test:
```suggestion
				expect(hourDifference).toBeLessThan(3)
				expect(hourDifference).toBeGreaterThan(1)
```

Your current test would be true on any (non-clock change) day of the year too.

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+					longitude: -74.01,
+					offset: -600, // -10 hours
+				}
+
+				timer.setSun('prev-day', params)
+				const time = timer.getSunNextExecuteTime('prev-day')
+
+				expect(time).not.toBeNull()
+				// May be in the past or future depending on calculation
+				expect(typeof time).toBe('number')
```
peternewman, line 2491: This doesn't feel like a very helpful test for confirming the code continues to behave the same...

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+
+				timer.setSun('extreme-plus', params)
+				const time = timer.getSunNextExecuteTime('extreme-plus')
+
+				expect(time).not.toBeNull()
+				expect(typeof time).toBe('number')
+
+				// Time should exist and be in reasonable range
+				const dateTime = new Date(time!)
+				expect(dateTime.getTime()).toBeGreaterThan(0)
```
peternewman, line 2737: Likewise here! Shouldn't it be the next day at least?

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+				}
+
+				timer.setSun('extreme-minus', params)
+				const time = timer.getSunNextExecuteTime('extreme-minus')
+
+				expect(time).not.toBeNull()
+				expect(typeof time).toBe('number')
+
+				const dateTime = new Date(time!)
+				expect(dateTime.getTime()).toBeGreaterThan(0)
```
peternewman, line 2762: And here

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+					type: 'sunset',
+					latitude: 40.71,
+					longitude: -74.01,
+					offset: 4320, // +72 hours
+				}
+
+				timer.setSun('large-plus', params)
+				const time = timer.getSunNextExecuteTime('large-plus')
+
+				expect(typeof time === 'number' || time === null).toBe(true)
```
peternewman, line 2783: This is only just testing it returns isn't it?

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+					type: 'sunrise',
+					latitude: 40.71,
+					longitude: -74.01,
+					offset: -4320, // -72 hours
+				}
+
+				timer.setSun('large-minus', params)
+				const time = timer.getSunNextExecuteTime('large-minus')
+
+				expect(typeof time === 'number' || time === null).toBe(true)
```
peternewman, line 2804: Likewise

peternewman:
> Assuming I understand your question, the tests you are looking for are in the "DST Boundary Jump Verification" test at line 1226.

I was meaning more "should handle offset that causes day wraparound during DST" but with an offset that takes it into the repeated/dropped hour depending on the DST direction.

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+					latitude: 40.71,
+					longitude: -74.01,
+					offset: 0,
+				}
+
+				timer.setSun('future-time', params)
+				const nextTime = timer.getSunNextExecuteTime('future-time')
+
+				expect(nextTime).not.toBeNull()
+				expect(nextTime).toBeGreaterThanOrEqual(now.getTime())
```
peternewman, line 2829: Shouldn't all the negative offsets include this test too?

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+				const laTime = timer.getSunNextExecuteTime('la')
+				const londonTime = timer.getSunNextExecuteTime('london')
+
+				expect(nyTime).not.toBeNull()
+				expect(laTime).not.toBeNull()
+				expect(londonTime).not.toBeNull()
+
+				// LA and London times should be different from NY (different locations)
+				expect(nyTime).not.toEqual(laTime)
+				expect(nyTime).not.toEqual(londonTime)
```
peternewman, line 3138: Without validating their approximate times, there's no way to prove that e.g. the long/lats haven't been shuffled between them in this test, making them all wrong, but still all different.

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+					latitude: 40.71,
+					longitude: -74.01,
+					offset: -1000, // ~16 hours earlier
+				}
+
+				timer.setSun('pre-dst-offset', params)
+				const time = timer.getSunNextExecuteTime('pre-dst-offset')
+
+				expect(time).not.toBeNull()
+				expect(typeof time).toBe('number')
```
peternewman, line 3342: Again minimal actual testing

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+					latitude: 40.71,
+					longitude: -74.01,
+					offset: -2880, // -2 days
+				}
+
+				timer.setSun('large-neg-offset', params)
+				const time = timer.getSunNextExecuteTime('large-neg-offset')
+
+				// May be in past or future depending on calculation
+				expect(typeof time === 'number' || time === null).toBe(true)
```
peternewman, line 3413: Has just returned!

Even past shouldn't be null!

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+					type: 'sunset',
+					latitude: 40.71,
+					longitude: -74.01,
+					offset: -3600, // -60 hours
+				}
+
+				timer.setSun('multi-day-past', params)
+				const time = timer.getSunNextExecuteTime('multi-day-past')
+
+				expect(typeof time === 'number' || time === null).toBe(true)
```
peternewman, line 3434: Weak test.

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+					latitude: 40.71,
+					longitude: -74.01,
+					offset: -120, // -2 hours exactly
+				}
+
+				timer.setSun('neg-hour-boundary', params)
+				const time = timer.getSunNextExecuteTime('neg-hour-boundary')
+
+				expect(time).not.toBeNull()
+				expect(typeof time).toBe('number')
```
peternewman, line 3456: Again vague test, could be much stronger.

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+			} finally {
+				vi.useRealTimers()
+			}
+		})
+	})
+
+	describe('Absolute Idempotency with State Verification', () => {
+		test('identical setSun calls on same ID should produce identical times', () => {
```
peternewman, line 3464: Is there an equivalent test where you check different IDs with the same lat/long/offset return the same value despite having different parameters?

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+				longitude: 180.0,
+				offset: 0,
+			}
+
+			timer.setSun('dateline-east-exact', params)
+			const time = timer.getSunNextExecuteTime('dateline-east-exact')
+
+			expect(time).not.toBeNull()
+			expect(typeof time).toBe('number')
+			expect(time).toBeGreaterThan(0)
```
peternewman, line 3575: Should be in the future surely?

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+			// Both should be the same type (both number or both null)
+			expect(typeof time1).toBe(typeof time2)
+			if (typeof time1 === 'number' && typeof time2 === 'number') {
+				expect(time1).toBe(time2)
+			}
+		})
+	})
+
+	describe('Maximum Valid Coordinate Precision', () => {
+		test('latitude +90 (north pole) should calculate or return NaN', () => {
```
peternewman, line 3534: Which one is it, calculate or return NaN?

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+				offset: 0,
+			}
+
+			timer.setSun('north-pole-exact', params)
+			const time = timer.getSunNextExecuteTime('north-pole-exact')
+
+			expect(typeof time === 'number' || time === null).toBe(true)
+		})
+
+		test('latitude -90 (south pole) should calculate or return NaN', () => {
```
peternewman, line 3548: Again

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+				longitude: -180.0,
+				offset: 0,
+			}
+
+			timer.setSun('dateline-west-exact', params)
+			const time = timer.getSunNextExecuteTime('dateline-west-exact')
+
+			expect(time).not.toBeNull()
+			expect(typeof time).toBe('number')
+			expect(time).toBeGreaterThan(0)
```
peternewman, line 3591: Future again

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+					type: 'sunset',
+					latitude: corner.lat,
+					longitude: corner.long,
+					offset: 0,
+				}
+
+				timer.setSun(`corner-${corner.name}`, params)
+				const time = timer.getSunNextExecuteTime(`corner-${corner.name}`)
+
+				expect(typeof time === 'number' || time === null).toBe(true)
```
peternewman, line 3613: Vague result

companion/test/Controls/Triggers/Events/Timer.test.ts
```diff
@@ -0,0 +1,3633 @@
...
+				longitude: -74.0159876543,
+				offset: 0,
+			}
+
+			timer.setSun('high-precision', params)
+			const time = timer.getSunNextExecuteTime('high-precision')
+
+			expect(time).not.toBeNull()
+			expect(typeof time).toBe('number')
+			expect(time).toBeGreaterThan(0)
```
peternewman, line 3630: Future again

peternewman (changes requested):
I'm going to assume the tests were the main part written by AI?

I'd argue a lot of those test cases should be a lot stricter, even if it's just to confirm that the edge case behaviour remains as is...

I've also only reviewed ones with an offset set (~170), not all the others that exist.

Julusian:
@darinpope There are a bunch of comments here that I think need resolving before this can progress.

Julusian:
The target file has been rewritten a bit since this was opened, as well as some additional work on timezone handling. So I have reworked this as 727440389e174a3993887d903c7a856b6bddc78c

This will be in v5.0.2. Hopefully it works, I don't know how to go about testing it

darinpope:
Thanks. I apologize for ghosting this issue. I'll set a reminder now to test when time changes again, assuming Daylight Saving Time legislation doesn't pass in the US :)

PR #3758, opened 2025-11-07
Fix "Sentry DSN not located" error in launcher

arikorn:
- To see the error use `yarn dev` or run Companion.exe (if Windows) from a cmd prompt or powershell.
- Note that dev builds will still say "Sentry error reporting is disabled". This is not an error, it is because the file: SENTRY is empty.

### DIFF
--- a/launcher/main.js
+++ b/launcher/main.js
@@ -159,7 +159,6 @@ if (!lock) {
 
 	let sentryDsn
 	try {
-		//console.log('Looking for sentry at: ' + new URL('../SENTRY', import.meta.url))
 		sentryDsn = fs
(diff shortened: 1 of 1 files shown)


arikorn:
Github complained in this PR about the format in peter newman's PR that you recently merged. Fixing it here just because it's a big nothingburger... (Except that now you have to merge this one! 😆 )

launcher/main.js
```diff
@@ -159,8 +159,9 @@ if (!lock) {
 
 	let sentryDsn
 	try {
+		//console.log('Looking for sentry at: ' + new URL('../SENTRY', import.meta.url))
```
Julusian:
```suggestion
```

Issue #3655, opened 2025-09-26
[BUG] Template literal parsing issues in some expressions


thedist:
### Make sure you're on the latest stable or beta build

- [x] I have tested this on the latest stable or beta release

### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion and not a specific module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

I've noticed that template literals are handled differently in button text compared to how they are handled in actions, such as `internal: Custom Variable: Set with expression`.

In button text typing a variable without treating it like an expression to be evaluated (ie, wrapping it in `${}`) will write out the actual characters, for example if you wrote \`$(custom:variable) dB\` you would get `$(custom:variable) dB` on a button, as expected. and if you did \`${$(custom:variable)} dB\` you would get `123 dB` or whatever the value is.

When using `internal: Custom Variable: Set with expression` though \`$(custom:variable) dB\` results in 2 `info Variables/Util Reached iteration limit for variable parsing ` logs, and the variable value becomes `$(custom:variable) dB dB dB dB dB dB dB dB dB...` with ` dB` repeating. Using  \`${$(custom:variable)} dB\` works as intended though.

### Steps To Reproduce

1. create a `internal: Custom Variable: Set with expression` action on a button
2. create a custom variable, such as test1 with a default value like 123
3. set the action to target test1
4. set the action to use the expression  \`$(custom:test1) dB\` 
5. test1 will be set to `$(custom:test1) dB dB dB dB dB` ...

### Expected Behavior

The expected behaviour is for Expressions in actions to match that of button text set to Expression mode, which is to not attempt to parse any Companion variable within a template string unless it is within dollar sign braces `${}`.

I'd also not expect that when there is a variable present but not being parsed it doesn't loop and create the wrong result such as repeating characters.

### Environment (please complete the following information)

```markdown
- OS: Win 11
- Browser: Chrome 140
- Companion Version: v4.2.0+8454
```

Issue #2577, opened 2023-09-15
[BUG] Custom variables, current value empty after powershell command


GeoDerp:
### Is this a bug in companion itself or a module?

- [X] I believe this to be a bug in companion

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Wanted to grab the current composition title from Resolume, so I made a powershell command to access the rest api and give me a string output:
```powershell
powershell Write-Host -NoNewline $(Invoke-RestMethod -Uri "10.0.0.15:8080/api/v1/composition" ).name.value
```
terminal output:
```cmd
>powershell Write-Host -NoNewline $(Invoke-RestMethod -Uri "10.0.0.15:8080/api/v1/composition" ).name.value
test
>
```
I then created a internal: System: Run shell path (local) button and set the target variable.
This seems to be modifying the custom variables output. However, instead of placing the string output it places null/empty.
Got any thoughts?

### Steps To Reproduce

- ran Resolume, saved composition as an name (*ResolumeCompositionName*)
- created [button ](internal: System: Run shell path (local))
   - set path:  powershell Write-Host -NoNewline $(Invoke-RestMethod -Uri "10.0.0.15:8080/api/v1/composition" ).name.value
   - set Target Variable (stdout): *theCustomVariable*
- pressed button 
- view *theCustomVariable*, Current value is: null

### Expected Behavior

- view *theCustomVariable*, Current value is: *ResolumeCompositionName*

### Environment (please complete the following information)

```markdown
- OS: windows 
- Browser: chrome
- Companion Version: 3.0.1
```

### Additional context

No warning, info or debug on log when button is pressed, when command ran in cmd, output is correct.

Julusian:
@GeoDerp the generic-http module is able to make http calls and store the result in a custom variable, so can probably handle this for you

Julusian:
oh wait, I know what is happening.. I always knew this day would come, but I wasn't expecting it from a powershell script..

That action is parsing the companion variables it finds in the command you provide, resulting in it executing `powershell Write-Host -NoNewline $NA.name.value`.

This is a flaw I have been wondering about for a while in how we parse variables, as there is not a way to escape the sequence. I am not sure how best to fix this right now, adding an escape character could break other situations so it needs to be approached carefully.

PR #3768, opened 2025-11-17
Fix handling of variables in Template Literals in button text expressions

arikorn:
This PR simplifies handling of Companion variables in button text-expressions so the user can write `` `Prefix_$(custom:var)` `` rather than `` `Prefix_${$(custom:var)}` ``, which is currently required.

Details: Currently button text expressions have a curious behavior:

If you put a companion variable in a template literal, the variable is not evaluated so the button shows the name of the variable instead of its value, so to evaluate it you'd have to write: `` `${$(my:var)}` ``, which is rather awkward.

OTOH, if you set a variable, say `$(custom:myvar)`, to the original template literal, i.e. `` `$(my:var)` ``, and then use the new variable as the button text value (whether expression or text), the button now shows the _value_ of `$(my:var)` rather than the string "$(my:var)" -- even though the string is shown as the value of `$(custom:myvar)` in the Variables tab.

This PR fixes the immediate problem for button text expressions and also suggests an alternate option, which is to resolve variables in backtick expressions at parse-eval time, just as is already done for rest of the template literal. (i.e. if  `$(custom:myvar)` is set to the expression `` `$(internal:time_s) ${unixNow()}` `` , the value shown on the button will update every second for the Companion Variable but not the function call.)

And tangentially: I realize now that I've blurred the distinction between "expression variables" and "custom variables" by using an expression in a custom variable! (But that's not really relevant to the button-text problem, just an aside...)

Edit:
Just to add a little confusion: if the button expression is `concat($(internal:time_s), '\n', unixNow())` then both values update once a second, as does `` `${$(internal:time_s)} ${unixNow()}` `` -- or `` `$(internal:time_s) ${unixNow()}` `` in this PR.

And finally, this PR does not break the current syntax, it just makes it unnecessary.

### DIFF
--- a/companion/lib/Controls/ControlTypes/Button/Util.ts
+++ b/companion/lib/Controls/ControlTypes/Button/Util.ts
@@ -36,7 +36,17 @@ export function parseVariablesInButtonStyle(
 				logger.error(`Expression parse error: ${parseResult.error}`)
 				style.text = 'ERR'
 			}
-			return parseResult.variableIds.size > 0 ? parseResult.variableIds : null
+			if (parseResult.variableIds.size > 0) {
+				// some variables were processed, assume they all were
+				return parseResult.variableIds
+			} else if (style.text.includes('$(')) {
+				// reparse for variables, since the jsep output doesn't resolve Companion variables
+				const parseResult = parser.parseVariables(style.text)
+				style.text = parseResult.text
+				return parseResult.variableIds.size > 0 ? parseResult.variableIds : null
+			} else {
+				return null
+			}
 		} else {
 			const parseResult = parser.parseVariables(style.text)
 			style.text = parseResult.text
--- a/shared-lib/lib/Expression/ExpressionResolve.ts
+++ b/shared-lib/lib/Expression/ExpressionResolve.ts
@@ -161,6 +161,16 @@ export function ResolveExpression(
 							}
 						}
 
+						// // resolve embedded Companion variables
+						// let startIdx;
+						// while ((startIdx = result.search(/\$\(/)) >= 0) {
+						// 	const endIdx = result.indexOf(')', startIdx + 2)
+						// 	if (endIdx < 0 ) break  // or throw error?
+						// 	const varname = result.substring(startIdx+2, endIdx)
+						// 	const value = getVariableValue(varname)?.toString() ?? ''
+						// 	result = result.replaceAll(`$(${varname})`, value)
+						// }
+
 						return result
 					}
 					case 'Compound': {


Julusian:
I am not sure if I want to go with this.

There is an alternate idea (I dont remember where), which is to allow referencing variables as something like `vars.custom.thing` inside expressions.  
I am quite tempted by this, but I feel like I have been putting it off for some reason..

Both of these pitches would have a similar effect, but with slightly different syntax `abc ${vars.custom.thing}` vs `abc $(custom:thing)`, but the global object approach would allow for cleaning up other things a bit, such as `a = a + $(custom:thing)` could become `a = a + vars.custom.thing`

thedist:
Isn't this a feature request rather than a fix? It appears the PR is trying to implement new functionality, rather than fixing broken functionality?


> And finally, this PR does not break the current syntax, it just makes it unnecessary.

But if `$(custom:var)` is equal to 1 for example, wouldn't `Prefix_$(custom:var)` become `Prefix_1` in with your PR? where as currently it would be `Prefix_$(custom:var)`? So while it may not be breaking for anyone doing `Prefix_${$(custom:var)}`, it is breaking for anyone who intentionally wants to use the variable text somewhere and use template literals to prevent it being parsed.

> There is an alternate idea (I dont remember where), which is to allow referencing variables as something like vars.custom.thing inside expressions.

This seems more of a reasonable idea, as anything inside `${}` within a template literal should be parsed, so if we can simplify the syntax within that then that has potential, but going from not parsing stuff outside of that to suddenly parsing things which would previously not be parsed is not good UX, and even for users who are familiar with JS syntax even they wouldn't be expecting things outside of `${}` to be manipulated or messed with by Companion either.

arikorn:
@Julusian:
Funny, I thought that `$(module:varname)` syntax was sacrosanct, considering discussions in one of my recent issues 😁 .

But seriously, unless you're proposing removing the current variable syntax altogether, it seems to me that it needn't be an either-or proposition. 

BTW, I don't think either solution I show in this PR is quite right, but that's just a matter of detail that can be easily fixed if we want to go forward. 

If there's no need for lazy evaluation of variables in template literals, I would definitely go with the commented-out version in this PR, since that's closer to the source. FWIW, my first inclination was to fork-and-fix the jsep template-literal plugin to recognize companion variables. Your solution of just changing the syntax does the same thing by changing companion instead of jsep! ...and implies that the lazy evaluation was not an intended feature.

As for personal preference on syntax, I slightly prefer the current syntax since it stands out better and is more compact. But I definitely see the value in being more Javascript-like, if that's the direction Companion is heading (i.e. have expressions use more standard JS syntax such as `mystring.replace('x', 'y')`...)

@thedist wrote:
> it is breaking for anyone who intentionally wants to use the variable text somewhere and use template literals to prevent it being parsed.

1. It's fine to call it a feature if you prefer. The main reason I called it a fix, I think, is because it corrects the inconsistent behavior described above.

2. If they want to prevent it from being parsed why would they put it in a template literal? Regular literals already exist to do that. So, yes, I suppose it's breaking if people use inappropriate operations to do things nobody really would want to do anyway 😀. This is certainly not the first time a minor code change "broke" rare edge cases.  [Edit: and to clarify my original statement, what I meant was: it doesn't break `` `${$(my.var)}` `` syntax, just makes it unnecesary. Breaking _that_ would be considered a breaking change by any measure. ]

3. If I understand your third point, it is a "vote" for the way I solve it in this PR -- i.e. it is restricted to button-text processing rather than globally affecting expressions. That said, I think you would have to provide examples of where "lazy" vs. "immediate" evaluation after parsing would make a difference.

Julusian:
>  Funny, I thought that $(module:varname) syntax was sacrosanct, considering discussions in one of my recent issues 😁 .

It is, but that doesn't mean we can't add another syntax for accessing variables from inside expressions.

> As for personal preference on syntax, I slightly prefer the current syntax since it stands out better and is more compact. 

That depends on what you are doing.  
`$(custom:abc)` is more compact than `vars.custom.abc`  
but not  
`getVariable(`$(custom:${a}_1)`)` or `getVariable('custom', `${a}_1)`)`  vs `vars.custom[ `${a}_1)`]`

> But I definitely see the value in being more Javascript-like, if that's the direction Companion is heading (i.e. have expressions use more standard JS syntax such as mystring.replace('x', 'y')...)

tbh, this is not something on my radar. For some reason i decided to make the functions quite functional and not very js, and I dont think adding methods like `mystring.replace('x', 'y')` would be helpful at this point.  
Something to consider if we did do this; what should happen if `mystring` is not a string? the current methods are intended to be very safe and will convert or fail gracefully. That will be possible to do with member methods, but potentially harder especially if we have any naming collisions.

Both of these things are adding multiple ways of doing things, so I suppose could argue if doing one then do both, but I think there is differing levels of value.  

> If they want to prevent it from being parsed why would they put it in a template literal? Regular literals already exist to do that. So, yes, I suppose it's breaking if people use inappropriate operations to do things nobody really would want to do anyway 😀. 

A pattern I believe is not uncommon is to put something with the name of a variable inside a custom variable, which can then be fully computed in button text.  
A silly example is setting a custom variable with the expression:
```
a = 'hms'
`Time: $(internal:time_${a})`
```
Currently, that will set the custom variable to `Time: $(internal:time_hms)`, meaning that when putting that in button text the time will update. With your change, the current `$(internal:time_hms)` value will be baked in.

This is a silly example, a more realistic one could be the same style but `Input ${a}: $(module:input_${a}_name)`

Maybe this is the wrong way now that expression variables are a thing, but many configs will predate that.

----

As a tangent, but related thing; We already have an issue with the parsing being over-eager. powershell uses a similar $( identifier for some things, causing #2577 where it isnt possible to pass those into the `Run shell path` action.  

As I write this, I am realising that once #3617 is merged, there might be a workaround/solution of using expression mode, as the output of that wont be parsed for variables. This PR would keep the expression mode as causing this bug.

arikorn:
@Julusian 

> Currently, that will set the custom variable to Time: `$(internal:time_hms)`, meaning that when putting that in button text the time will update. With your change, the current `$(internal:time_hms)` value will be baked in.

The clock ticks if you use the code in this PR -- I just tried it. The same would be, no doubt, true for the more realistic example. The code just reparses the jsep result, it shouldn't change other behaviors. (There is a bug in my code that will fail the parse the backtick portion if there's a variable outside the template literal and inside it, but that's what I meant above, that it's easily fixed if we go this way.)

One place where this does fail more like Jeff's (@thedist) original complaint, is that it will reparse string literals too. so `'Time: $(internal:time_hms)'` will show the time instead of the variable. This argues in favor of the commented-out solution in this PR, which differentiates between string literals and template literals

But the ultimate question is whether we want to parse $(xx:yy) in string literals or introduce a new syntax. Other than syntax, I see no difference in behavior between your proposal and the commented out code here (sans bugs in my code).

Julusian:
Oh oops, I didn't realise what file this was being done in :facepalm:   
I saw the change and assumed it was changing how all expressions were parsed, not specifically for button text.

This reminds me of https://github.com/bitfocus/companion/issues/3655, which may have no impact on this at all but is another difference in behaviour we had for the button text modes.

-----

so, ignore all I was saying about the alternate proposal, that isnt really relevant. probably most of what I have written. 

After more carefully re-reading everything you have written, my first thought is whether users will get confused because this will behave differently depending on where they are writing the expression.  
An expression `` `test: $(custom:abc)` `` that works nicely in button text, won't work the same if they put it in the 'internal: Custom Variable: Set with expression' action.

I can't speak for anyone else, but I often use button text as a way to debug expressions, as I can write it there and get a live preview of the value.

I am not entirely against it; we do some not entirely dissimilar recursive parsing of variables in some other places. But I am not immediately convinced at this one place have a slightly different expression behaviour.

I'm also not keen on the `// some variables were processed, assume they all were` branch, I worry that confuse users.  
An expression could work, then you change a line above the final result and the variables stop parsing. I expect there will be at least one bug report from that.
(Disclaimer: I havent run it to verify this, but I am 95% sure this is how it behaves from reading)

eg:
```
a = 'test'
`${a} = $(custom:b)`
```
so far my expression seems good, and gives `test = 1`

then that changes to:
```
a = $(custom:a)
`${a} = $(custom:b)`
```
thats weird, now it broke and gives `test = $(custom:b)`


> OTOH, if you set a variable, say $(custom:myvar), to the original template literal, i.e. `$(my:var)`, and then use the new variable as the button text value (whether expression or text), the button now shows the value of $(my:var) rather than the string "$(my:var)" -- even though the string is shown as the value of $(custom:myvar) in the Variables tab.

Right ok, it took me a little bit to figure out what this was saying  
So in the context of that, this pr does make more sense.  
I'm not sure that means this should be done, or as that is 'legacy' behaviour it shouldnt.. I might get back to you on this

> This PR fixes the immediate problem for button text expressions and also suggests an alternate option, which is to resolve variables in backtick expressions at parse-eval time, just as is already done for rest of the template literal. (i.e. if $(custom:myvar) is set to the expression `$(internal:time_s) ${unixNow()}` , the value shown on the button will update every second for the Companion Variable but not the function call.)

I suppose I am not opposed to this idea, but it needs a bit of refinement.  
Elsewhere we can handle `abc $(custom:$(custom:b))`, which this probably should try to too? That might be as simple as using the `functionsRaw.parseVariables()` when it exists (and when it doesn't, skip this parsing?)  
This should probably also support doing a `\$(asd` for when you want the unparsed literal. How to do that without affecting parsing elsewhere will be 'fun'  
And ideally, the expression editor should be able to syntax highlight these.


That last point is perhaps another factor in why this has come up, the editor is incorrectly suggesting variables while writing a string literal or template string. It wasnt supposed to be doing that.

dnmeid:
I'd say this isn't a "fix" because the current syntax is not broken, it is like intended. I also wouldn't call it curious or awkward.
I don't see at all why we should change the way how template strings are evaluated. If you want to use an expression inside a template, you have to surround it with ${}. The expression syntax itself should be the same everywhere.
We had a different discussion somwhere if if would be a good idea to drop or change the $() for variables in expressions. Then you could write `` `Prefix_${ custom:var }` ``, but of course in every expression, not only button text. I think that discussion is still around somwhere and this PR should be rejected.

arikorn:
@Julusian wrote:
>This reminds me of https://github.com/bitfocus/companion/issues/3655, which may have no impact on this at all but is another difference in behaviour we had for the button text modes.

Ah yes, similar but different. It also helps me understand Jeff's (@thedist) POV.

We start with the same goal: to make Companion behave consistently, but whereas Jeff is comparing button text expressions to action expressions, I'm comparing two variations on button text expression. To unpack my OP, consider two variables:

~~~
$(custom:myVar1) = 123
$(custom:myVar2) = `$(custom:myVar1)` // (probably need to set this with an action)
~~~

With these you can set button text two ways:

1. Text expression: `` `$(custom:myVar1)` ``. Button shows "$(custom:myVar1)"
2. Text expression: `$(custom:myVar2)`. Button shows "123", even though the Variables tab confirms that the value is `$(custom:myVar1)`

This PR considers the second case more correct and fixes the first case to match. Jeff appears to value the first case, in which case the second case would need to be fixed.

>An expression `` `test: $(custom:abc)` `` that works nicely in button text, won't work the same if they put it in the 'internal: Custom Variable: Set with expression' action.

We can lay this all aside, because in the end between Jeff's objection and your "more realistic case" I've realized that fixing it at the button level will probably never work correctly because it can't differentiate template literals from regular literals. I.e. it will substitute the value even when the user wrote `"$(custom:abc)"` and that's almost definitely wrong. So the only choice here is to do it when resolving parser output, i.e. my commented-out code, or your alternative syntax (`var.custom.abc`) -- or potentially, either of them

>I suppose I am not opposed to this idea, but it needs a bit of refinement.
>Elsewhere we can handle abc $(custom:$(custom:b)), which this probably should try to too? That might be as simple as using the functionsRaw.parseVariables() when it exists (and when it doesn't, skip this parsing?)
>This should probably also support doing a \$(asd for when you want the unparsed literal. How to do that without affecting parsing elsewhere will be 'fun'
>And ideally, the expression editor should be able to syntax highlight these.

Totally agree. FWIW, `CompanionVariablesPlugin` (_CompanionVariables.ts_) fails most of these cases too, but only if the variable is not inside a template literal, since template processing overrides variable processing.

Julusian:
It looks like there is an agreement that doing something special for button text is not what we want.

Having slept on the parser approach, I am less sold.  
I am worrying that it will lead to more issues like #2577, where we are overeager in parsing variables, and there is no way to opt out.

I have another thought; the recursive parsing of variables like we have today:
```
$(custom:myVar1) = 123
$(custom:myVar2) = `$(custom:myVar1)` // (probably need to set this with an action)
```
where using `$(custom:myVar2)` gives `123`, is perhaps a relic of a pre expression and especially a pre expression-variable companion. Every use case I can think of for it can be solved with newer features.  



So rather than expanding upon that, maybe we should start to phase that out.  
This could be done in a few steps:
1) Disable this behaviour for all module variables (or is this something that modules will be using?)
1) Make this behaviour opt-in for all custom-variables (enabled for any existing custom-variables/old imports)
1) Write a reminder for ourselves to remove it whenever we get to 5.0

If anyone is relying on this functionality, they can do it explicitly in an expression `parseVariables($(custom:myVar2))` would give `123`, so if we do break anyones config, we can give guidance on how to fix it.  
It is wordier/more effort for users, but it avoids quirks in the recursive parsing and removes some unavoidable magic

edit: I keep glossing over that this is intended to work specifically for template strings, assuming it operates on all strings. I can see why it would make sense to support this syntax, but I'm not sure its worth the complexity and diverging from js syntax for this.  
And it is worth noting that the existing behaviour you refer to of chaining variables is not specific to template strings, but to any string. So if template strings in expressions do this interpolation, why not others?


---

As a side thought; perhaps as part of this, once we support expressions in action/feedback input fields, we should start phasing out any form of the old variable interpolation on these. keep that as something specific for button text and the function in expressions?  
Idk, a random thought that it is way way too early to act upon, and I am not sure about, so not something to debate here

arikorn:
Closing as per @dnmeid's suggestion

PR #3774, opened 2025-11-21
Fix Install/Build Issues for Webui on Windows

arikorn:
- Make the post-install script platform independent

Note - my Windows used an old Cygwin version to run the shell commands but this added "special permissions" that blocked copying the files, which caused `yarn dist:winui` to fail. This commit protects users from quirks of the OS and also reduces the external dependencies one would need to start development on Windows.

[fixed] ~~Note: I would have preferred to put _post-install.ts_ in webui/ rather than webui/src, but lint kept on complaining...~~ 

P.S. I have now successfully built companion on Windows (i.e. `yarn dist`) without needing to manually fix anything -- though it does need elevated permissions for Windows to make symbolic links (need when unzipping the signtool).

### DIFF
--- a/webui/post-install.ts
+++ b/webui/post-install.ts
@@ -2,8 +2,6 @@
 
 import { fs } from 'zx'
 
-//  rm -Rf public/_deps; mkdir public/_deps && cp -R ../node_modules/monaco-editor/min/vs public/_deps/monaco
-
 // Trash an old folder and/or create a new one
 fs.emptyDirSync('public/_deps')
 


webui/post-install.ts
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env zx
+
+import { fs } from 'zx'
+
+//  rm -Rf public/_deps; mkdir public/_deps && cp -R ../node_modules/monaco-editor/min/vs public/_deps/monaco
```
Julusian:
```suggestion
```

PR #3779, opened 2025-11-23
feat: Support major firmware update of mirabox 293S variant

njibhu:
This is a follow-up PR of https://github.com/bitfocus/companion/pull/3632
Soomfon is now shipping their XF-CN001 with a new firmware (V3), which fixes pretty much all the issues that limited the previous version:
- Supports up/down events
- Tile refreshing is much faster
- Screen tiles can be configured to be larger to better fit the buttons.

Sadly, there's no firmware update publicly available to patch older units, probably because it would break whatever config was created with the previous firmware version. So the other implementation is still very useful for older units.

### DIFF
--- a/companion/lib/Surface/Controller.ts
+++ b/companion/lib/Surface/Controller.ts
@@ -1213,18 +1213,16 @@ export class SurfaceController extends EventEmitter<SurfaceControllerEvents> {
 											await this.#addDevice(deviceInfo.path, {}, '203-mystrix', SurfaceUSB203SystemsMystrix)
 										}
 									} else if (
-										(deviceInfo.vendorId === 0x1500 ||
-											deviceInfo.vendorId === 0x6602 ||
-											deviceInfo.vendorId === 0x6603 ||
-											deviceInfo.vendorId === 0x5548) && // Mirabox
-										(deviceInfo.productId === 0x1001 ||
-											deviceInfo.productId === 0x1003 ||
-											deviceInfo.productId === 0x1007 ||
-											deviceInfo.productId === 0x1005 ||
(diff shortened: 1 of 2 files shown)


companion/lib/Surface/Controller.ts
```diff
@@ -1213,15 +1213,17 @@ export class SurfaceController extends EventEmitter<SurfaceControllerEvents> {
 											await this.#addDevice(deviceInfo.path, {}, '203-mystrix', SurfaceUSB203SystemsMystrix)
 										}
 									} else if (
-										(deviceInfo.vendorId === 0x6602 ||
+										(deviceInfo.vendorId === 0x1500 ||
```
Julusian: please do these as some matched vendorId and productIds.

There is a rework in progress that will require these to be paired, so it would be best to start doing this now to minimise the guesswork and breaking when that rework is done

companion/lib/Surface/USB/MiraboxStreamDock.ts
```diff
@@ -2004,6 +2302,8 @@ class StreamDock extends EventEmitter {
 			this.model = StreamDock.models['293N3']
 		} else if (this.info.productId === 0x1014) {
 			this.model = StreamDock.models['HSV 293S']
+		} else if (this.info.productId === 0x3003) {
```
Julusian: same as elsewhere, these should be checking the vendorId.

you dont need to update them all, and can do just the ones you know.  
having any of these paired fully will reduce the guesswork I do in the refactor later on

njibhu:
@Julusian Makes sense!
Does it look better now ?

PR #3790, opened 2025-11-27
feat: support Loupedeck touch fader enhancements

Beastlybear2017:
This PR expands on the current vertical display fader feature by:
-  Adding the functionality to the Loupedeck Live.
-  Adding a toggle to invert the fader. (This affects both the visual display as well as the value output)
-  Updating the draw logic to use the button background colour. 

I know that this may become redundant once the graphics overhaul comes into full effect with bitfocus/companion-surface-loupedeck#15, however I would love to see this developed into customisable faders so that we could mix and match smaller buttons and faders on the same display.

### DIFF
--- a/companion/lib/Surface/USB/Loupedeck.ts
+++ b/companion/lib/Surface/USB/Loupedeck.ts
@@ -51,8 +51,8 @@ export class SurfaceUSBLoupedeck extends EventEmitter<SurfacePanelEvents> implem
 
 	config: Record<string, any>
 	DisplayColors = {
-		LeftColor: {red: 0, green: 100, blue: 0},
-		RightColor: {red: 0, green: 0, blue: 50},
+		LeftColor: { red: 0, green: 100, blue: 0 },
+		RightColor: { red: 0, green: 0, blue: 50 },
 		LeftValue: 0,
 		RightValue: 0,
 	}
@@ -151,7 +151,7 @@ export class SurfaceUSBLoupedeck extends EventEmitter<SurfacePanelEvents> implem
 					default: false,
 					label: 'Invert Fader Values',
 					tooltip: 'If set, the fader values will be inverted, with the value being between 256 and 0.',
-				}
+				},
 			]
 
 			/**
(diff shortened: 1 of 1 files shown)


companion/lib/Surface/USB/Loupedeck.ts
```diff
@@ -312,8 +356,34 @@ export class SurfaceUSBLoupedeck extends EventEmitter<SurfacePanelEvents> implem
 					})
 			} else if (control.feedbackType === 'lcd') {
 				this.#writeQueue.queue(control.id, render)
+			} 
+		} else if (control.type === 'lcd-segment') { // Update the slider display on render call to change color
+				const styleBg =
+					typeof render.style === 'object' && render.style !== null && 'bgcolor' in render.style
+						? (render.style as any).bgcolor
+						: undefined
```
Julusian: This can be simplified and kept type safe:
```suggestion
				const styleBg = render.style?.bgcolor
```

companion/lib/Surface/USB/Loupedeck.ts
```diff
@@ -50,6 +50,7 @@ export class SurfaceUSBLoupedeck extends EventEmitter<SurfacePanelEvents> implem
 	readonly #writeQueue: ImageWriteQueue<string, [ImageResult]>
 
 	config: Record<string, any>
+	DisplayColors: Record<string, any>
```
Julusian: It would be good to have stronger types for this object, to minimise the risk of type errors in future refactorings

companion/lib/Surface/USB/Loupedeck.ts
```diff
@@ -312,8 +356,34 @@ export class SurfaceUSBLoupedeck extends EventEmitter<SurfacePanelEvents> implem
...
+				const styleBg =
+					typeof render.style === 'object' && render.style !== null && 'bgcolor' in render.style
+						? (render.style as any).bgcolor
+						: undefined
+
+				if (control.id == "left") {
+					this.DisplayColors.LeftColor = (({ r, g, b }) => ({ red: r, green: g, blue: b }))(colorToRgb(styleBg));
+				} else if (control.id == "right") {
+					this.DisplayColors.RightColor = (({ r, g, b }) => ({ red: r, green: g, blue: b }))(colorToRgb(styleBg));
+				}
```
Julusian: I think this is quite hard to understand what is going on here.
I find this to be simpler and more readable:

```suggestion
				if (control.id == "left") {
					const col = colorToRgb(styleBg)
					this.DisplayColors.LeftColor = { red: col.r, green: col.g, blue: col.b }
				} else if (control.id == "right") {
					const col = colorToRgb(styleBg)
					this.DisplayColors.RightColor = { red: col.r, green: col.g, blue: col.b }
				}
```

Julusian (commented):
Thanks for this!

I have a few small requests for tweaks to make to this before I merge it.

Also could you run `yarn format` as there are some formatting errors in this.

companion/lib/Surface/USB/Loupedeck.ts
Julusian: I was wrong, it wasnt that simple..

elliotmatson:
Is it possible to add these features to the Razer Stream controllers as well?

PR #3791, opened 2025-11-28
Feature: Two-line entries for variable-name dropdowns

arikorn:
This PR makes the variable-name dropdown menus behave in actions and feedbacks similarly to the way they work in the expression editor. (_**Edit: I updated the formatting below -- see the second comment, also edited. Although this first approach may have advantages in simplicity the other is more consistent with the existing UX**_...)

Each entry in the dropdown is two lines:
~~~
$(module:varname)
Variable description text
~~~

image

<BR/>
<BR/>

NB
----
The main effect of the CSS change in this PR is to allow `\n` to break lines in menu-items. In addition, it removes some "excessive" spacing between menu items.

Wrapping behavior is not changed: Prior to this PR long values were wrapped in menu items but not in the current value (i.e. when dropdown is not "open). This PR does not change the behavior, but it would be very easy to wrap text even in the current value -- in fact I had to do extra work to prevent it!

Conversely, if preferred: with a little effort the CSS changes in this PR can be isolated to affect only variable dropdowns.

Current Behavior before this PR:
----
- In Action, Dropdown "closed"
image

----

- In Action Dropdown, open

image

----
Unchanged:
| In Button Text  (v4.1, v4.2 text-mode)| v4.2 expression-mode|
|---|---|
|image |  image |

### DIFF
--- a/webui/src/Components/ExpressionInputField.tsx
+++ b/webui/src/Components/ExpressionInputField.tsx
@@ -119,9 +119,9 @@ export const ExpressionInputField = observer(function ExpressionInputField({
 					folding: false,
 					fixedOverflowWidgets: true,
 					fontSize: 15,
+					// set suggestFontSize to 16 to make menu font and row height more like the other variable suggestion menus.
+					suggestFontSize: 14,
 					// Make suggest widget rows tall enough for two lines (name + description)
-					// and make menu font and row height more like the other variable suggestion menus.
-					suggestFontSize: 16,
 					suggestLineHeight: 53,
 				}}
 			/>
--- a/webui/src/scss/_expression.scss
+++ b/webui/src/scss/_expression.scss
@@ -17,7 +17,6 @@
 
 	/* Override Monaco suggest widget to show details below name */
 	.monaco-editor .suggest-widget {
-		width: 25em !important;
 		z-index: 10000 !important; /* otherwise the top goes under the top banner in action/feedback expression suggestions */
 
 		.monaco-list .monaco-list-row {


arikorn:
The final commit, above, formats "options" the same as in the button-text suggestions and also formats the "singe-value" similarly, but there may be better/other ways to do the same...  

Incidentally, there's a difference in what the single-value box shows when clicked between  actions (set custom variable -- shows same text either way) and feedbacks (variable check value -- shows only the value when selected or has focus as shown below) -- I think because the latter allows the user to type in the value.

image

<BR/>
<BR/>
Open:
<BR/>
image

webui/src/scss/_controls.scss
```diff
@@ -152,6 +167,19 @@ label.disabled {
 	}
 }
 
+/* Allow \n line-breaks and remove "excess" line spacing between pulldown options. */
+.select-control__option {
+	white-space: pre-wrap !important;
+	line-height: 1.1;
```
Julusian: These feel a bit squished together to me, but that could just be because I am used to how the variables one looks

webui/src/Components/DropdownInputField.tsx
```diff
@@ -23,6 +23,9 @@ interface DropdownInputFieldProps {
 	onBlur?: () => void
 	onPasteIntercept?: (value: string) => string
 	checkValid?: (value: DropdownChoiceId) => boolean
+	// The following is a compromise since if we just use typeof components,
+	// Typescript objects to `OptionProps<DropdownChoiceInt>` in the custom property functions in InternalModuleField.tsx
+	customComponents?: Partial<Record<keyof typeof components, any>>
```
Julusian: I think it would be better to instead of this, have a simple `showOptionsOver2Lines` (or some better named) property, and leave the custom components to be handled inside of this file.

I am wondering if it should be enabled for all/most dropdowns.. Perhaps only useful after #3617, so dont worry about doing this for now.

webui/src/Controls/InternalModuleField.tsx
```diff
@@ -225,6 +227,28 @@ export const InternalPageIdDropdown = observer(function InternalPageDropdown({
 	}
 })
 
+// Formatting for variable pulldown options:
+const CustomOption = React.memo((props: OptionProps<DropdownChoiceInt>) => {
```
Julusian: As these are similar (identical?) to the ones for the text input dropdown, perhaps it should be the same component in a 'DropdownUtil.tsx'?
There might be some differences though which makes this not viable.

webui/src/scss/_controls.scss
```diff
@@ -37,19 +37,34 @@ label.disabled {
...
 	}
 
+	&.var-label {
+		font-weight: inherit;
+		white-space: wrap;
+		font-weight: normal;
+	}
+}
+
+/* note that using `span` instead of `div` doesn't format correctly for the ellipsis */
```
Julusian, line 55: This ellipsis is probably something we want for the other version of this too?

webui/src/Controls/InternalModuleField.tsx
```diff
@@ -225,6 +227,28 @@ export const InternalPageIdDropdown = observer(function InternalPageDropdown({
...
+	return (
+		<SelectComponents.Option {...props} className={(props.className ?? '') + 'variable-dropdown-option'}>
+			<span className="var-name">{data.value}</span>
+			<span className="var-label">{data.label}</span>
+		</SelectComponents.Option>
+	)
+})
+
+// Formatting for variable "single value" (shown when dropdown is closed) -- uses a different CSS class
+const CustomSingleValue = React.memo((props: OptionProps<DropdownChoiceInt>) => {
```
Julusian: As with elsewhere, this feels rather squeezed. Visually it doesnt match how we do padding elsewhere.

I know there have been some comments that we have too much spacing/padding, but I think being consistent is more important

webui/src/Components/DropdownInputField.tsx
arikorn: Done but it may be more than you anticipated, and quite honestly it doesn't really improve upon this particular `type`-ing mess.  Yet... maybe what I did is a good thing in itself anyway. Also, considering your comment on line R231, I took a slightly different approach:

The the current "custom" code and formatting appeared to be specific to variables since it uses `DropdownChoiceInt` which is defined in a misleadingly-named file, _LocalVariableDefinitions.tsx_. In fact it's used for all types of variables and even some things that aren't variables. So I did a bit of refactoring and renamed the file _DropDownInputFancy.tsx_, which now includes the `CustomOption` (which is shared with text input) & `ConstomSingle` functions too. So now any caller can request fancy-formatted two-line input as long as the options conform to  `DropdownChoiceInt`.)

webui/src/Controls/InternalModuleField.tsx
arikorn: Done. Combined as described in  my response to your comment on line R28.

webui/src/Controls/InternalModuleField.tsx
arikorn: Hmm... well, it feels a bit squeezed...but different than elsewhere! Personally, I don't mind it, but I added a little padding in CSS `.select-control__value-container ` -- feel free to fine-tune it to your taste. Happily, it doesn't seem to affect the one-line single-values (for small-enough values), so that's good.

webui/src/scss/_controls.scss
arikorn: Is there another version?  This is for the closed menu, but only applies to the "Custom" feature. The ellipsis works on standard pulldowns, and there's no "single-value" for the suggestions menu. (And multi-value is its own beast, of course.) In the dropdown itself, options wrap -- this was true before this PR as well.

webui/src/scss/_controls.scss
arikorn: Well, to quote you on line R242 😀 :
>I know there have been some comments that we have too much spacing/padding, but I think being consistent is more important

Well, the goal of this PR _is_ indeed to make the menus more consistent, considering that there are currently [three styles of variable pulldowns](https://github.com/bitfocus/companion/pull/3791#:~:text=Action%20Dropdown%2C%20open-,Unchanged,-%3A) with completely different looks -- even the word-order differs! However, may I suggest going the other way and tightening up the variable-suggestions menu in text-expressions instead? We could try it, considering that it has already been requested, and if other people complain I've made it very easy to change.

As far as the choice of spacing, this is actually a compromise -- I started at 1.0 and still like it. I changed it to 1.1 because of an error in my that that was later fixed, so I'd actually be happy to go back to 1.0! FWIW, as you know, 1.1 is the line-spacing you use to draw buttons.  Also, note that this bit of CCS affects all dropdowns that use react-select (or more precisely, DropDownInputField.tsx, which sets the class to `select-control`) so the UI would be consistent. The exception, previously, was the variable-suggestion menus, which I've now fixed for the non-monaco implementation.

arikorn:
> I like the idea of this, but I think that this dropdown and the variables dropdown should be styled identically.

Done. Unless you count the monaco-editor menu, which hopefully I can leave to you (or you can point me in the right direction?). It looks so different that I've assumed there's a reason you didn't just reuse the existing TextInputField version.

Everything should be in order for this PR now, barring fine-tuning of things like padding, to your taste. (It's all near the end of _controls.scss.), or if I didn't understand the comment about the ellipsis.

Oh, I added _.cache_ to the eslint exclusions because it made `yarn lint` blow up on something left over from your surface plugins branch.

arikorn:
This may or may not help... here is a visual aid to the effect line-height on the drop-down menu, "actual size":

image

<br/><br/>

And here's a mapping of  line-height to `suggestLineHeight` for the monaco editor

|Line-height | Row Height (pix) |
|---|---|
| 1.5 | 59 |
| 1.4 | 56 |
| 1.3 | 53 |
| 1.2 | 51 |
| 1.1 | 48 |
| 1.0 | 45 |


These were done by adjusting line-height in the CSS for `.select-control__option, .variable-select-root__option` in _controls.scss (which affects all dropdown menus but monaco) and measuring the height of the blue section. For the monaco editor suggestions, `line-height` in `.main` (_expressions.scss_) should be set to around 40% of  `suggestLineHeight`.

webui/src/scss/_controls.scss
Julusian: it looked like this was enabling it for the dropdown options, but only for some dropdowns. but as long as the different types behave the same I am happy

Julusian:
monaco being different didnt bother me. Perhaps it was partly laziness (I avoided touching the styling as much as possible, and partly because I didnt want to spent days going down the rabbit hole of trying to make it look good/match.

I am not keen on the spacing change though:
image
partly because its visibly uneven, with no padding above the first line, but loads below the second line.

I am also not 100% sold on the font size of those suggestions, purely because I think the suggestions are now larger than the actual text field.

----

For me the main thing is that the rest of the dropdowns should be visually consistent with each other and the rest of the app. So I dont think that reducing spacing within the dropdowns is a good thing to do right now, because while they can all match, they wont match the rest of companion.  
(yeah I suppose I am arguing for styling monaco, but I really dont enjoy fighting css..)

----

My only other comment is that I dont like the size change of this based upon where the focus is

https://github.com/user-attachments/assets/f4009809-4762-4b69-84d2-142def3ad07a

arikorn:
I fixed your three issues, see details below. That said, if you're still not satisfied with the monaco suggest-widget changes, lets just revert it for now: the main goal of this PR is to make the "standard" dropdowns more readable/neatly formatted. Monaco was really just a "bonus" change since the text is already formatted in two lines, so it's just a bit of icing...

To summarize this PR: 
- This PR formats action & feedback dropdown menus that select Variables to match the variable suggestions in button text (not button expression text), i.e. two lines, the first line boldface name of the variable. 
- Any caller of `DropdownInputField` can now request this "fancy" formatting by setting `fancyFormat` in the call and using `DropdownChoiceInt[]` for the options.
- A side-effect/bonus is that it allows developers to make two-line entries using `'\n'` in any dropdown menu.
- A second side-effect is allowing menu row-heights to be adjusted in one place (_control.scss_) for all callers of `DropdownInputField` and `TextInputField`. Previously line-height was effectively 1.5; I've set it to 1.3 in the latest commit for both the "closed-menu" value and the "open-menu" options (see my previous comment for size comparisons). While it's easy enough to set to 1.5, that big seemed "too-much" on the closed-menu values, and I suspect only the extreme "connoisseur" would notice a difference vs. static parts of the UI. (And as I pointed out earlier, the buttons themselves use 1.1 line-heights...). And just to reiterate -- please feel free to change this or request me to do so if you still disagree. 
- As a bonus in this PR, I adjusted the monaco suggest-widget to look more like the others. This change is _not_ essential to the PR and can be broken off as a separate PR (or abandoned) if it's not working out.

Details
---

1. Input box size change: **_Fixed_**. Notes:

    > I [Julian] dont like the size change of this based upon where the focus is

   Size change based on focus is a side-effect of the fact that when `allowCustom = true`, and the user selects the input, it bypasses `CustomSingleValue`.  I couldn't find a direct fix, so I added a special CSS class to fix it.  (Personally, I would have just set `allowCustom` to false, but a comment on the relevant line makes it clear that someone feels it is important. No problem!)

   Size is now stable, though text jumps down a bit to stay centered.

2. Vertical centering of options (rows) in monaco suggest widget: **_Fixed_**.
3. Width of the monaco suggest widget: **_Improved_**. Notes:

    This is the pre-existing behavior and is independent of font size (see examples below, using the latest beta version in `main`). I couldn't find a fully satisfactory dynamic-sizing solution because monaco didn't let me wrap the text (or should I say, I couldn't figure out the magic "key" to this door), so with dynamic sizing, when the parent-window is too narrow, the text is clipped too much. 

    My current best solution is to make it "reasonably wide" but not too wide.  I added `width: 25em` under `.monaco-editor .suggest-widget`, which seems the ideal size given the current font size.

Monaco `suggest-widget` behavior prior to this PR
---
Comparison of this PR (top-half) to main (below red line)
image

<br/><br/>
Another example of the suggest-widget width in main (in an action expression)

image

arikorn:
Oh dang! I probably misinterpreted what you meant by:
> I am also not 100% sold on the font size of those suggestions, purely because I think the suggestions are now larger than the actual text field.

I thought you meant the width of the dropdown, but you probably meant that you want the font-sizes to match.  [~~Can do!~~ _edit_: Turns out that the font size is the same - see next comment] (~~Although, dare I point out that the increased font-size was to make it match the other variable-suggest pull-down menu?~~ And that in fact it's the font-size inside the monaco editor that's not standard?)

But this brings us back to the broader uniformity issue. I just realized that we're both interested in uniformity of the UI but focusing on different aspects: you are focusing on overall text-line spacing while I've been focusing on the size of the "single-value" (i.e. closed-menu) box. With line-height around 1.0, the input-box height is about the same as the one-line boxes (see comparisons below) -- which helps keep overall vertical size down too -- and it also keeps the box closer to the height of the label, which is always one line. For me, the decreased line-spacing within each option seems natural because it emphasizes the line groupings (there's still plenty of space between options). 

In the end, the differences are fairly subtle, which is why I'm more than happy to go with whatever you choose.

OK. I'd better wait for your reply before "fixing" anything!

---
Input-box line-height 1.0 (compare the size of the "**Custom Variable**" input-box in "internal: Custom Variable: Set with expression" with "**Surface / group**" input-box in "internal: Surface: Set to brightness"):
image

---
Line-height 1.5:
image

arikorn:
My claim about font-size of the monaco-suggestions menu was wrong: The font is exactly the same size as the text in the edit box (even though the code says it's one pixel larger) -- in the image, below, I copied "toFixed" from the editor text, made it translucent and overlaid it on the suggest-menu text to illustrate the point. The line spacing in the suggest-menu was a bit more than in the text editor, so in the latest commit I adjusted it so that the line-spacing within each two-line suggestion is identical to the line spacing in the monaco editor. (not shown). Bottom line: this PR should be good to go!

image

---
P.S. note the edit in my [list of features, above](https://github.com/bitfocus/companion/pull/3791#issuecomment-3621825751): this PR also makes it "easy" for other (internal) menus to adopt the two-line style.

arikorn:
OK. Last change, I sure hope...  It seems I can't even make up my _own_ mind about line spacing, so here it is again complying with the standard line-height ( `--cui-body-line-height`, which is 1.5), as per your last comment, and with a little tidy-up. It's still very easy to change it to anything else...

I also noticed that in certain circumstance the monaco suggest menu get covered over by the red banner on top, so I "improved" its stacking order.

**NOTE: The lint error is not in a file touched by this PR!**

Julusian:
I am not immediately sold on the new width of the monaco suggestions, a few of my modules have variables which are getting close to being the full width of it, which makes it feel a bit tight. That said, Im not against it, but I suspect there a bunch of modules will be too wide, and it will make this a little annoying to use.

An easy case of this is some internal variables:

image


> My claim about font-size of the monaco-suggestions menu was wrong: The font is exactly the same size as the text in the edit box (even though the code says it's one pixel larger) -

You are right that it is the same size, but I think its because its bold it feels bigger which makes it feel off to me. As shown in your screenshot, it isnt taller, but is wider

arikorn:
After comparing bold with normal and this PR with the current beta it turns out that the bold-face does not change the width, it's all due to the font size. I had bumped it up to 16px from 14px, because 14px is really small and 16px is more standard... 

I just pushed with the restriction on width removed and font size back to 14px. The amount of text shown is now identical to the current beta.  I have to say that (a) since the width is fixed w/o wrapping you'll always run into this problem but (b) the two-line format ameliorates the problem because it "fills in" the clipped information -- at least in your example above.

The main gain is in expanding the width, so I do recommend resetting `suggestFontSize` font size back to 16  (or at least 15 to match the expression text size) in `ExpressionInputField` before merging, but I'm leaving the final decision to you, of course. Yes, you do clip a few more characters in these edge cases, but you gain readability in the remaining "90%" of the cases and at least in this example, the missing information is shown on the second line...

---
FWIW, here's the **behavior in the current beta** using default emulator IDs, to show that you can't really win on this as long as we can't wrap. 

image

PR #3796, opened 2025-11-30
Feat: Improve TCP/UDP Error reporting

arikorn:
- log exactly what was received. (debug)
- log the string that caused a syntax error (info)
- document the need for `\n` terminating commands in TCP
- tidy up some internals (`type` of `chunk`)

This PR was inspired by [difficulties encountered by a FB user](https://www.facebook.com/groups/companion/permalink/4127561674128818/).

Note: the TCP/UDP code makes around 40 calls to report parsing errors (`ApiMessageError`) but this PR only "improves" four of them (particularly the "syntax error" ones) . The others could be improved too, if desired, but this PR makes it a bit less necessary by logging what was received.

### DIFF
--- a/companion/lib/Service/Tcp.ts
+++ b/companion/lib/Service/Tcp.ts
@@ -88,11 +88,12 @@ export class ServiceTcp extends ServiceTcpBase {
 				})
 				.catch((e) => {
 					this.logger.silly(`TCP command failed: ${e}`)
-					this.logger.info(`TCP command failed. ${e}`)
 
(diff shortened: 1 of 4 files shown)


arikorn:
With the second commit (Fix tests, tread more lightly), TCP output in responses to syntax errors is preserved. However, the TCP response _is_ changed for USB rescan failures.

If the latter is anticipated to cause problems, it's easily reverted. 

Conversely, if a more detailed TCP response to syntax errors is desired, that too is easily done.)

PR #3797, opened 2025-11-30
Restore default webpack optimizations

arikorn:
The optimization had been removed because of a bug in 5.102.1 that is fixed in 5.103.0

FWIW, my previous testing suggested it doesn't do too much, but it does reduce the number of modules in the final build, which in theory can improve performance.

### DIFF
--- a/companion/webpack.config.js
+++ b/companion/webpack.config.js
@@ -83,22 +83,4 @@ export default {
 				})
 			: '',
 	].filter(Boolean),
-	optimization: {
-		// The following optimizations are enabled only in 'production' mode
-		// 	flagIncludedChunks: optimizeUseProduction,
-		// 	moduleIds: 'named', // optimizeUseProduction ? "deterministic" : 'named'
-		// 	chunkIds: 'named', // optimizeUseProduction ? "deterministic" : 'named'
-		// 	sideEffects: 'flag', // () => (production ? true : "flag"),
-		// 	usedExports: optimizeUseProduction,
-		// 	innerGraph: optimizeUseProduction,
-		// 	mangleExports: optimizeUseProduction,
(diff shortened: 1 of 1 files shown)


companion/webpack.config.js
Julusian, line 86: It might be good to remove this whole object now?
I don't expect we will touch it again, so its just some stale code that will gradually fall out of sync with webpack

arikorn: OK, done. My thinking was that it might be nice to have in case someone ever wanted to play with them, especially if a similar type of error cropped up. Of course I was assuming that the list of optimizations would never change 😀 ! Reconstructing the list is not trivial but not excessively difficult, FWIW...

Julusian: Presumably it is almost entirely in https://webpack.js.org/configuration/optimization/ ?
But personally I would only bother defining the ones I am changing or want to explicitly set, so compiling that list should be pretty easy.

But this will be in the git history still, if someone does want it

arikorn:
@Julusian 
> Presumably it is almost entirely in https://webpack.js.org/configuration/optimization/ ?

Well, just for the record and to make things a bit easier if it's ever needed again: The [documentation on `mode`](https://webpack.js.org/configuration/mode/) was a bit vague as to what changes between `developer` and `production` mode, so I collated the optimizations that are affected by the webpack mode from here: [webpack/lib/config/defaults.js](https://github.com/webpack/webpack/blob/0f8b31bffe566178ca48782a96eff53bade8ba80/lib/config/defaults.js#L1822), FWIW. (There's a link to it, too, on the mode documentation page, though without the line number.)

If something goes wrong when only in one mode or the other, it's a good list to work through!

PR #3800, opened 2025-12-02
Feat: Restore the two-button version of Full Import

arikorn:
The problems with a single "import" button are (a) that it's easier to forget to check or uncheck the "full reset" checkbox and (b) that the state of the checkbox is not preserved.

This PR also addresses an concern brought up by a FB user, that the "advantage" of a partial reset is not made clear, i.e., that it preserves Settings.

(Incidentally, the rearrangement of `onSubmit` and `onClick`, etc., is taken directly from the Tanstack help pages, [here](https://tanstack.com/form/v1/docs/framework/react/guides/submission-handling) )

### DIFF
--- a/webui/src/ImportExport/Import/Full.tsx
+++ b/webui/src/ImportExport/Import/Full.tsx
@@ -146,8 +146,13 @@ const defaultFullImportConfig: ClientImportSelection = {
 
 const { fieldContext, useFieldContext, formContext } = createFormHookContexts()
 
+type FormMetaData = { fullReset: boolean }
+
+const defaultMeta: FormMetaData = { fullReset: true }
+
 const importFormOpts = formOptions({
 	defaultValues: defaultFullImportConfig,
+	onSubmitMeta: defaultMeta,
 })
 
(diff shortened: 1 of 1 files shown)


arikorn:
Thank Julian!

Julusian:
ooh thats fun, didnt realise sentry was able to do that.

not really sure what to make of what it is saying though, or if anything should be done

image

arikorn:
When I opened _Full.tsx_, VS Code signalled errors starting with:
> Cannot find module '@tanstack/react-form' or its corresponding type declarations.

I ended up removing then re-adding `@tanstack/react-form` from webui to fix the problem.

Is it possible that something got corrupted in the GitHub repo too?  The error message does suggest that `i` (aka. `form`) is believed to be undefined -- though why it flags that one and not the earlier one is a bit strange.

FWIW, the code works and I don't see any error messages.

Julusian:
I felt like digging into this today, it turned out to be a bug in the form library.. https://github.com/TanStack/form/pull/1924

PR #3802, opened 2025-12-03
feat: rework presets panel to use collapsible panels instead of a second page of buttons

Julusian:
Reworks the presets page.

Inner drawing is the same, and collapsed preset categories are not rendered (avoiding extra cost of streaming the bitmaps)

image
image


arikorn:
I like this idea! Much smoother flow than all the back-and-forths.

PR #3803, opened 2025-12-04
Add the non-developer pages from Companion Wiki

arikorn:
Most of the [companion wiki](https://github.com/bitfocus/companion/wiki) pages have now been transfered to the "For Developers" section of the new [Companion Documentation site](https://companion.free/user-guide/v4.1/). However the pages that are more suited to the user guide must be added here rather than in the website repo...

Consider this PR a suggestion and/or accounting for the remaining files in Companion Wiki.  With this PR, all pages in the Wiki can be removed or replaced by links to the new website docs.

Note: I've updated part of the "local modules" page for v4.1, but otherwise did not change content except to fix links.

Note 2: Although it may seem to not make sense for the Install page to be included in pages that are seen from the Admin-page sidebar-link, it's actually still relevant because it discusses additional options. 

(One page, "installing Companion Pi", was just links, so it was redundant with the main install page.)


docs/user-guide/1_getting-started/Installation.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Windows (x64)
+
+Note: As of 3.0 we only support Windows 10 and newer.
+
+Download, install and run!
+
+### Raspberry Pi
+
+> **Please note:** Companion is only supported on the Raspberry Pi 4 (2, 4, or 8 GB). Companion can be installed on a Raspberry Pi 3B, or 3B+, **but it is not recommended or supported**. Should you choose to do so, you do so at your own risk and with the understanding that the community will not be able to help you if something goes wrong.
```
Julusian:
```suggestion
:::tip
**Please note:** Companion only supports the Raspberry Pi 4 (2, 4, or 8 GB). Companion can be installed on a Raspberry Pi 3B, or 3B+, **but it is not recommended or supported**. Should you choose to do so, you do so at your own risk and with the understanding that the community will not be able to help you if something goes wrong.
:::
```

docs/user-guide/1_getting-started/local-modules.md
Julusian, line 1: I dont think that this page is relevant anymore, as 4.0 has the newer way to import pacakged modules, and doing all this should be left for module developers.

docs/user-guide/1_getting-started/Using-a-reverse-proxy.md
Julusian, line 1: I feel like this should be in some 'advanced setup' grouping rather than getting started.

I worry that this will scare or confuse non code users

docs/user-guide/5_remote-control/satellite.md
Julusian, line 1: I dont think this should be here, but should be under maybe the root level of the 'for developers' section on the website. This file is a definition of the protocol after all, and can only realisitically be used by developers to make devices which talk the satellite protocol

docs/structure.json
Julusian, line 1: yeah this is still needed for now, as this arranges the docs on the bitfocus website

arikorn: Thanks Julian.

docs/user-guide/5_remote-control/satellite.md
arikorn: Done here and there. Reverted the satellite page to its original text then changed the link to the new website (which will be broken until that PR is concluded).

docs/user-guide/1_getting-started/local-modules.md
arikorn: Done. I moved it to the developer pages for now (and edited it to make it specific to module development).

docs/user-guide/1_getting-started/Using-a-reverse-proxy.md
arikorn: For now I've renamed it "Advanced Setup Options" and left it at the bottom of "Getting Started", where it should be mostly out of the way. Also, made it clear in the text that it's not something for the average user to worry about.

Please suggest any alternative you would prefer, if this is not acceptable!!

PR #3804, opened 2025-12-04
Add support for Stream Dock M18V3 model

allgood:
- Recognize productId 0x1012 as Stream Dock M18V3
- Define M18V3 model with 18 buttons and 15 LCD outputs
- Update model selection logic for new device

### DIFF
--- a/companion/lib/Surface/USB/MiraboxStreamDock.ts
+++ b/companion/lib/Surface/USB/MiraboxStreamDock.ts
@@ -2574,7 +2574,7 @@ class StreamDock extends EventEmitter {
 		} else if (this.info.productId === 0x1003) {
 			// modelType = 'N3'
 			this.model = StreamDock.models['293N3']
-		} else if (this.info.productId === 0x1012) {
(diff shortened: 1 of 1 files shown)


companion/lib/Surface/USB/MiraboxStreamDock.ts
```diff
@@ -2302,6 +2574,8 @@ class StreamDock extends EventEmitter {
 		} else if (this.info.productId === 0x1003) {
 			// modelType = 'N3'
 			this.model = StreamDock.models['293N3']
+		} else if (this.info.productId === 0x1012) {
```
Julusian: Please also include a check of the vendorId here, we will soon need this for all models so by defining this new one now it avoids the risk of me guessing wrong when filling it in later

allgood: > Please also include a check of the vendorId here, we will soon need this for all models so by defining this new one now it avoids the risk of me guessing wrong when filling it in later

Ok, done!

Issue #674, opened 2019-06-01
[Request] Background Blinking


cgarwood:
**Describe the feature**
The ability to set the background and/or foreground colors to blink between two sets of colors. Bonus if we can configure the speed at which it changes.

**Is this platform dependent (windows, mac, ..)?**
no

**If documentation is required to implement, do you know where to find it?**
n/a

**Usecases**
If you're recording or streaming and want to draw more attention to that fact, you could have the background of the recording button blink between black and red.

Blinking the button for an input that's currently live on a switcher

Drawing attention to something that's in an error state.

etc.

DachsbauTV:
Just somer more usecase examples:

- a muted microfon (never forget to unmute a commentator again after a break)
- a timer went to 0:00
- callback from an other position or device
- Battery level feedback
- on buttons with a lot of delayed steps it could be a great addition to see when it reached the last step

plasticpool:
I agree with all the use cases here – just knowing that a microphone channel is open (or not) with a flashing/blinking button state would be super helpful, and I'd be happy to contribute to a bounty toward implementing this.

mdgnys:
A great way to accomplish this would be a module that supplies instance feedback that is simply a constantly switching Boolean back and fourth.

shaunar2006:
This would be awesome! I need this!

justingiffard:
This would really be awesome, from a dev point of view it would be cool if there was a way to specify the frequency of the flashes too. I'd like to add feedback to the X32 module for the tap delay that it gives the user a visual idea of what the delay is like the tap delay on the console

HaxNobody:
Here's how I implemented blinking buttons:
1. Set up a new custom variable, let's call it "Flash" for now. You should be able to use the same variable for multiple different buttons, and you can create more variables for different flash patterns.
2. Set up a trigger that is also called Flash or something similar, and configure the Condition to be Time Interval
3. Set the time interval to 1 (seconds)
4. Tick the "Relative Delays" box
5. Add an Action for "internal: Set custom variable value", configure the Delay to 500ms, select your custom variable, and set the Value to 0
6. Add another Action the exact same as step 5, but set the Value to 1
7. Configure your button to add a Feedback of "internal: Check variable value" and set it to your custom variable, with default '=' Operation and the Value as 1.
8.  Finally, configure the text color and background color that you want to switch to with each flash, different from the initial setting at the top of the button config.

Done - you can adjust all the trigger timings to make your flasher work the way you want it to.

[...]

PR #3828, opened 2025-12-10
Feat: native blink feedback

alex-Arc:
[#674](https://github.com/bitfocus/companion/issues/674)

### DIFF
--- a/companion/lib/Controls/ControlTypes/Button/Normal.ts
+++ b/companion/lib/Controls/ControlTypes/Button/Normal.ts
@@ -53,6 +53,7 @@ export class ControlButtonNormal
 		color: 0xffffff,
 		bgcolor: 0x000000,
 		show_topbar: 'default',
+		blink: null,
 	}
 
 	readonly supportsActions = true
--- a/companion/lib/Controls/Entities/EntityInstance.ts
+++ b/companion/lib/Controls/Entities/EntityInstance.ts
@@ -478,6 +478,9 @@ export class ControlEntityInstance {
 				if (key === 'png64' && !newStyle[key]) {
 					newStyle[key] = null
 				}
+				if (key === 'blink' && !newStyle[key]) {
+					newStyle[key] = null
+				}
 			}
 
 			if (key === 'text') {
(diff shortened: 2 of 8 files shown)


thedist:
I'd recommend holding off on doing anything like a native blink feedback until after the graphics rework as that may change things.

Julusian:
Thanks for looking at this, but my immediate thought is that this isn't something I want to tackle before the graphics overhaul is done.

I have 2 immediate concerns:
1) I expect that many users will find the choice to blink the background to black to be too limiting. I suspect that this will result in this needing to be expanded to allow configuring every style property.
  This concerns me as this patch of ui is already not very nice; I have been holding off adding more style options for a while as adding anything more makes it feel incredibly cluttered.
2) I don't currently see how we can translate this to the new drawing ability in the graphics overhaul. I believe that everything else translates well enough that we should be able to auto-migrate existing buttons. But I think that this blink largely goes against the new drawing ideas and will be very hard to migrate across without making a mess in some form.

The combination of those concerns has me leaning towards waiting until the graphics overhaul is finished (or abandoned). Until then, I hope that either a custom/expression variable or the generic-blink module are good enough (although something native would be nicer)

alex-Arc:
Thanks both for the comments.
I will take a look at the overhaul branch.
I understand this might not be soon but will keep my hopes up 🙂

Julusian:
@alex-Arc this PR got me inspired to have a play. I had an idea earlier of what if there was simply an expression function to do this blinking? That way it can be used in a feedback to change whatever properties the user wants, and that will fit into the graphics overhaul.

I hope you don't mind, but while I was feeling inspired I gave it a try, and I think it works quite nicely. #3831
I am a little worried about the performance cost of piping so many variable updates through the system to facilitate this, so I am thinking it wants to wait for some investigation into that. (Unless @thedist disagrees; you have more experience with large amounts or frequent variables updates)

alex-Arc:
> I am a little worried about the performance cost of piping so many variable updates

Yeah that was also my concern and the reason I moved it down into the render function

Julusian:
Thanks for the nudge on this, but I am going to close this in favour of #3831 which I am hoping will be merged before 4.3 (should happen unless anyone objects to it)

PR #3851, opened 2025-12-18
feat: Report UI/Server URL at end of startup (feat for developers)

arikorn:
Near the end of startup, Companion reports something of the form:

> info UI/Server new url: http://127.0.0.1:8000/

In VS Code and PowerShell ctrl-click on this text provides a convenient way to start the Admin UI, but with the extra reporting coming from the new surface module feature, many items are logged after this line, making it difficult to find.

This PR delays the the call to `UIServer.rebindHttp` briefly so that the log line ends up near the end of the log at startup. (Also: log the same URL as is sent out over IPC.)

Note: moving the call to the end of the calling function was not sufficient. OTOH, I could just delay the log message rather than the whole call...

## Summary by CodeRabbit

* **Chores**
  * Ensure binding configuration updates are applied before triggering the HTTP rebind operation.
  * Delay server connection and error log messages by 2 seconds for clearer, formatted URL output.

### DIFF
--- a/companion/lib/Registry.ts
+++ b/companion/lib/Registry.ts
@@ -478,7 +478,7 @@ export class Registry {
 		this.services.https.updateBindIp(bindIp)
 		this.internalModule.updateBindIp(bindIp, bindPort)
 		this.usageStatistics.updateBindIp(bindIp)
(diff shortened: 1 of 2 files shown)


arikorn:
Scolded by the rabbit AI to "do the right thing"! 🤣

PR #3869, opened 2025-12-29
Fix: "Unset" surface variable "breaks" equipment

arikorn:
Companion v4.2 eliminated the internal jog/shuttle/t-bar variables in favor of user-specified custom variables. While this is a clear improvement in behavior, the default of "none" (no variable) causes "null" values to be sent to my mixer, which caused it to fail internally (needed reboot to restore full functionality).

This PR restores the now-deprecated internal variables in the case that 'none' is selected, so that it becomes non-breaking. (A very minor side-effect: a variable will always be set upon one of the surface events, even if 'none' is selected. Since this corresponds to the behavior in v4.1, it shouldn't be an issue.)

**Note that this PR is a hotfix for v4.2**: it restores v4.2 to be non-breaking (at least in this regard) as might be expected in SemVer....

If this PR is accepted, I will leave it to Julian to decide whether this should be implemented in v4.3 (i.e. `main`) as well, where surfaces are in independent modules. (And whether he wants me to do it!)

**Note 2**: I tested this on a Contour Shuttle and assume it will therefore work with the XKeys and Loupedecks as well.

**Note 3**: BlackmagicController is not reverted here because it was already using variables prior to v4.2

### DIFF
--- a/companion/lib/Internal/Surface.ts
+++ b/companion/lib/Internal/Surface.ts
@@ -155,6 +155,14 @@ export class InternalSurface extends EventEmitter<InternalModuleFragmentEvents> 
 		this.#controlsController = controlsController
 		this.#pageStore = pageStore
 
+		setImmediate(() => {
+			this.emit('setVariables', {
+				't-bar': 0,
+				jog: 0,
+				shuttle: 0,
+			})
+		})
+
 		const debounceUpdateVariableDefinitions = debounceFn(() => this.emit('regenerateVariables'), {
 			maxWait: 100,
 			wait: 20,
@@ -248,7 +256,20 @@ export class InternalSurface extends EventEmitter<InternalModuleFragmentEvents> 
 	}
 
 	getVariableDefinitions(): VariableDefinitionTmp[] {
-		const variables: VariableDefinitionTmp[] = []
+		const variables: VariableDefinitionTmp[] = [
+			{
+				label: 'Legacy T-bar position (deprecated)',
+				name: 't-bar',
+			},
+			{
+				label: 'Legacy Shuttle position (deprecated)',
+				name: 'shuttle',
+			},
+			{
+				label: 'Legacy Jog position (deprecated)',
+				name: 'jog',
+			},
+		]
 
 		const surfaceInfos = this.#surfaceController.getDevicesList()
 		for (const surfaceGroup of surfaceInfos) {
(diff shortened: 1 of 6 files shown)


Julusian:
Starting on the future of this in 4.3; I really dont want to do this for 4.3, as that would mean adding it to the new surface api, which means it will be present there for many years. Seeing as we are still on module-base 1.x after 3 years, the surface api will likely move even slower. So I really don't want to have this method/ability be there for years after its functionality is finally killed off (in I guess Companion 5 if this does get done).  

Additionally, in the new surface flow, the surfaces don't know if the value they are producing is bound to a variable, so the behaviour of this would want to be adjusted to always produce both to match what would be possible in the surface-api.

Then the question would become whether to restore this functionality in 4.2.3, after being documented as being removed in 4.2.0, to then remove it again in 4.3.0.
Tbh I think that keeping the old behaviour is just postponing the slight pain users will experience. Most won't notice that the way they are doing it has been deprecated, so wont update until it breaks. Which is why I didn't do this mixed behaviour to begin with.

Or it could be added to the surface-api so that it stays until we remove it in hopefully 5.0 (although I already forgot to do this when 4.0 was done) whenever that happens to be.  

Tbh, I treat the versioning of companion as semver-ish. While this is a breaking change, it is low impact enough that it felt fine to drop in a minor. (In my mind, it should be as simple as updating the couple of triggers that are reacting to these vars, and should impact less than 1% of users based on our knowledge of types of surfaces being used)

arikorn:
I totally agree with your sentiment, Julian, yet having been "burned", see things a little differently:

1. My main concern is that this bug affected equipment in a way that couldn't be fixed in middle of recording (in this case even the GoStream's built-in menu stopped working after being sent the null value).
2. It's not that unusual, in a patch release, to revert a feature -- in fact you just did it in 4.2.2 (and in 4.1.2, I think, you reverted a feature introduced in a major update, for example).
3. I would guess that most bugs that gets past a minor/major release affects less than 1% of users. In fact, in this case, one could argue the other way around too: since the code itself is reached only by the affected user base, there's no harm in restoring it...
4. While I can't speak for others, I use these variables in button actions, not triggers, so it's easier to miss some.

So here's my proposal for a phased removal that may be a bit gentler on the affected users...
1. ~~Apply this PR to 4.2 (but if you really object, step 2 is still acceptable)~~ [edit: never mind, better to go directly to step 2 in 4.2.3]
2. In 4.3 (and 4.2.3) just keep the internal variables defined but not updated. This will let the "forgetful" user know that something is wrong without sending null/NaN values to modules/equipment that aren't hardened enough. [edit I ~~can transfer~~ transfered the lines into PR #3872 . So if you agree, you can close this one and just take that one. ]
3. In 5.0 the internal variables go away too.

This way the surface-module API is not affected but the main issue is addressed (i.e. equipment is not sent illegal values).

arikorn:
Just to call out the edit above: I'm now recommending PR #3872 over this one: it contains the bug-prevention portions without restoring the old functionality.

Julusian:
Replaced by #3872

PR #3872, opened 2025-12-29
Fix invalid values being sent to a surface

arikorn:
This restores the internal t-bar/jog/shuttle variables and marks them as deprecated. The variables are never updated but exist to prevent null values being sent to a surface.  See PR #3869 for details.

**NOTE: I am posting this as a convenience, not a nudge**! If you disagree with the proposal in [my comment on 3869](https://github.com/bitfocus/companion/pull/3869#issuecomment-3697186142) just ignore or close this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Legacy surface controls (t-bar, jog, shuttle) are now emitted and initialized immediately when the surface starts, ensuring these controls are defined and available for use right at startup. This improves reliability of control state on launch and prevents transient unavailability during initial setup.

### DIFF
--- a/companion/lib/Internal/Surface.ts
+++ b/companion/lib/Internal/Surface.ts
@@ -256,23 +256,7 @@ export class InternalSurface extends EventEmitter<InternalModuleFragmentEvents> 
 	}
 
 	getVariableDefinitions(): VariableDefinition[] {
-		const variables: VariableDefinition[] = [
-			{
-				name: 't-bar',
-				description:
-					'DEPRECATED Legacy T-bar position. Value is never updated: link a custom variable in the surfaces page instead.',
-			},
-			{
-				name: 'shuttle',
-				description:
-					'DEPRECATED Legacy Shuttle position. Value is never updated: link a custom variable in the surfaces page instead.',
-			},
-			{
-				name: 'jog',
-				description:
-					'DEPRECATED Legacy Jog position. Value is never updated: link a custom variable in the surfaces page instead.',
-			},
-		]
+		const variables: VariableDefinition[] = []
 
 		const surfaceInfos = this.#surfaceController.getDevicesList()
 		for (const surfaceGroup of surfaceInfos) {


companion/lib/Internal/Surface.ts
```diff
@@ -248,7 +256,23 @@ export class InternalSurface extends EventEmitter<InternalModuleFragmentEvents>
 	}
 
 	getVariableDefinitions(): VariableDefinition[] {
-		const variables: VariableDefinition[] = []
+		const variables: VariableDefinition[] = [
```
Julusian: These definitions aren't needed, just the setting of the values.

While I don't particularly like having to keep these variables existing to handle a subset of possible bugs in some modules (the same crashes could happen in numerous other ways, such as simply mistyping a variable name),  I don't care enough to block this

arikorn: >These definitions aren't needed, just the setting of the values.

Interesting. But part of the point here is to give the user information about why the variable isn't updating and how to fix it. As CodeRabbit wrote above: it's " Excellent deprecation messaging!" while taking up almost no space. (For example, I count 50-100 Internal variables in my fairly simple standard config, so 3 extra variables is pretty inconsequential.)

As to module crashes happening in other ways, the problem this PR is addressing is to avoid breaking existing code not to prevent errors in general.  So I hope you can accept it with the deprecation messaging as-is.

Thanks!

arikorn: Well, in the end I don't care enough either, and since I don't know if your silence is due to distraction or to disagreement, I leave the rest to you: I just pushed a commit removing the definitions. If you prefer going with the Rabbit's and my preference, just undo the commit!

Thanks.

Julusian: distraction mostly.
I prefer without as it satisfies the task of helping broken modules, without adding noise to the ui.

PR #3892, opened 2026-01-11
feat: ability to enable/disable individual surfaces

Julusian:
Adds the ability to enable/disable individual surfaces. This is done within the settings of each surface:

image

There are some caveats:
* It does not apply to those connected over satellite (for now, that may be done as a follow up)
* It does not apply to any streamdecks connected over the network (either SDS or network dock)
* It may not work for modules which do their own detection mechanisms
These are mostly because we can't guarantee we are able to 'abort' the opening of the surface (not without entering a constant reconnect loop), so at this stage these cases are not handled, and may be improved upon later.

When disabled, surfaces will indicate as such in the table:
image

And there is a new setting to change the default behaviour of newly connected surfaces:
image

## Summary by CodeRabbit

* **New Features**
  * Added "Auto-enable newly discovered surfaces" setting and UI controls; per-surface enabled/canChangeEnabled fields and a visual "Disabled" indicator.
  * Surface discovery now uses a registry-based flow with stable, collision-resolved IDs and an explicit surface-close action; panels expose isRemote metadata and closing now triggers an IPC close.

* **Tests**
  * Added comprehensive tests for the new discovery registry; removed legacy ID-generator tests.

### DIFF
--- a/companion/lib/Surface/Controller.ts
+++ b/companion/lib/Surface/Controller.ts
@@ -62,7 +62,7 @@ import type { CheckDeviceInfo } from '../Instance/Surface/IpcTypes.js'
 export interface SurfaceScanHandler extends SurfaceOpener {
 	/**
 	 * Scan for surfaces
-	 * This either processes HID devices to find supported surfaces, or can prtform its own non-hid scan.
+	 * This either processes HID devices to find supported surfaces, or can perform its own non-hid scan.
 	 * @param hidDevices - HID devices with serialNumber already populated
 	 * @returns Promise resolving to array of discovered surfaces
 	 */


arikorn:
Oh... before you run with this, Julian, I have a suggestion/PR that interacts with the UI side of this, which was a follow-up to our dialog on your main surface-modules PR: the idea of bringing the settings and instance-table into the Configured surfaces page. It should be compatible, though may require a bit of merge-conflict resolution. (Sorry: I didn't realize you were so close to starting/finishing on this one!)

Julusian:
I think I would rather not wait for that before merging this, from what I remember of that discussion I dont think this change will cause any massive complications for that (other than perhaps some small merge conflicts because of the new ui fields, but they should be easy to translate across)
And I remember I wasn't convinced by your proposal, so I suspect it wont be a quick merge meaning it would be good to get this out the way first

PR #3902, opened 2026-01-13
fix: improve text drawing performance

Julusian:
The aim of this is to optimise the text drawing a bit. Improvements isn't massive, except when it has a cache hit.

The text segmenting and measurement step has been reshuffled and some minor tweaks made. And a cache has been added for the measurements, to try to reduce the cost of performing the line calculating

## Summary by CodeRabbit

* **New Features**
  * Added a reusable text layout utility and an exported text-layout cache for faster, consistent text rendering.

* **Performance**
  * Switched to layout-driven text rendering with caching to reduce repeated measurements and improve render throughput.

* **Behavior**
  * Rendering now uses precomputed layouts and will skip drawing when sanitized input is empty.

* **Tests**
  * Added extensive unit tests and a performance benchmark for text segmentation, sizing, and layout.

### DIFF
--- a/companion/test/Graphics/TextParser.perf.ts
+++ b/companion/test/Graphics/TextParser.perf.ts
@@ -8,7 +8,6 @@
  * to measure the actual computation cost.
  */
 
-import QuickLRU from 'quick-lru'
 import { Image, TextLayoutCache } from '../../lib/Graphics/Image.js'
 
 interface BenchmarkResult {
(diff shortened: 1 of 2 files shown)


arikorn:
Cool! I wonder if this would help performance of my animation module (it use button-text as the "pixels"). So to get a performance report on my own stuff, I'd have to change _companion/test/Graphics/TextParser.perf.ts_ ?

Julusian:
> Cool! I wonder if this would help performance of my animation module (it use button-text as the "pixels"). So to get a performance report on my own stuff, I'd have to change companion/test/Graphics/TextParser.perf.ts ?

Yeah quite likely it would help at least a bit.

Yeah you could try that. I now dont remember what that tests or if you should expect to see anything in it, and you need to run it before and after this change to get a comaprison

Issue #666, opened 2019-05-26
Media Player Classic (MPC)


oliveramays:
**Describe the feature**
I have not seen any mention or been able to find anything related to Media Player Classic and Companion so apologies if someone has already looked at this.
A module that uses the "Web remote" feature in Media Player Classic, able to send a whole range of commands for Media Playback, Playlist Controls, Image adjustments and more.

**Is this platform dependent (windows, mac, ..)?**
Windows based application, popular for video playback.

**If documentation is required to implement, do you know where to find it?**
I have already got this working with companion via the HTTP GET feature but would be great to have an actual module for it. I have made Companion pages on my stream deck that work great and have attached an Export (Pages 5 - 12 are my (Work in progress) Media Player Classic pages) I hope a ZIP file with the export is ok.
[MPC Companion Export.zip](https://github.com/bitfocus/companion/files/3220506/MPC.Companion.Export.zip)

Media Player Classic requires 'Listen On Port' ticked under the 'Web interface' settings.

Example of a URL for sending a command is `http://127.0.0.1:13579/command.html?wm_command=887`

This string is added as a "GET" command and is put into the URL box for each command when editing buttons.
The IP address is the Localhost IP and the Port is the number that MPC is listening on. The command is defined by changing the number at the end. 887 is the command for 'Play'

All command numbers as follows:
> 800 Open File
801 Open DVD
802 Open Device
805 Save As
806 Save Image
807 Save Image (auto)
809 Load Subtitle
810 Save Subtitle
804 Close
814 Properties
816 Exit
889 Play/Pause
887 Play
888 Pause
890 Stop
891 Framestep
892 Framestep back
893 Go To
895 Increase Rate
894 Decrease Rate
896 Reset Rate
905 Audio Delay +10ms
906 Audio Delay -10ms
900 Jump Forward (small)
899 Jump Backward (small)
902 Jump Forward (medium)
901 Jump Backward (medium)
904 Jump Forward (large)
903 Jump Backward (large)
898 Jump Forward (keyframe)
897 Jump Backward (keyframe)
921 Next
920 Previous
919 Next Playlist Item
918 Previous Playlist Item
817 Toggle Caption&Menu
818 Toggle Seeker
819 Toggle Controls
820 Toggle Information
821 Toggle Statistics
822 Toggle Status
823 Toggle Subresync Bar
824 Toggle Playlist Bar
825 Toggle Capture Bar
826 Toggle Shader Editor Bar
827 View Minimal
[...]

PR #3903, opened 2026-01-13
Feat: Make disabled surfaces clearer

arikorn:
The power-switch icon indicating disabled status on the left of the row is a bit too subtle for me -- you need to unequivocally hit me over the head with the status! 😀 

- Use "disabled" color for main text
- Show "Disabled" instead of "Offline" and make it red

image

## Summary by CodeRabbit

* **New Features**
  * Surfaces now display a clear "Disabled" status in the location/status column when appropriate.
  * A single, consistent condition determines the disabled indicator shown on surface rows.

* **Style**
  * Row and info-column styling updated to visually differentiate disabled surfaces.
  * Disabled rows use a muted color and dedicated disabled styling for improved readability.

### DIFF
--- a/webui/src/Surfaces/KnownSurfacesTable.tsx
+++ b/webui/src/Surfaces/KnownSurfacesTable.tsx
@@ -252,6 +252,7 @@ const SurfaceRow = observer(function SurfaceRow({
 
 	const surfaceDisabled =
 		!surface.enabled &&
+		!surface.isConnected &&
 		surface.integrationType !== 'emulator' &&
 		surface.integrationType !== 'elgato-plugin' &&
 		surface.integrationType !== 'satellite'


webui/src/Surfaces/KnownSurfacesTable.tsx
```diff
@@ -275,7 +277,9 @@ const SurfaceRow = observer(function SurfaceRow({
 							<FontAwesomeIcon icon={faCopy} color="#000" />
 						</CButton>
 					</CopyToClipboard>
-					<span className="surface-location">{surface.isConnected ? surface.location || 'Local' : 'Offline'}</span>
+					<span className={`surface-location${surfaceDisabled ? ' surface-disabled' : ''}`}>
```
Julusian:
```suggestion
					<span className={classNames('surface-location', { 'surface-disabled': surfaceDisabled })}>
```

Julusian:
I'm not a fan of the red disabled. I think that makes those rows stand out, having the opposite effect of dimming the row.

I woudnt be surprised if some users have mostly disabled rows

webui/src/Surfaces/KnownSurfacesTable.tsx
arikorn: You and the Code Rabbit both. Though for some reason the Rabbit retracted that comment in its final edit.
Ok, done.

arikorn:
@Julusian wrote:
>I'm not a fan of the red disabled.

Yeah, I was on the fence on that one, too. Just changed it to the disabled color. Note: Without the override, the status text color would be #666, which is also an option but a bit odd since instead of being lighter than the rest it would be darker relative to the row color.

I also changed the name `surface-location` to `surface-status` which I think is a bit more general given that the values are "Offline", "Disabled", "Local" or `location`.

Issue #3881, opened 2026-01-07
[BUG] secondstoTimestamp not working correctly in 4.2


sackboydad:
### Make sure you're on the latest stable or beta build

- [x] I have tested this on the latest stable or beta release

### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion and not a specific module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

The SecondsToTimestamp function does not seem to be working correctly in v4.2.   in v4.16 if I used SecondsTo Timestamp(385786) - or a variable that equals 385786 - it would return 107:09:46.  However, if I use the exact same function in v4.2.3 it will return 11:09:46.  This also happens on the current Beta of 4.30+8842 MacOS 26.2.

We use this all the time to calculate how many seconds from now until Sunday at 10:56:43.  It's like the secondstotimestamp function can't handle more than 24 hours any more.

### Environment (please complete the following information)

```markdown
- OS: MacOS 26.2
- Browser: Safari
- Companion Version: 4.23, 4.30+8842
```

thedist:
I believe it was this PR https://github.com/bitfocus/companion/pull/3668 by @alex-Arc where the function was updated with new functionality but it is a breaking change so it is going to mess up a lot of peoples setups unfortunately.

If you check out the docs for `secondsToTimestamp` you can see how to provide a format string to have it display days as well, for example `secondsToTimestamp(385786, 'd:hh:mm:ss')` would show `4:11:09:46`. Or you could split it up into 2 functions and display `4 days 11:09:46` or something like that.

It's not a breaking change I agree with, but it got merged so you may need to either change to working with the new formatting, or write an expression yourself that will parse the seconds to `107:09:46`.

alex-Arc:
Yeah i do see that this could be annoying 
Can we conceive of a formatting expression to help this?

Maybe `dhh` would not include day overflow and `hh` would not

Something in that direction?

sackboydad:
> Yeah i do see that this could be annoying Can we conceive of a formatting expression to help this?
> 
> Maybe `dhh` would not include day overflow and `hh` would not
> 
> Something in that direction?

You probably weren't addressing to me (the person that initially wrote), but I think this would work.  And I like the choice of the old way and the new way.

alex-Arc:
Don't know if we could come up with a better lettering 

An a second thing would be if it should be generic so that minutes also could be the largest unit

Julusian:
[...]

PR #3904, opened 2026-01-15
Feat: extended time formatting

alex-Arc:
#3881

## Summary by CodeRabbit

* **New Features**
  * Enhanced timestamp formatting: flexible multi-character unit tokens, bracket notation to mark a single primary (largest) unit, millisecond precision/padding controls (1–3 digits), AM/PM and 12-hour handling, improved negative/zero behavior, and clearer validation errors.

* **Documentation**
  * Updated guide for combined seconds/milliseconds formats, default formats, bracket semantics, padding rules, and time-offset examples.

* **Tests**
  * Added extensive tests for formats, edge cases, validations, padding/precision, and AM/PM behavior.

### DIFF
--- a/shared-lib/lib/Util.ts
+++ b/shared-lib/lib/Util.ts
@@ -145,6 +145,7 @@ export function msToStamp(v: number, format: string): string {
 		}
 		processChar(c)
 	}
+	if (inBracket && !outBracket) throw new Error('"[" was never closed')
 	processChar('')
 	return result.join('')
 }
--- a/shared-lib/lib/__tests__/expressions-functions.test.ts
+++ b/shared-lib/lib/__tests__/expressions-functions.test.ts
@@ -627,6 +627,9 @@ describe('functions', () => {
 			expect(ExpressionFunctions.msToTimestamp(0, 'hh a')).toBe('12 AM')
 			expect(ExpressionFunctions.msToTimestamp(12 * HOUR, 'hh a')).toBe('12 PM')
 
+			// brackets not closed
+			expect(() => ExpressionFunctions.msToTimestamp(1234567890, 'dd:HH:mm:[ss')).toThrowError('"[" was never closed')
+
 			// largest unit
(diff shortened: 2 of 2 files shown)


thedist:
One thing I'd suggest is that when this get merged it's ensured that the release notes for the next Companion release make clear this will be the time formatting functionality going forward, as last time it was not properly documented that it was a breaking change leading to not only breaking some peoples productions but also unsure of it would be reverted or not as it was never appropriately communicated.

Julusian:
Am I correct in thinking that this wont affect any existing expressions? (It looks that way from the tests)

I am wondering if this is safe to backport to 4.2 or if there is some risk of small behaviour changes from 4.2.3 (in which case it should be held for 4.3)

alex-Arc:
the parsing is a bit different, but the intended result should not be different 
unless some one for some reason used brackets inside the values then it would break

arikorn:
@alex-Arc, not sure how much this is really related to your PR, but I just noticed that  `msToTimestamp(unixNow(), 'hh:mm')` gives UTC time rather than local time., which makes some sense. OTOH, the`$(internal:time_...)` variables show local time, for example (except the `time_unix`, which is just unixNow()). It would be nice to have an option for local time here too, considering the compact formatting...  (Your reward for a job well done?)

alex-Arc:
@arikorn not quite sure what are we missing?
a localNow() function?

arikorn:
@alex-Arc that's a possibility, but a bit nonstandard: there's only "one" unix timestamp and it's UTC.

I don't think the current functions were meant for clocktime anyway, so probably the simplest would be to just make a new `date` or `fmtdate` or `datetime` function that takes your formatting arguments and uses whatever logic the `$(internal:time_...)` variables use to get local time.  (Assuming @Julusian et. al don't object.)

 Or, probably better, use the library that Julian mentioned in one of your previous PRs for this task - `dayjs` I think?

arikorn:
Oh, maybe PR #4021 is already planning to do this...

PR #3909, opened 2026-01-18
Feature: Make Sidebar groups more distinct

arikorn:
...and make the collapse/expand caret more visible.  (I have a hard time distinguishing the subgroup from the main sidebar items, and in particular, the caret is nearly invisible to me!)

Just some CSS tweaks. (The hover coloring is more noticeable when doing it interactively, but regardless, it's pretty easy to tweak.)

| Normal | Hover |
|----|----|
|image<br /><br />|image|

## Release Notes

* **Style**
  * Enhanced sidebar group visuals with a defined group background and improved contrast between group containers and individual links.
  * Expanded group items receive adjusted padding, subtle shadowing and clearer link spacing for better hierarchy.
  * Caret area styling refined for consistent alignment and background.
  * Darkened hover effect for group buttons to improve interactive feedback.


Julusian:
I'm not opposed to this, but I think this needs some polish.

A really zoomed in screenshot:
image


I am having a hard time seeing that the middle tab is the current page now, as it has lost the differing background.

I intentionally didnt offset the left edge, so that all the icons would line up. (I dont know if that is a good idea or not, but it looked better to me)  
I think the background should only apply to the buttons, not the surrounding div, the colour change looks a bit weird spanning outside (especially with the sharp corners)  
I think that colour is too light, `#252525` looks a bit better to me

arikorn:
@Julusian wrote:
>I am having a hard time seeing that the middle tab is the current page now, as it has lost the differing background.

Fixed.

>I intentionally didnt offset the left edge, so that all the icons would line up. 

Hmm... well from my POV, indenting helps make it clear that these are all subcategories, so as long as you're on the fence, I'll leave it this way. Specifically: part of the purpose is to make it clear (a) that the group is open and (b) where it ends. The color change by itself is a bit subtle (see below); the indentation really helps.

> I think the background should only apply to the buttons,...especially with the sharp corners

Done. FWIW, I didn't change the corner rounding, but put in a box-shadow, which blurs the boundaries. Without that, the rounded corners just look too stark compared to the rest of the list, in which the rounded corners are barely visible.

> I think that colour is too light, #252525 looks a bit better to me

Done, but... it took me a moment of head-scratching to realize: I have the brightness/contrast of my screen turned down quite a bit (It's a bright screen and I usually leave it at 40-50% or so to reduce eye-strain.) With my low-contrast screen, #252525 is barely distinguishable from the rest. When I doubled the brightness, your point "suddenly" made sense!  Personally, I still favor 333. How about we meet halfway at #2a2a2a? If you still don't like it, just change `$group-background-color:` on line 398

Julusian:
> Done, but... it took me a moment of head-scratching to realize: I have the brightness/contrast of my screen turned down quite a bi

With my laptop screen on minimum brightness (I usually have it at about 10-20%) and with all the room light on, #252525 is still very clear to me :laughing:   
But fine #2a2a2a it is

arikorn:
Thanks Julian. That's too funny that we're competing for who can "see" on the dimmest screen!!

PR #3916, opened 2026-01-20
feat: consolidate surface settings

arikorn:
**Consolidate settings and integrations with the configured-surfaces page**

Previously, the information needed to diagnose surface connection problems was spread across two pages: configured surfaces and settings/surfaces. In 4.3, the with the introduction of the Integrations page, information is spread across three pages. This PR rearranges and consolidates all of this information into a single page, taking advantage of the blank default right-panel on the configured-surfaces page.

(NB: This PR is the continuation of a conversation in PR #3703 that we agreed to defer until after that one was merged, so a purpose of this PR is to give a concrete visualization of my ideas discussed there.)

There's quite a bit of detail here. It may be best to start by watching the video, below.

Features:
- The layout is similar to the Connections page:
- When nothing is selected the right-panel shows the settings and instance table -- making it easy to disable/enable an integration or setting and to confirm their state.
- Clicking on any configured surface/group brings up the edit panel as usual, but with the "X" close-icon visible.
- Clicking on the selected surface or the "X" icon closes the edit panel/re-reveals the settings
- Clicking on an item in the instances table (right panel) brings up the instance configuration in that panel (with the "X" close-icon, of course).
- Click on "Add Integration" above the instances table to add a new instance directly from this page.
-  When screen is narrow, an "open settings panel" button appears.

https://github.com/user-attachments/assets/e9e5eea5-e895-4b52-9e81-220467615aa7

General Notes:
- For the most part this PR repurposes existing code, with some changes to visibility, rather than adding new Components, so although it affects many files, the changes to most are minor.
- The current implementation allows the Configured Surfaces page to coexist with the Integrations page, but that page is no longer necessary. Since the Integrations page has never been "released in the wild", I would recommend replacing it with this PR (I can add that to the code.)  Note that removing the Integrations page as a separate entity could simplify the conditional coding included here.
- Likewise, the settings/surfaces page would not be necessary but one might want to preserve it for historical reasons (though note that most of its functionality has already been move or removed in 4.3 beta).
- (The inspector in Chrome insists I use `thead`/`tbody` to tables or else it emits dozens of DOM errors.)

Formatting notes (all of which should be taken as suggestions):
- Put the integrations table in a table-looking box, make the instructions more consise, fiddle with spacing.
- The original instructions for integrations are in the hover-help.
- Rearrange settings and improve formatting slightly.
- In particular "Enable Elgato software Plugin" is now grouped with the instance table, since it serves a similar function from the user POV (enabling a class of inputs).
- add "close" and "help" icons for the panel headers, visually formatted for enhanced visibility.
- add context-help icons to the table-headers inside the right panel.

_**The best way to think about the benefits of this PR is in terms of workflows**_:
A. Adding new integrations will always be rare relative to configuring/diagnosing surfaces, so for the most part the "root" right-panel is giving you instant enable/disabled status, as well as showing when it's time to update an instance all in the one page. Consider when a surface isn't showing up: is it disabled? is the integration-instance loaded? is the integration disabled? does it need to be updated? All of this information is now immediately seen in one page.

B. Here's how I envision a typical add-integration workflow: 1. plug in your surface 2. nothing shows up in the left panel. 3. look to the right, is the module enabled? 4. oh, it's not installed yet! click on "Add Surface Integration" and add it. 5. now everything's working and you didn't have to find or visit other pages: all the information you need is right there (and taking up what used to be blank space).

 Notice that the interaction starts by looking at the left panel and is all done with the goal of getting something to show up in the left panel, i.e., it starts and ends with the configured-surfaces list (without having to flip back and forth between pages), so the new right panel fits naturally on this page.

C. An alternate add-integration workflow would be the person "knowing" that they need to add a new brand of surface and now that they've seen the "Add Surface Integrations" button (and list of installed integrations) every time they were on the surfaces page, they know exactly where to go.. (so maybe an extra click or maybe saves four clicks if they start at the settings page and then have to go back to the integrations page...)

The video illustrates these workflows -- in the first part of the video my surface didn't connect because the module was missing, so I add the module. In the last part the module was disabled, which is instantly apparent. The middle part of the video shows how the layout changes with screen-width changes (and appearance of the "open settings" button).

Commit notes: 
 The PR is split into several commits because some of the changes here can be applied more generally. The commits are:
1. The main change
2. Fixing some other pages affected by a CSS improvement.
3. Add the "show settings" button, which could be ported to the Connections page, for example
4. Change "Cancel" to "Done" when surface instance properties are not dirty. This is alot less confusing to see.
5. Add option hover-help via a CircleQuestion icon in user-config headers (table headers, not the panel)
6. Add the enhanced "Close" and "Help" buttons for the top-right of panels.
7. Fix a formatting bug introduced somewhere in 4.3 beta -- the bug does not originate from this PR.

Comment:  All settings pages would probably benefit from rewriting as Containers with rows & columns instead of tables...

## Release Notes

* **New Features**
  * Two-panel responsive layout for managing surfaces and integrations
  * Contextual help buttons with user-guide links throughout settings
  * Dynamic form labels that update based on content state

* **Navigation & Organization**
  * Surface Integrations consolidated under Configured Surfaces (moved from Settings menu)
  * Streamlined routes for surface and integration management
  * Improved sidebar navigation structure

* **UI/UX Improvements**
  * Consistent close button styling across panels
  * Enhanced accessibility with improved tooltips and titles
  * Better mobile responsiveness and layout adaptation
  * PIN Lockout settings UI reorganization

* **Tests & Documentation**
  * Added test infrastructure for layout components
  * Updated user guide for new surface configuration workflow

### DIFF
--- a/webui/src/Surfaces/Instances/SurfaceInstanceList/SurfaceInstanceList.tsx
+++ b/webui/src/Surfaces/Instances/SurfaceInstanceList/SurfaceInstanceList.tsx
@@ -86,34 +86,36 @@ export const SurfaceInstancesList = observer(function SurfaceInstancesList({
 
 	return (
 		<div className="connections-list-container flex-column-layout">
-			<div className="connections-list-header fixed-header d-flex align-items-center">
+			<div className="connections-list-header fixed-header d-flex flex-column">
 				<MissingVersionsWarning moduleType={ModuleInstanceType.Surface} instances={surfaceInstances.instances} />
 
 				<GenericConfirmModal ref={confirmModalRef} />
 
-				<CButtonGroup className="connection-group-actions m-1 me-auto">
-					<CButton
-						color="primary"
-						size="sm"
-						onClick={() => void navigate({ to: '/surfaces/configured/integrations/add' })}
-					>
-						<FontAwesomeIcon icon={faPlug} className="me-1" />
-						Add Surface Integration
-					</CButton>
-					<CreateCollectionButton />
-				</CButtonGroup>
-				<ContextHelpButton
-					size="xl"
-					tooltip={
-						<>
-							<p>
-								Surface integrations are like connections but for input surfaces: they provide the ability to use
-								different hardware or virtual surfaces to trigger buttons in Companion.
-							</p>
-							<p>Click on any row to configure the integration.</p>
-						</>
-					}
-				/>
+				<div className="d-flex align-items-center">
+					<CButtonGroup className="connection-group-actions m-1 me-auto">
+						<CButton
+							color="primary"
+							size="sm"
+							onClick={() => void navigate({ to: '/surfaces/configured/integrations/add' })}
+						>
+							<FontAwesomeIcon icon={faPlug} className="me-1" />
+							Add Surface Integration
+						</CButton>
+						<CreateCollectionButton />
+					</CButtonGroup>
+					<ContextHelpButton
+						size="xl"
+						tooltip={
+							<>
+								<p>
+									Surface integrations are like connections but for input surfaces: they provide the ability to use
+									different hardware or virtual surfaces to trigger buttons in Companion.
+								</p>
+								<p>Click on any row to configure the integration.</p>
+							</>
+						}
+					/>
+				</div>
 			</div>
 
 			<div className="connections-list-table-container scrollable-content mt-2">


arikorn:
(Note: I'll wait on comments/approval before dancing with the Rabbit.)

Julusian:
I can understand some of the thoughts behind this, but I'm still not sure about it.

I am a little worried about this page becoming a bit of an information overload, with multiple tables and settings all thrown in on the default state.  

As I have said before, I am not keen on the way the right panel switches between a few different pages, often switching between themselves.  
The only other page which has more than 2 views for the right pane is the button grid.

A lot of the arguments for including this in the right pane also apply to the 'remote surfaces' page. (I actually was debugging something the other day and was constantly switching between the configured and remote pages. ok, I actually opened them in separate windows side by side after a while)


I think what needs to happen here is for me to chase the others to get a third (or more) opinion, and see where this stands

---

> showing when it's time to update an instance all in the one page

side thought; perhaps we need a new 'dashboard' page to provide an overall system health/status page. It could list some things instances which are out of date, as well as a summary of statuses etc. Similar to what Buttons, truenas, opnsense and presumably more have. maybe could be the first place we try user customisable ui layouts..
Not at all related to this PR, just thought of it while reading this.

arikorn:
Thanks Julian, I really appreciate your openness to this idea, especially since you don't fully agree! I also totally understand and am even concerned myself about information-overload, but I think after just a couple of uses the end-user would find it very comfortable and comforting to see everything at a glance.: For the most part you don't have to read anything, just look for "red flags."

Regarding nesting, the behavior seems to me to be essentially the same as the Connections page. As far as I can tell there's only one level of nesting with wide windows: when the user presses "Add Surface Integration" the modules list shows up and is then replaced by the "Edit Surface Integration" panel-- same as with the Connections page after you add an instance. When the user taps out of that panel they're back at the top level. Maybe with narrow windows it becomes one more level.

...but I want to emphasize that this PR is not an either-or deal: the PR is designed to coexist with your current Integrations page, so if there's benefit to both ways of presenting the information, we could certainly keep both, for example: if a person needs to do intensive instance-maintenance, they can turn to the dedicated page.

Another compromise option would be to have the "Add Integration" button take the user to the Integrations page instead of opening in the same panel. Not recommending it, just noting the possibility.

Another thought: settings could be accordioned. Again, I'm not recommending, but it might help reduce information overload.

As for the  'remote surfaces'  page: see -- I did restrain myself in this PR by not adding that page too!!! 😀, but yeah, adding it would probably want a shift to tabs, more like the buttons page. Needless to say, I'm not recommending, just thinking out loud.. 

P.S. Your system-status dashboard side-thought sounds like a great idea -- maybe even as the home page instead of Connections?

webui/src/routes/_app/surfaces/configured/index.tsx
```diff
@@ -1,27 +1,114 @@
...
+								<tbody>
+									<UserConfigSwitchRow
+										userConfig={userConfigProps}
+										label="Enable Elgato software Plugin support"
+										field="elgato_plugin_enable"
+									/>
+								</tbody>
+							</table>
+						</b>
+					)}
```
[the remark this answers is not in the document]
arikorn: The bolding is intentional, to match the Integrations table, to which it is visually appended.
That said, the bolding could probably be moved inside tbody, if that's clearer.
No need for a <div> the fragment is already a single HTML element.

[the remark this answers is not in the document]
arikorn: moved. <b>

webui/src/routes/_app/surfaces/configured/index.tsx
```diff
@@ -1,27 +1,114 @@
...
+					</table>
+				)}
+			</div>
+		</>
+	)
+}
+
+function SettingsPanelTitleBar() {
+	const setPanelVisibility = useConfiguredSurfaceContext()
+	// note that the close button is currently hidden
```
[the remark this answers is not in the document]
arikorn: done.

webui/src/Surfaces/Instances/SurfaceInstanceList/SurfaceInstanceList.tsx
```diff
@@ -84,24 +90,31 @@ export const SurfaceInstancesList = observer(function SurfaceInstancesList({
...
+				<div className="connection-group-actions mb-2 mx-1">
 					<CButtonGroup>
 						<CButton
 							color="primary"
 							size="sm"
-							className="d-xl-none"
-							onClick={() => void navigate({ to: '/surfaces/integrations/add' })}
+							className={toDir ? 'undefined' : 'd-xl-none'}
+							onClick={() => void navigate({ to: `/surfaces/${toDir ?? 'integrations'}/add` })}
 						>
```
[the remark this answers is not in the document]
arikorn: done.

Julusian:
> Regarding nesting, the behavior seems to me to be essentially the same as the Connections page. As far as I can tell there's only one level of nesting with wide windows: when the user presses "Add Surface Integration" the modules list shows up and is then replaced by the "Edit Surface Integration" panel-- same as with the Connections page after you add an instance.

On the connections page there is the default 'add' page, and the 'edit config' page.  
In this, there is the default 'settings' page, the 'add' page and the 'edit config' page. So there is an extra one. (and perhaps more if the remote+discover should move into here too)

> ...but I want to emphasize that this PR is not an either-or deal: the PR is designed to coexist with your current Integrations page, so if there's benefit to both ways of presenting the information, we could certainly keep both, for example: if a person needs to do intensive instance-maintenance, they can turn to the dedicated page.

I don't think both is a good idea. It will be confusing to have the same thing in multiple places, and I couldn't justify why it should be in two slightly different forms.  
Tbh I doubt users would use the one that makes the most sense to their workflow, but instead use the one they feel more comfortable with (ie, which one they found first). (Control panel on windows has taught me that)

----

Yeah I wouldn't try and polish this at all yet, I haven't looked at the code at all or picked at the ui. Still at the stage of thinking about the bigger picture.

arikorn:
@coderabbitai probably best to hold off on reviews for now

Julusian:
OK, so I need to try and give this a review and get this merged within the next couple of weeks..

I feel like the darker close/help icons in the 2nd panel headings feel a bit too bold. Perhaps it could be a bit lighter so that it still stands out but not quite as much?  
Also it would be good for the other panels that use these panel headers to have the same styling. (that could be a separate pr to change just those?)

I feel like these tables need some indicator that the rows can be clicked on to do something. It was maybe subtle before, but elsewhere the right panel usually hints to click something with its placeholder. (although connections are in the same boat so maybe this is a pre-existing issue..)

The 'old' integrations page should be removed, there should be one way to access these ui panels

I'm not keen on the 'settings' parts being split above and below the table, I think it would be better to have them all below.  
Part of me is wondering if they should be made a bit more subtle in some collapsing panel or something? not sure on that though

I'm not a fan of where 'Enable Elgato software Plugin support' has been placed. but I can see why you put it there.
I don't know if it should be there, or if it should be pushed to the services settings.  
Tbh I am considering whether it is time to get rid that special websocket for that plugin, satellite isnt capable of handling everything it can do yet, but it could be expanded.

I havent looked at the code yet, it might be best to remove the old page and do some tidying before I do that (and I'm not expecting any surprises as this is mostly just shuffling existing code)

webui/src/routes/_app/surfaces/configured/integration.$instanceId.tsx
```diff
@@ -0,0 +1,26 @@
...
+	const navigate = useNavigate()
+
+	// Ensure the selected instance is valid
+	useComputed(() => {
+		if (!surfaceInstances.instances.has(instanceId)) {
+			void navigate({ to: '/surfaces/configured' })
+		}
+	}, [navigate, surfaceInstances, instanceId])
+
+	return <SurfaceInstanceEditPanel key={instanceId} instanceId={instanceId} />
```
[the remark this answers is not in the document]
arikorn: fixed a similar but different way

webui/src/Surfaces/ConfiguredSurfacesPage.tsx
```diff
@@ -58,11 +64,11 @@ export const ConfiguredSurfacesPage = observer(function ConfiguredSurfacesPage()
...
 			}
 		},
-		[navigate]
+		[navigate, selectedItemId]
 	)
 
-	const showPrimaryPanel = !selectedItemId
-	const showSecondaryPanel = !!selectedItemId
+	const showPrimaryPanel = !selectedItemId && !showSettings
+	const showSecondaryPanel = !!selectedItemId || showSettings
```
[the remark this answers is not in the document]
arikorn: This currently works correctly or I don't understand the comment. In narrow  mode the secondary panel should not display on first load.

arikorn:
>I feel like the darker close/help icons in the 2nd panel headings feel a bit too bold. Perhaps it could be a bit lighter 

Done and not it's now a variable in _layout.scss, with explanatory comments, so should be easy to adjust further.

> Also it would be good for the other panels that use these panel headers to have the same styling. (that could be a separate pr to change just those?)

Agreed on both points! (I.e. good to do, but after this PR)

> I feel like these tables need some indicator that the rows can be clicked on to do something. ... maybe this is a pre-existing issue..)

I've added "advice" to that effect at the top, in the hover-help, and in the tooltips when hovering over a row.
I'm open to any further suggestions as well and also note that if you move the mouse over the table-rows, it changes to the "navlink" hand and the row "lights up"

> The 'old' integrations page should be removed, there should be one way to access these ui panels

Done. There may be some now-unneeded code here and there but it shouldn't be much and doesn't interfere -- if anything it makes the code more resilient to relocations...

> I'm not a fan of where 'Enable Elgato software Plugin support' ...Tbh I am considering whether it is time to get rid ...

I will await your further instructions then. (As you've implied, I believe the current placement will make sense to the end-user since Elgato integration is just another, well... integration! Even if it's completely different under the hood.) 

> I'm not keen on the 'settings' parts being split above and below the table, I think it would be better to have them all below.

Done.

> Part of me is wondering if they should be made a bit more subtle in some collapsing panel or something? not sure on that though

 I've left it alone for now. It does raise another issue: 
- should we remove the page from "Settings" as it too is now redundant? Maybe just link it to here?  FWIW, when I get to import/export settings, my intention is to have the surface settings under surfaces anyway, i.e., removing the surface settings page will not harm that flow.

- Another question: the code for the main page is now in the routes folder-tree (_app/surfaces/configures/index.tsx). Are you OK with that or would you prefer it moved to the Surfaces/ folder?

@coderabbitai resume

Julusian:
> Agreed on both points! (I.e. good to do, but after this PR)

could be in parallel to this pr, it will be a quicker one to merge, and this one is probably going to have a few more rounds (I havent yet looked at any code)

> should we remove the page from "Settings" as it too is now redundant?

yes I think so. having a link from somewhere in the settings to here would be good though, so that users know it is possible. I'm not sure where

webui/src/Surfaces/ConfiguredSurfacesPage.tsx
```diff
@@ -45,9 +46,14 @@ export const ConfiguredSurfacesPage = observer(function ConfiguredSurfacesPage()
 		addGroupModalRef.current?.show()
 	}, [])
 
+	const [showSettings, setShowSettings] = useState(false)
```
Julusian: elsewhere we are using a route for this settings view (for the add pages), would probably be good to be consistent

webui/src/Surfaces/Instances/SurfaceInstanceEdit/SurfaceInstanceEditPanel.tsx
```diff
@@ -67,10 +67,14 @@ function useInstanceEditPanelService(
 ): InstanceEditPanelService<ClientSurfaceInstanceConfig> {
 	const { surfaceInstances } = useContext(RootAppStoreContext)
 
-	const navigate = useNavigate({ from: `/surfaces/integrations/$instanceId` })
+	// from: is only needed to resolve relative paths (and using string var caused problems)...
+	const navigate = useNavigate() //{ from: `/surfaces/configured/integrations/$instanceId` }
+	const { pathname } = useLocation()
+	const homepage = pathname.split('/', 3).join('/') // first two elements of the path ex: /surfaces/configured
```
Julusian: I don't like these, they feel a little brittle/fragile, and lose out on the type safety of the router. 
as there is only one use of these components they should be fine to replace with concrete paths.

If we need something like this in the future, I would do something with injecting via props, to keep it the concern of the parent page

webui/src/Surfaces/Instances/SurfaceInstanceList/SurfaceInstanceList.tsx
Julusian, line 1: same to various paths in here too

webui/src/Surfaces/Instances/AddSurfaceInstancePanel.tsx
Julusian, line 1: and here which is doing a bunch of logic to handle this

webui/src/Layout/PanelHeaderIcons.tsx
```diff
@@ -0,0 +1,70 @@
...
+ Caller can supply a link to the user guide or a custom function.
+ The default size is intended for panel headers.
+*/
+export function ContextHelpButton({ hoverText, userGuide, size = '2xl' }: ContextHelpButtonProps): React.JSX.Element {
+	// note: an alternative way to do the following would be a default function using
+	//  router.buildLocation() and window.open() to navigate to the link in a new window...
+	const userGuideProps =
+		typeof userGuide === 'string'
+			? {
+					to: userGuide,
```
Julusian: this will need to use the `makeAbsolutePath` helper so that it works correctly when under a subdirectory

If you run vite with `--base=something` then you can test this locally (I think, I've not tried it in a while. my main deployment uses this subpath stuff just so that I test it)

webui/src/Surfaces/ConfiguredSurfacesPage.tsx
arikorn: @Julusian: `showSettings` is used to show the settings-panel (the default right panel) in narrow view, similar to what you do for  configured-surface settings in this file.  As best as I can tell the route determines what goes in `<Outlet>`, but in this case, we're already on `/surfaces/configured` so re-routing to  `/surfaces/configured` (or any route that just reroutes to that page) appears to be a no-op. Just to be clear: in narrow mode we need to replace the default left panel with the right panel on button click and remove it when clicking the close button, which is in the `<Outlet>`-loaded page, and not here. (I could add `reloadDocument: true` to the close-button there, but that's pretty nuclear, definitely not pretty). And of course, when the page is wide-enough, we just want to see both panels. 

  So I'm stuck on the way I'm currently doing it -- how would you do this with routes?

arikorn: Hmmm...  I guess this, 
~~~ts
	const routeMatch = matchRoute({ to: '/surfaces/configured/$itemId' })
	const selectedItemId = routeMatch ? routeMatch.itemId : null
~~~
maybe with a special token could be used for the close button? Seems like it would require a double-reroute (once with token then without so that the user can show the right-panel again) I'm pretty sure that we'd still need this `useState()` to show that default right-panel so maybe not much gained? (Another way to say this is: you're finessing a bit here with how you handle showing the configured-surface editor in narrow-mode, but it seems harder to do that with the new default right-panel.)

OK. Too much musing for now...

Julusian: Give http://localhost:5173/connections/add a look.
That route resolves to the same as http://localhost:5173/connections when in 2 panel, and when in 1 panel, it changes the content

arikorn: Ok, I see that route now (in both senses of the word!). Thanks.
While a variation of that would probably work here -- it's a bit trickier here because we have a second level to deal with and I'm concerned about some edge-cases -- I've been reading a bit more of the Tanstack docs and am wondering if a search param might be more appropriate (i.e. the button navigates to `search: { showingSettings: true }` and the close button resets it; the children, if any, inherit it). At the very least it would avoid requiring a duplicate route, but I _think_ it may also handle the nesting better.... Thoughts or objections before I naively start down that path?

webui/src/Layout/Sidebar.tsx
```diff
@@ -307,15 +302,9 @@ export const MySidebar = memo(function MySidebar() {
 					<SidebarMenuItem
 						name="Configured"
 						icon={null}
-						notifications={SurfacesConfiguredTabNotifyIcon}
+						notifications={SurfacesTabNotifyIcon}
 						path="/surfaces/configured"
```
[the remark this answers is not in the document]
arikorn: this is pre-existing behavior. FWIW, the "parent" just redirects to the same place as the "child"

webui/src/Layout/PanelIcons.tsx
```diff
@@ -0,0 +1,70 @@
...
+	return (
+		<CButton
+			color="dark"
+			className={`float_right${visibilityClass ? ' ' + visibilityClass : ''} p-1 ms-2 panel-close-button`}
+			onClick={closeFn}
+			title="Close"
+		>
+			{/* The inline styling here is to make the icon square */}
+			<FontAwesomeIcon icon={faTimes} size="lg" style={{ marginRight: '1px' }} />
+		</CButton>
```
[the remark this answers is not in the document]
arikorn: done

webui/src/routes/_app/surfaces/configured/index.tsx
```diff
@@ -1,27 +1,121 @@
...
+function SurfaceSettingsPanel() {
+	const userConfigProps = useUserConfigProps()
+
+	return (
+		<>
+			<SettingsPanelTitleBar />
+			<p style={{ marginBottom: '0em', fontStyle: 'italic', padding: '0.5em 2em' }}>
+				To configure a surface integration, select an item from the table below. For a specific surface, select it in
+				the Configured Surfaces table to the left.
+			</p>
```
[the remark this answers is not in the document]
arikorn: made it conditional

webui/src/UserConfig/Components/UserConfigHeadingRow.tsx
```diff
@@ -1,14 +1,29 @@
...
-				<th colSpan={3}>{label}</th>
+				<th colSpan={3}>
+					<span className="d-flex justify-content-start">
+						{label}
+						{tooltip && (
+							<span className="ms-auto px-2">
+								<InlineHelp help={tooltip}>
+									<FontAwesomeIcon icon={faQuestionCircle} />
+								</InlineHelp>
+							</span>
```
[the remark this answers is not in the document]
arikorn: done

webui/src/Surfaces/Instances/SurfaceInstanceEdit/SurfaceInstanceEditPanel.tsx
arikorn: removed.

webui/src/Surfaces/Instances/SurfaceInstanceList/SurfaceInstanceList.tsx
arikorn: removed.

webui/src/Surfaces/Instances/AddSurfaceInstancePanel.tsx
arikorn: removed, assuming I've parsed your pronouns correctly!

webui/src/Surfaces/ConfiguredSurfacesPage.tsx
arikorn: My latest commit (c918139) implements my search-term solution. There are still two edge cases whose behavior "makes sense" or not depending on preferences: 
1. if you go into a sub-panel and _then_ make the window narrow, closing the panel will return you to the configured surfaces panel. I have no reason to favor where it returns to in this case, so that's the current behavior. It's easy enough to click the "show settings" button while narrow to get the "normal" narrow behavior. Conversely, 
2. if you start with a narrow window, click "show settings", and then make the window wide, the "show settings" state remains active, so that the next time you make the window narrow, it will flip to the the settings panel. (this behavior is the same as in the connections page; the first edge case doesn't exist for the connections page). Again, one click and your back on the configured surfaces page, so neither case seems like a big deal.

Let me know if you prefer different behavior.

arikorn:
@Julusian:

- Pushed code for handling window-width state, for your review.
- Still outstanding: your last code comment, removing the settings page, and perhaps some polish

webui/src/Surfaces/ConfiguredSurfacesPage.tsx
arikorn: I "fixed" both issue mentioned in my previous comment in the latest commit (adba2a6). Fixed is in quotes because the first issue is really a coin-toss in my mind, but now the close-button action consistent in narrow windows whether-or-not the window was narrow at the outset.

Note that this solution (the second item in the previous comment) can be ported over to the connections component too. I'm not sure if it requires the method I'm using (search rather than a different route), but I suspect that using search params avoids a render cycle.

webui/src/Layout/PanelHeaderIcons.tsx
arikorn: OK, this has me very confused. No problem implementing it (and it works), but then I thought "shouldn't I do it for the help menu too? and how/why does the sidebar work?" On the built/packaged app, the sidebar/help-menu correctly take me to the user-guide _BUT_ 
1. if I run the webui with `yarn dev --base=/testme/` then the links fail by going to a message that offers to redirect me to the correct link.  _BUT_
2. if I wrap /user-guide/ in `makeAbsolutePath`, it goes to `/testme/testme/user-guide` which fails and goes back to the connections page...

Since  `makeAbsolutePath` is working correctly, i.e. it returns `/testme/user-guide`, I'm guessing that just /user-guide/ gets some special handling, but I couldn't see where. Also, I didn't see a Tanstack route for it, which might be a good idea (and just have it redirect to href, which would be more correct when using Navlink with `to`, I suspect??).

Thoughts? Advice? (before I push code that will make your code-comment "outdated")

Julusian: huh. maybe the router is doing the fixup to these for us. it does seem to be for a few elsewhere

arikorn: Well, it's another mystery so far...  I got as far as determining that if you use `<CNavLink to... as=Link>` it will add the basepath but if use use `<a...>` or `<CNavLink href...>` it will not.  It appears to be added by webui/src/index.tsx upon creating the `RouterProvider` component (which is OK but not entirely conventional, apparently it's more conventional to use the basepath prop in createRouter...), which explains the "to" behavior, but it's still odd:

- In the DOM, the URL is hardcoded to include the basepath, so the sidebar link (which uses `<CNavLink to="/user-guide/"...>` ) shows `<a href="/testme/user-guide" target="_blank" class="nav-link">...</a>`, yet when I click on it, the link goes to `/user-guide` not /testme/user-guide.
- but if I set `...path={makeAbsolutePath("/user-guide/")...` (which sets `to` in the `<CNavLink>`), then the target is `<a href="/testme/testme/user-guide" target="_blank" class="nav-link">...</a>`, and the link goes there w/o stripping anything.

I have a feeling the weird behavior is related to the fact that Tanstack isn't really serving the user-guide. Maybe something to investigate later?

arikorn:
All of the current review items should have been addressed now. I deleted the settings pages but left the card and sidebar entries, linked to the configured-surfaces page. That said, those should probably be removed in the next Companion version.

Outstanding items:
- separate PR for panel-heading close/help buttons
- whether to move `SurfaceSettingsPanel` code to its own file. I've left it alone for now for ease of comparison and because I'm not entirely convinced it _should_ be moved.
- whether to consolidate remote surfaces into the configured-surfaces page as well. If yes, I suspect the best way would be to add two buttons to the left-panel: "add remote", and "discover remote", though currently I don't understand the relationship between remote surfaces and configured surfaces. Perhaps I need an actual remote surface for it to show up in configured surfaces? Maybe it's all best left alone for now?

arikorn:
@coderabbitai please note that using href for external links with Link components is perfectly valid Tanstack usage.

arikorn:
@coderabbitai resume

webui/src/routes/_app/surfaces/configured/$itemId.tsx
```diff
@@ -36,22 +34,17 @@ const RouteComponent = observer(function RouteComponent() {
...
-		if (itemId && !itemInfo) {
-			void navigate({ to: `/surfaces/configured` })
-		}
-	}, [navigate, itemId, itemInfo])
-
 	if (!itemInfo) {
-		return null
+		// Redirect if item not found
+		// condition was: itemId && !itemInfo but if itemId is missing, we wouldn't reach this route: it would go to index.tsx
+		return <Navigate to="/surfaces/configured" replace />
```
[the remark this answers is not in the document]
arikorn: I am leery of the whole "deep-linking" concern at the moment. OTOH, this one is pretty simple to fix, so sure.

webui/src/Surfaces/KnownSurfacesTable.tsx
```diff
@@ -171,12 +171,13 @@ const ManualGroupRow = observer(function ManualGroupRow({
...
 	)
 
+	const groupName = group.displayName || 'Surface Group'
 	return (
 		<>
 			<div className={classNames('grid-row', { 'grid-row-selected': isGroupSelected })} onClick={handleGroupClick}>
 				<div className="grid-cell">#{group.index}</div>
-				<div className="grid-cell">
-					<b>{group.displayName || 'Surface Group'}</b>
+				<div className="grid-cell" title={`Click to edit ${groupName} group settings.`}>
```
Julusian:
```suggestion
				<div className="grid-cell" title={`Click to edit ${groupName} settings.`}>
```

webui/src/Surfaces/KnownSurfacesTable.tsx
```diff
@@ -284,9 +292,7 @@ const SurfaceRow = observer(function SurfaceRow({
 					)}
 				</div>
 				<div className="surface-id-row">
-					<span className="surface-id" title={surface.id}>
-						{surface.id}
-					</span>
+					<span className="surface-id">{surface.id}</span>
```
Julusian, line 297: I think I added this as a title because the id gets `...` when its too long to fit. but maybe this is fine now with the other title youve added?

webui/src/routes/_app/surfaces/configured/integrations.$instanceId.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { createFileRoute, Navigate } from '@tanstack/react-router'
+import { RootAppStoreContext } from '~/Stores/RootAppStore'
+import { SurfaceInstanceEditPanel } from '~/Surfaces/Instances/SurfaceInstanceEdit/SurfaceInstanceEditPanel'
+import { observer } from 'mobx-react-lite'
+import { useSurfaceInstancesSubscription } from '~/Hooks/useSurfaceInstancesSubscription'
+
+const ModuleConfigComponent = observer(function ModuleConfigComponent() {
+	const { instanceId } = Route.useParams()
+	const { surfaceInstances } = useContext(RootAppStoreContext)
+	const dataReady = useSurfaceInstancesSubscription(surfaceInstances)
```
Julusian: I dont think this is a good idea.
This will be updating the surfaceInstances store with the subscription as well as this being done at the root level. Which will either result in the store emitting changed events multiple times and causing re-renders, or spending more time diffing the changes to figure out whether to trigger a re-render.

webui/src/routes/_app/surfaces/configured/$itemId.tsx
```diff
@@ -1,16 +1,16 @@
...
 import { observer } from 'mobx-react-lite'
 import { RootAppStoreContext } from '~/Stores/RootAppStore.js'
+import { useSurfacesSubscription } from '~/Hooks/useSurfacesSubscription'
 
 const RouteComponent = observer(function RouteComponent() {
 	const { surfaces } = useContext(RootAppStoreContext)
 	const { itemId } = Route.useParams()
-
-	const navigate = Route.useNavigate()
+	const dataReady = useSurfacesSubscription(surfaces)
```
Julusian: same comment as elsewhere

Julusian:
I'm not keen on the `Surface Integrations (Plugins)` heading. I think it just feels a bit much to have that heading below the panel heading with most of the words the same in both (we also don't refer to them as plugins anywhere else). And that heading is causing a border to be added to the content below which I dont think works.

Maybe the `The following settings affect all surfaces.` line should have a proper heading above it, matching the pincode one?

I suppose generally the italic feels a bit off, as we arent doing that anywhere else

webui/src/Layout/PanelHeaderIcons.tsx
Julusian: I havent noticed any broken links in my testing with a basepath, but there could be some. but we probably should be directing them all away from the router, as they are supposed to open in a new tab and things anyway

webui/src/Surfaces/ConfiguredSurfacesPage.tsx
Julusian: hmm.. I think it looks visually cleaner to not use a search parameter, so I would prefer it to not, but that nested closing does make that harder to do correctly. a solution for that though would do the close button route based on the current width of the page though, so it would be consistent no matter which route was taken

the connections doesnt have the same nesting problem. or at least I see adding an instance as being reasonable to close the add-instance panel, so returning to the list of instances after adding makes sense to do.

Julusian:
another small bug (that could be preexisting), the right panel jumps vertically by a few pixels when it is changed from the default state. this icnludes when it changes between /configured and /configured?showSettings=true

webui/src/Surfaces/ConfiguredSurfacesPage.tsx
```diff
@@ -45,9 +46,48 @@ export const ConfiguredSurfacesPage = observer(function ConfiguredSurfacesPage()
...
+	const { showSettings: showSettingsParam } = useSearch({ from: '/_app/surfaces/configured' })
+	const handleShowSettings = useCallback(() => {
+		// note: the search term will propagate to the other sub-windows, so when the window is narrow,
+		// subwindows will return to the settings panel (since the user must have gotten to the sub-panel from there).
+		void navigate({ to: '.', search: { showSettings: true } })
+	}, [navigate])
+	const showSettings = showSettingsParam || showingSubpanel
+
+	// 3. Handle changes in panel visibility
+	const primaryPanelRef = useRef<React.ElementRef<typeof CCol>>(null)
```
Julusian: rather than fiddling with refs and checking styles directly, I would think to solve this by some form of `useMediaQuery` either with `d-xl-block` directly if that is possible, or making a reusable component for that query and some constants which have to be kept in sync with the css

then something like:
```
useEffect(() => {
	if (!is2Panel) return
	const t = setTimeout(() => {
      void navigate(....)
    }, 1000)
  
  return () => {
    clearTimeout(t)
  }
}, [is2Panel])

```
probably all wrapper up in a reusable component so that the other pages which have this special panel handling can reuse it

webui/src/Layout/PanelHeaderIcons.tsx
arikorn: While I'm still not sure how accurate my diagnosis above is, using Tanstack here does appear to be inappropriate. Technically, if I wrote `<Link to="/user-guide/"...>` it's a typescript error since that is not a valid route. We've been bypassing the type-safety and relying on Tanstack to pass on the outside link but as demonstrated above, this can result in inconsistent behavior. In PR #4024 I fixed my own code for help- and context- menus to be type-safe, and would suggest doing it for sidebar, etc, as well. In that same spirit, I've just updated the code here so it doesn't use Tanstack's `<Link>` element to link the user-guide. That and already using makeAbsolute path should resolve this comment completely.

webui/src/routes/_app/surfaces/configured/integrations.$instanceId.tsx
arikorn: done.

webui/src/routes/_app/surfaces/configured/$itemId.tsx
arikorn: done.

webui/src/Surfaces/KnownSurfacesTable.tsx
arikorn: Ah, that makes sense. Still it was pretty much hidden since it's in the "fine-print" (smallest font size). I just reformatted the title to start with the surface-id, so the most relevant information is up-front in the tooltip. Also moved the title up to cover the whole row. (It was missing the initial grid-cell.)

So, yes, it should be fine now: it was a "move" rather than a "delete" refactor.

webui/src/Surfaces/KnownSurfacesTable.tsx
arikorn: Well, this may be a naming-style issue. Many of my group names are the name of a device such as "Contour Shuttle" -- so I can swap out devices without breaking buttons or having to configure a start page for each. In this case, it's helpful to have the word "group" in the tooltip. OTOH, if you name it "My Big Group" then the extra "group" here is redundant.  My compromise for now is to test for the word group in the group name and add the word "group" conditionally.  That said, it won't suppress "group" on "Meine große Gruppe", but maybe that's ok for now -- we're not really handling i18n in the webui are we? 

I'll resolve the issue but feel free to re-open it if you object.

webui/src/Surfaces/ConfiguredSurfacesPage.tsx
arikorn: Not sure this was worth doing or not (fbda614). Perhaps it simplified the code slightly. You also get an unrelated bonus: I fixed the pre-existing 5-pixel jump that you noted in the comments. I really can't support routing the  close-button action based on current width: it's WET, breaks modularity a bit and relies and a hidden double-chain of assumptions about when the left panel is visible (i.e. (1) at what breakpoint it becomes visible and (2) the mapping of bootstrap breakpoint to pixels). Luckily, with the "fix-it" code in the parent page/route there a simple solution - always return to the "/surfaces/configured/integrations" route and let the "fix-it" change the route instantly and invisibly if necessary. Depending on the window-width there's a one-time extra render cycle when closing a panel. It's not worth worrying about. Hopefully you agree and we can close this section!!!

arikorn:
Latest look (26df9df), should address all of your "style" comments and definitely looks cleaner. (Now that this page "owns" all of these components we have much more freedom to modify them.)
[image updated for subsequent commit (7792d5e)]
image

arikorn:
@Julusian don't forget that documentation needs to be added/updated for Surface Integrations!  (67600e6 just deletes outdated settings docs.) or at least I couldn't find it...

This PR links to the surface settings page. That should probably be changed or modified once the documentation is updated for Integrations.

webui/src/Surfaces/ConfiguredSurfacesPage.tsx
arikorn: OK, done, (hopefully correctly) now that you've enlightened me offline. 
The commit creates a "single source of truth" which is used now throughout the ConfiguredSurfacesPage. (And ultimately to be used on other pages too.)

Oh wait, I suppose this could be turned into a custom hook, as you suggested, but you'd have to pass it six parameters or else a checkVisibility function, so it may not be worth doing?

arikorn: The custom hook was manageable after all: just the two routes were actually needed, the rest could be recreated in the hook function.  So all done!

I'm marking this resolved but if you disagree (i.e. I totally or even partly misunderstood you) then just reopen it or start a new thread.

webui/src/Surfaces/ConfiguredSurfacesPage.tsx
arikorn: I'm resolving this as done. If you disagree, please just re-open or start a new thread.

webui/src/Hooks/useShowSecondaryPanel.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+export function useShowSecondaryPanel({ baseRoute, secondaryRoute }: UseShowSecondaryPanel): boolean {
+	const twoPanelMode = useTwoPanelMode()
+	const matchRoute = useMatchRoute()
+	const navigate = useNavigate()
+
+	// useLocation() forces the hook to re-render whenever the URL path changes
+	const location = useLocation()
+	const prevLocation = useRef(location) // to change immediately on route change
+
```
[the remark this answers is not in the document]
arikorn: Just to be clear your complaint is strictly for "deep linking" since the situation would never occur by clicking in the UI. And the only consequence would be to shorten the time to the invisible reroute. Still:
1. the extra test in `const newLoc` is not necessary since location won't be null.
2. the fix doesn't work even if implemented exactly as shown: it still waits a second to redirect (which is fine)
3. However, the current code redirects `/surfaces/configured/integrations` but not `/surfaces/configured/integrations/` (trailing slash) both resolve to the same route. Any idea why or how to fix?

webui/src/Hooks/useLayoutMode.ts
```diff
@@ -0,0 +1,83 @@
...
+		for (const name of Object.keys(breakpointValues) as BreakpointName[]) {
+			const value = computedStyle.getPropertyValue(prefix + name)
+			if (value) {
+				breakpointValues[name] = value
+			} else if (name !== 'xs') {
+				// strict: complain if definition is missing so CI testing will flag the problem
+				throw new Error('Missing breakpoint definition for: ' + name)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
arikorn: The placement is intentional, so it only fails once. The odds of a failure or of us using non-default breakpoints is still pretty low and the failures are mainly intended for CI (so maybe there's a gentler way to do this?)

webui/src/Hooks/useShowSecondaryPanel.ts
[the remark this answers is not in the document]
arikorn: The global advice (trailingSlash: 'never',) did nothing and the rest of the advice is not worth doing. At the very worst someone who deep-links with the trailing slash will get the settings panel if the make the window narrower (assuming it wasn't already narrow) and all they have to do is close it that one time.

arikorn:
@Julusian as best as I can tell, all of your review comments have been addressed (including the pre-existing 5-pixel jump).

Outstanding issues (not part of this PR):
- User-guide documentation for Integrations (only tangentially related to this PR; I couldn't find any but when available the integrations-list help-icon can point there)
- ~~Bug: When clicking on a hash link to the user-guide it doesn't go to the `#` even though simply refreshing the user-guide page will do it. Not sure if that's fixable on our end? (Example: click on the help icon for General Surface Settings.)~~ [edit, see PR #4036]
- outside this PR: add the new help/close panel icons where appropriate
- outside this PR: add the new `useLayoutMode` hooks where appropriate.
- whether to consolidate the remote surfaces page into here. (probably not for now?)

arikorn:
Woohoo!!! Thanks Julian!

fireworks

PR #3922, opened 2026-01-22
fix confusing terminology "depress"

arikorn:
The word "depress" mean [to press down](https://www.merriam-webster.com/dictionary/depress).  (Chalk up another one for the English language!)

In Companion "button depress" is used to mean "button release" in some places (whereas "button release" is correctly used in other places), and while it may break things to change the internal `button_depress`, this commit fixes what the user sees -- especially in Trigger Events.

NB. It would be nice to get rid of terms like this and 'bank' in the codebase, but no doubt not quite worth it.

## Summary by CodeRabbit

* **Documentation**
  * Updated user guide and inline help to use "release" terminology and clarify press vs. release semantics.
* **UI**
  * Updated event and action labels to say "release" (e.g., "On any button release") and adjusted checkbox wording to "Force release even if currently pressed".
* **Chores**
  * Added a migration that remaps existing button event entries to the new "release" naming.

### DIFF
--- a/companion/lib/Controls/ControlTypes/Triggers/Trigger.ts
+++ b/companion/lib/Controls/ControlTypes/Triggers/Trigger.ts
@@ -321,7 +321,7 @@ export class ControlTrigger
 					case 'button_press':
 						eventStrings.push('On any button press')
 						break
-					case 'button_depress': // button release
+					case 'button_release':
 						eventStrings.push('On any button release')
 						break
 					case 'condition_true':
@@ -419,7 +419,7 @@ export class ControlTrigger
 				case 'button_press':
 					this.#miscEvents.setControlPress(event.id, true)
 					break
-				case 'button_depress': // button release
+				case 'button_release':
 					this.#miscEvents.setControlPress(event.id, false)
 					break
 				case 'condition_true':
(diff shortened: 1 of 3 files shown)


arikorn:
OK it's really not me this time, but I fixed it.
> companion/lib/Data/Upgrades/v10tov11.ts(12,40): error TS2307: Cannot find module '@companion-app/shared/Model/Expression.js' or its corresponding type declarations.

**_BUT_** It makes me wonder: could I add a rule to remove `button_depress` in the new v10tov11.ts? (Would that be enough or would something have to be done with imports?)

Julusian:
> BUT It makes me wonder: could I add a rule to remove button_depress in the new v10tov11.ts? (Would that be enough or would something have to be done with imports?)

Yeah that would be enough. In that file there is a function that gets run for upgrading config, and one for upgrading imports. you should be able to find triggers already being considered there, but not their events

arikorn:
Thanks Julian, if I understand the code correctly the one fix I did handles both config and imports.  Let me know if that's wrong! (It does seem to work.)

Incidentally, I couldn't figure out where the database was being saved for the dev runs and ended up just exporting the db. Where is it? It wasn't being updated in appdata/roaming/companion/v4.3 or any of the other subdirs

Julusian:
It should be in the same place `appdata/roaming/companion/v4.3 `, but maybe its in `appdata/roaming/companion-nodejs/v4.3`? I think I've noticed that mismatch occasionally, but itll be too messy to be worth adjusting now

arikorn:
There it is!  Thanks Julian. Not sure how I missed the second companion folder right in front of my eyes.

PR #3948, opened 2026-02-07
fix: improve blink docs and reduce monaco suggest-details line-spacing

arikorn:
These are really independent but the second one arose while I was fixing the first, and they're both very short.

Feel free to adjust the line spacing -- 4-lines was just way too much.

for the first, see https://github.com/bitfocus/companion/issues/674#issuecomment-3864522512 and my reply there.

Spacing: 

Note: the `min-height` is to avoid the scrollbar, because monaco seems to insist on making the pop-up window too small.

<hr/>
before:
<br/>
image

<hr/>

after:
<br/>
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Clarified parameter names shown in autocomplete so suggestions reflect the intended argument semantics.

* **Style**
  * Improved layout and spacing of the autocomplete suggestion details panel for better readability (adjusted line height, minimum height, padding, header and close-control spacing).

### DIFF
--- a/webui/src/scss/_expression.scss
+++ b/webui/src/scss/_expression.scss
@@ -76,6 +76,14 @@
 			padding-top: 0 !important;
 			padding-bottom: 0 !important;
 		}
+
+		.codicon-close {
+			height: 1em !important;
+			width: 1em !important;
+			margin-top: -1.5em;
+			margin-right: -1.2em;
+			background-color: lightgray;
+		}
 	}
 	.docs {
 		padding-top: 1em;


Julusian:
woah, I've never seen that suggestion panel before.. looks like it could do with more refinement tbh  
the red and the cross seem a bit weird..

but this is better so am happy to merge as is (I do want to do a 4.2.4 as soon as I've finished chasing a couple of bugs, should have done it days ago really)

arikorn:
Haha. Just fixed the header. Hang on...

arikorn:
OK, that should do for now. Honestly if even you didn't realize that it exists, it's probably not worth more effort than this!

I did learn a new trick for debugging ephemeral pop-ups though, so that was cool.

arikorn:
Forgot to add screenshot. I suppose a light blue BG could also work...:


image

arikorn:
Oh.. if you want to add this to 4.2.4 do you want me to split it into two PRs? (Assuming you're not adding blink to 4.2.4?)

PR #3967, opened 2026-02-16
support insecure-algorithms manifest permissions flag

phillipivan:
Combines with [this](https://github.com/bitfocus/companion-module-base/pull/189) PR for companion-module-base to address [this](https://github.com/bitfocus/companion-module-base/issues/187) issue.

`//@ts-expect-error` can be removed once PR for companion-module-base is merged and released

## Summary by CodeRabbit

* **Bug Fixes**
  * Apply legacy OpenSSL provider when requested in manifest permissions for compatibility with packages requiring insecure algorithms.
  * Improve permission determination for modules with unsupported API versions to avoid incorrect permission exposure.

* **New Features**
  * More granular filesystem read/write allowances for native addons, packaged modules, and certain network paths to prevent permission failures.
  * Automatically allow reading the host package when present to support dev-version integrations.

### DIFF
--- a/companion/lib/Instance/NodePath.ts
+++ b/companion/lib/Instance/NodePath.ts
@@ -2,7 +2,7 @@ import fs from 'fs-extra'
 import { isPackaged } from '../Resources/Util.js'
 import path from 'path'
 import { doesModuleSupportPermissionsModel } from './Connection/ApiVersions.js'
-import type { SomeModuleManifest } from '@companion-app/shared/Model/ModuleManifest.js'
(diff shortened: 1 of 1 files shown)


companion/lib/Instance/NodePath.ts
Julusian: your addition probably wants to be above this `enableInspect` check, so that it applies when that is enabled.

Probably needs this line split up a bit to allow different routes.

phillipivan: This should be addressed now, Julian.

PR #3976, opened 2026-02-21
fix module and other instances of fuzzy filtering (searching)

arikorn:
1. current `threshold` of -10,000 results in no filtering since the expected range is 0-1, resulting in many false positives.

   Based on a bit of testing, 0.3 seems very permissive yet a decent filter. 0.5 is better if you want stricter filtering.

2. Also: add a note to the "Modules are ignored" warning to better explain what it means and how to include the ignored modules.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fuzzy search scoring to make results more selective and accurate, reducing noisy/irrelevant matches.
* **Documentation**
  * Extended in-app guidance explaining how to include previously ignored modules in search results by enabling Installed, Available, and Deprecated visibility.

### DIFF
--- a/webui/src/Buttons/Presets/fuzzyMatch.ts
+++ b/webui/src/Buttons/Presets/fuzzyMatch.ts
@@ -1,6 +1,7 @@
 import Fuzzysort from 'fuzzysort'
 
-const FUZZY_THRESHOLD = -5000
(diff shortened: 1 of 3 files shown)


arikorn:
The rabbit had a good insight about the nature of the bug -- an API change from fuzzysort v2 to v3. Google says that  -10,000 is equivalent to 0.5 in the new api, so I've updated the threshold to match. That said, 0.3 is fine too, if you want to be more permissive.

PR #3938, opened 2026-02-02
feat: right-click (aka. context-) menu capability

arikorn:
This PR does two things:

1. Provide general-purpose (right-click) context-menu capability for Companion's admin interface (the first commit)
2. Provide a "worked-example" implementing a context-menu for the sidebar (the next two commits)

The context menu is implemented as a combination of a custom-hook, which maintains the state of the menu, and a React component that manages the visual look.

The worked example is done in two commits to separate the generic changes needed to support the context menu in the sidebar from the specific menu-functionality added. The former involves adding `onContextMenu` and `<ContextMenu>` to the right places. The latter required moving group-visibility-setting functions up to the parent object.

**NOTE**: This PR largely reuses the Dropdown scheme I wrote for the header Help menu in PR #3849. I've left the generic parts in Header.tsx in order to make it easier to see what's changed. If this PR is approved, then before merging into main, I would like to move the generic parts either into their own file (ActionMenu.tsx, for example) or into ContextMenu.tsx or Common.tsx.

image

<hr/>

And a video showing off the menu features:

https://github.com/user-attachments/assets/1bf755e0-2ae0-47d1-ac45-d42a65e409e9

## Summary by CodeRabbit

* **New Features**
  * Right-click context menus with fixed-position rendering and a reusable declarative action menu for help and other menus; shared menu item/separator primitives and a hook to manage context-menu state.
* **Refactor**
  * Sidebar now tracks per-group visibility, supports group expand/collapse, accordion mode, and integrates context-menu controls for group actions.
* **Style**
  * Unified help/context menu styling, adjusted dropdown sizing, spacing, and typography.

PR #3994, opened 2026-03-02
fix one more rabbit context-menu nitpick

arikorn:
- only collapse others in accordion mode if opening a group

(note this does mean that if many groups were open before starting accordion mode, closing one won't close the others. It's probably the correct behavior -- either way is an odd edge-case.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined sidebar accordion behavior. Groups now collapse only when expanding a new item in accordion mode, rather than collapsing all groups whenever accordion mode is toggled. This provides the expected accordion functionality and improves navigation experience in the sidebar.


arikorn:
this PR concludes PR #3938

PR #4004, opened 2026-03-05
feat: suppress header notifications

arikorn:
This PR adds a command-line switch to suppress showing notifications in the header. Its main use (for me) is to help take screen images, for docs, without the "EXPERIMENTAL" notification/warning, however it's possible that people who prefer not to upgrade will find it useful too.

TBH, if you don't like this, it's no big deal. I did it already, so am posting in case others might want it.

## Summary by CodeRabbit

* **New Features**
  * Added a CLI option to disable startup and runtime notifications; notifications remain enabled by default.
* **Behavior**
  * When notifications are disabled, the app will no longer produce notification updates or related messages during runtime.

### DIFF
--- a/companion/lib/UI/Update.ts
+++ b/companion/lib/UI/Update.ts
@@ -82,13 +82,17 @@ export class UIUpdate {
 				}
 
 				this.#logger.debug(`fresh update data received ${JSON.stringify(res.data)}`)
+				if (!this.#appInfo.notifications) {
+					this.#logger.debug(
+						'Notification display has been disabled by the command-line: not showing the update message.'
+					)
+					return
+				}
+
 				this.#latestUpdateData = {
 					link: res.data.link,
 					message2: undefined,
 					...res.data,
-				}
-				if (!this.#appInfo.notifications) {
-					this.#latestUpdateData.message = ''
 				}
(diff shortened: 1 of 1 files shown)


companion/lib/UI/Update.ts
```diff
@@ -87,6 +87,9 @@ export class UIUpdate {
 					message2: undefined,
 					...res.data,
 				}
+				if (!this.#appInfo.notifications) {
```
Julusian: I think it would be better to make this if block the whole `this.#latestUpdateData = {` line and probably the `this.#updateEvents.emit('info', this.#latestUpdateData)` too.

As it stands it isnt clear whether it is expected that the ui should still display a message2 or not (and the exact content of those lines isnt really defined, I don't think we ever send anything yet)

arikorn: Thanks Julian. Done.

FWIW, if we're not storing the message, we could skip the whole callback altogether unless you're using this for the tracking on companion.free?  (I see the Rabbit had the same thought but without considering the possible utility of the HTTP request.)

Julusian: yeah the request is part of our active installation counting, so should be kept

PR #4005, opened 2026-03-06
feat: move tanstack devtools toggler off the header

arikorn:
Put it in bottom right under the query button 
This is a game-changer!!! (I hope you agree.)

## Summary by CodeRabbit

* **Chores**
  * Adjusted the positioning of the developer tools panel from top-left to bottom-right for improved interface layout and visibility.


arikorn:
Just to be clear, despite the prop name, it's changing the position of the button. The panel still shows up at the bottom.

PR #4019, opened 2026-03-12
fix: change debounce library for consistency

arikorn:
This is just a suggestion in case the usage was unintentional:

`useColor` is the only file that uses `lodash debounce`, so this PR changes it to `debounce-fn` for consistency with the rest of the codebase. It should have no material effect.

(There are also two places that use `p-debounce`, but presumably that's actually necessary...)

## Summary by CodeRabbit

* **Refactor**
  * Improved color picker dependency management for better performance optimization.


Julusian:
yeah the color picker is taken from https://github.com/casesandberg/react-color and only lightly tweaked, so doesnt necessarily match the style of elsewhere.

`p-debounce` does promises better, so is used in a few places where that is needed.

PR #4021, opened 2026-03-12
feat: add date expression functions

makstech:
After creating a [module with date/time utilities](https://github.com/makstech/companion-module-generic-date/blob/main/companion/HELP.md) and based on our discussion in the [Slack thread](https://bitfocusio.slack.com/archives/CFG7HAN5N/p1769475285440269), it was suggested that the date functionality should be implemented directly in the Companion. So here it is. This is the first batch of expression functions that bring some of the functionality from my module, directly to Companion. 

## Summary

- Add `parseDate`, `dateYear`, `dateMonth`, `dateDay`, `dateHour`, `dateMinute`, `dateSecond`, and `dateWeekday` expression functions
- All functions accept an optional IANA timezone parameter (e.g. `'UTC'`, `'America/New_York'`) and default to local time when omitted
- All functions default to current time when called without arguments (e.g. `dateHour()` returns the current hour)
- Uses native `Intl.DateTimeFormat` for timezone support, so no new dependencies

## Summary by CodeRabbit

* **New Features**
  * Added expression support for date operations: parse dates (numeric Unix ms or date strings) to Unix‑ms timestamps and extract components (year, month, day, hour, minute, second, weekday) with optional IANA timezone handling.

* **Documentation**
  * Added a "Date operations" section with usage examples, return semantics (null on invalid input), and timezone guidance.

* **Editor**
  * Expanded expression completions/help to include the new date operations.

* **Tests**
  * Added tests covering parsing, component extraction, timezones, and invalid inputs.


shared-lib/lib/Expression/ExpressionFunctions.ts
```diff
@@ -269,4 +326,52 @@ export const ExpressionFunctions: Record<string, (...args: any[]) => any> = {
...
+		const d = toDate(v ?? Date.now())
+		if (!d) return NaN
+		const parts = getDateParts(d, tz)
+		return parts ? parts.second : NaN
+	},
+	dateWeekday: (v, tz) => {
+		const d = toDate(v ?? Date.now())
+		if (!d) return NaN
+		const parts = getDateParts(d, tz)
+		return parts ? parts.weekday : NaN
```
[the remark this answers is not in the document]
makstech: That rather is a feature. For example, if I want to get the current hour in other timezone, I can skip calling `unixNow()`, and just do `dateHour(null, 'America/New_York')`

docs/user-guide/4_expressions/functions.md
```diff
@@ -333,3 +333,49 @@ eg `timeDiff($(internal:time_hms), "18:00:00")` will return the seconds until `1
...
+
+eg `dateYear()` returns the current year, `dateMonth(unixNow(), 'UTC')` returns the current month in UTC.
+
+**dateHour(value?, timezone?)**, **dateMinute(value?, timezone?)**, **dateSecond(value?, timezone?)**
+
+Extract time components from a date value. When called without arguments, uses the current time.
+
+- `dateHour` returns 0–23
+- `dateMinute` returns 0–59
+- `dateSecond` returns 0–59
```
peternewman, line 368: Hopefully actually up to 60/61 for leap seconds?

docs/user-guide/4_expressions/functions.md
```diff
@@ -333,3 +333,49 @@ eg `timeDiff($(internal:time_hms), "18:00:00")` will return the seconds until `1
...
 The returned seconds can also be used within `secondsToTimestamp` to format the result as needed.
+
+### Date operations
+
+**parseDate(value?)**
+
+Parse a date value and return the Unix timestamp in milliseconds. Accepts:
+
+- Numbers: treated as Unix milliseconds
+- Strings: ISO 8601 format (e.g. `"2024-06-15T14:30:00Z"`) or other parseable date formats
```
peternewman: It would be nice to document for end-users what other formats are supported (e.g. a link to some external site for the base parsing).

E.g. I'd assume that both US and everywhere else's M/D/Y and D/M/Y wouldn't be supported due to ambiguity.

I guess the follow on there is a future extension where you can pass in a format string to use to parse it...

shared-lib/lib/__tests__/expressions-functions.test.ts
```diff
@@ -495,6 +495,101 @@ describe('functions', () => {
...
+			// Number input (Unix ms)
+			expect(ExpressionFunctions.parseDate(0)).toBe(0)
+			expect(ExpressionFunctions.parseDate(1700000000000)).toBe(1700000000000)
+
+			// String number input
+			expect(ExpressionFunctions.parseDate('1700000000000')).toBe(1700000000000)
+
+			// ISO string input
+			expect(ExpressionFunctions.parseDate('2024-01-15T12:00:00.000Z')).toBe(
+				new Date('2024-01-15T12:00:00.000Z').getTime()
```
peternewman: Confirm you're parsing all the bits of the datetime
```suggestion
			expect(ExpressionFunctions.parseDate('2024-01-15T12:13:14.321Z')).toBe(
				new Date('2024-01-15T12:13:14.321Z').getTime()
```

It would probably also be good to add some other (failing) parse attempts, e.g. US format, European, with/without timezone.

docs/user-guide/4_expressions/functions.md
```diff
@@ -333,3 +333,49 @@ eg `timeDiff($(internal:time_hms), "18:00:00")` will return the seconds until `1
 An example using a Date Time String could be `timeDiff($(internal:time_hms), "2024-07-04T20:00-04:00")` which would return the number of seconds from the current Companion time until 4th July 2024, 8pm, in the UTC-4 Timezone.
 
 The returned seconds can also be used within `secondsToTimestamp` to format the result as needed.
+
+### Date operations
```
peternewman, line 337: Personally to me, especially given we've already got lots of time functions, having date functions like dateHour seems a bit odd, and something like datetimeX would be less ambiguous.

The timeDiff docs above already refer to "date time string" too.

docs/user-guide/4_expressions/functions.md
makstech: JavaScript's `Date.getSeconds()` is spec'd as `0-59`, the JS Date API doesn't support leap seconds. So the documentation is accurate for this implementation.

docs/user-guide/4_expressions/functions.md
makstech: Good point. I'll add a note that ISO 8601 format is recommended for reliable parsing, and that ambiguous formats like `MM/DD/YYYY` or `DD/MM/YYYY` are not reliably supported since we use JavaScript's native `Date()` constructor under the hood.

shared-lib/lib/__tests__/expressions-functions.test.ts
makstech: Yes, will update the test to use distinct values for all components and add some failing format attempts

docs/user-guide/4_expressions/functions.md
makstech: The date prefix follows JavaScript's `Date` convention, where `Date` objects inherently hold both date and time components. It also keeps function names concise, `dateHour()` vs `datetimeHour()`.  
But happy to rename if the maintainers prefer `datetime`.

Julusian:
Looks good.

I am not keen on the usage of NaN here, I would prefer to keep everything in expressions to be json-safe, as at times various bits are transferred or stored in json. We don't return NaN in other functions (not intentionally), and we don't have an equivalent for isNaN, so users wont be able to handle NaN very well.  
And I'm not sure about treating null/undefined as now, I think that will be surprising

Im curious, how much more are you thinking of adding? I'm a little worried by how many functions this is adding for expressions, but I dont know how to avoid that.

makstech:
@Julusian 
On NaN, good point about JSON safety. I'll switch all error returns from NaN to null, it's JSON-safe and kinda indicates no value.

On null/undefined as "now", agreed. I'll remove that behavior and users can just use unixNow() explicitly.

On function count, after this PR, I planned only 2 more functions: `dateFormat(value, formatString, tz?)` and `dateAdd(value, amount, unit)`. That's it. Although my initial idea with module would have brought much more functionality, I think this should cover the majority of use-cases.

shared-lib/lib/Expression/ExpressionFunctions.ts
```diff
@@ -7,6 +7,63 @@ function toString(v: any): string {
...
+	if (v instanceof Date) return v
+	if (typeof v === 'number') return new Date(v)
+	if (typeof v === 'string') {
+		const num = Number(v)
+		if (!isNaN(num) && v.trim().length > 0) return new Date(num)
+		const d = new Date(v)
+		if (!isNaN(d.getTime())) return d
+	}
+	return null
+}
```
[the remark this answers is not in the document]
makstech: Should be fixed and the function much simplified

shared-lib/lib/Expression/ExpressionFunctions.ts
```diff
@@ -7,6 +7,62 @@ function toString(v: any): string {
...
+	return d && !isNaN(d.getTime()) ? d : null
+}
+
+const weekdayMap: Record<string, number> = { Sun: 0, Mon: 1, Tue: 2, Wed: 3, Thu: 4, Fri: 5, Sat: 6 }
+
+function getDateParts(
+	d: Date,
+	tz?: string
+): { year: number; month: number; day: number; hour: number; minute: number; second: number; weekday: number } | null {
+	if (!tz) {
```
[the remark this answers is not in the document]
makstech: Treating an empty string as "no timezone" is intentional, it's the same forgiving approach other functions in the codebase use for missing params (e.g. `decode`/`encode` default `enc`). Most likely scenario for '' is an empty variable, where falling back to local time is more useful than returning null.

shared-lib/lib/Expression/ExpressionFunctions.ts
```diff
@@ -7,6 +7,62 @@ function toString(v: any): string {
...
+		const wd = parts.find((p) => p.type === 'weekday')?.value
+		return {
+			year: get('year'),
+			month: get('month'),
+			day: get('day'),
+			hour: get('hour'),
+			minute: get('minute'),
+			second: get('second'),
+			weekday: wd ? (weekdayMap[wd] ?? NaN) : NaN,
+		}
```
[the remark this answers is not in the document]
makstech: `toDate()` already validates the Date before it reaches `getDateParts()`, so the parts extraction always operates on a valid Date. Your own earlier research confirmed that `en-US` with `weekday: 'short'` is guaranteed to return Sun/Mon/Tue/.... Adding this extra validation is unnecessary overhead for a case that won't occur with valid inputs.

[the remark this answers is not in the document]
makstech: But the more I look at it, the more I don't like it. So I just basically did what you suggested

arikorn:
I'm a bit late to the game but am curious: couldn't most of this be replaced by a single `dateFormat()` function (which I see you intend to add). Also, in other PRs @Julusian has suggested using `dayjs` which is already being used by Companion and has a formatting function. Any thoughts on that from either of you?

Julusian:
@makstech any thoughts on the comment from @arikorn ?

I'm happy with this using `Intl.DateTimeFormat` instead of dayjs, I dont know what future parts might want, but avoiding dependencies is always good if it doesnt add much complexity

makstech:
@arikorn @Julusian 
The component functions return numbers, which is important for math and conditionals in expressions, e.g. `dateHour(value) > 17` or `dateWeekday(value) == 0`. `dateFormat` returns strings for display purposes, so it covers a different use case. Using `dateFormat` for extraction would require users to parse strings back to numbers.

PR #4026, opened 2026-03-16
feat: develop webui and docs concurrently

arikorn:
This PR add two options for webui developers, which may help debugging new links to the user-guide from the Admin interface, for example, particularly with a BASE_URL set.

(in the webui workspace):

- `yarn dev` now forwards user-guide requests to port 4000. It assumes the docusaurus server was started independently.
- `yarn dev:docs` (or `env WEBUI_DOCS=1`): starts up docusaurus when you start up Vite and ensures that BASE_URL is shared by both.
- `yarn dev:no_docs` (or `env WEBUI_NO_DOCS=1`): reverts to the old behavior of forwarding user-guide requests to Express on the back-end. (And overrides WEBUI_DOCS.)

## Summary by CodeRabbit

* **Documentation**
  * Updated placeholder docs text to indicate the docs server isn't running and added guidance for running docs separately or alongside the web UI.

* **New Features**
  * Added dev script aliases to run the web UI with or without docs (new dev commands).
  * Added options to run a local docs server alongside the web UI and environment flags to enable/disable that behavior.

### DIFF
--- a/tools/webui-dev-docs.mts
+++ b/tools/webui-dev-docs.mts
@@ -17,7 +17,9 @@ import { resolve } from 'path'
 import { $, argv, question, usePowerShell } from 'zx'
 import { normalizeBasePath } from './webui-dev-utils.ts'
 
-usePowerShell()
+if (process.platform === 'win32') {
+	usePowerShell() // to enable powershell
+}
 
 // Parse --base arg and normalize
 const normalizedBase = normalizeBasePath(argv.base ?? '')
@@ -75,7 +77,7 @@ if (await isPortInUse(4000, docusaurusHost)) {
 				process.exit(1)
 			}
 		} else {
-			await $`lsof -ti:4000 | xargs kill -9`.quiet().nothrow()
+			await $`lsof -ti TCP:4000 -sTCP:LISTEN | xargs kill -9`.quiet().nothrow()
 			console.log('Killed process on port 4000')
 		}
 	} else {


webui/vite.config.ts
```diff
@@ -1,13 +1,17 @@
...
+import { spawn, spawnSync } from 'child_process'
 import fs from 'fs'
 import path from 'path'
 import tsconfigPaths from 'vite-tsconfig-paths'
 
-const upstreamUrl = process.env.UPSTREAM_URL || '127.0.0.1:8000'
+// note: in order to correctly work with Docusaurus we need to use localhost rather than 127.0.0.1
+//  presumably to avoid IPv4/IPv6 mismatches.
+const upstreamUrl = process.env.UPSTREAM_URL || 'localhost:8000'
+const upstreamHost = upstreamUrl.split(':')[0]
```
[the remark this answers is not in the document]
arikorn: makes more sense to honor the upstreamHost. Done.

webui/vite.config.ts
```diff
@@ -40,6 +44,50 @@ let normalizedBase = basePath
...
+						spawnSync('taskkill', ['/pid', String(proc.pid), '/f', '/t'], { shell: true })
+					} else {
+						process.kill(-proc.pid)
+					}
+					proc = null
+				}
+			}
+			server.httpServer?.on('close', cleanup)
+			process.on('beforeExit', cleanup) // catches ending Vite by typing "q"
+			process.on('exit', cleanup) // final safety net
```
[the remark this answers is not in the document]
arikorn: Well, that's not entirely correct: if the old docuserve proc hangs around after restart, we can't close it. The proc.stdin?.write('n\n') is only ever used if something went wrong in this code, so it's a fallback to prevent the console from locking. Still, the advice here is probably wise, so it's now done.

webui/vite.config.ts
```diff
@@ -40,6 +44,83 @@ let normalizedBase = basePath
...
+			docuProc = spawn('yarn', ['start', '--no-open', `--host ${upstreamHost}`], {
+				cwd: path.join(import.meta.dirname, '../docs'),
+				env,
+				stdio: ['ignore', 'pipe', 'pipe'], // stdin  piped (to answer prompts), stdout+stderr piped to add a prefix,
+				shell: true,
+				detached: process.platform !== 'win32', // needed for process.kill(-pid) on Unix
+			})
+			// answer "n" to any port-in-use prompt so it exits cleanly, just to be safe (the particular underlying problem appears to have been solved for now)
+			docuProc.stdin?.write('n\n')
+			docuProc.stdin?.end()
```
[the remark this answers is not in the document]
arikorn: drat it! that's a regression. Fixed.

Julusian:
What is the value in the `dev:nodocs` command? I can't see when I would actually use that instead of docs and just get page load errors. I added that docs html page so that it didnt return an error page, I dont think it makes sense to have to opt into getting that placeholder, and have the default return to being a 500 error page.  
I was curious, and claude gave some good  pointers on how to make the vite proxy fallback to serving the placeholder when docusuarus is not running, so I have pushed that.

When I ctrl-c the `yarn dev:webui:docs` command, docusuarus stays running (on linux). I'm not sure I will ever use this tbh, but I suppose I dont mind it existing.

arikorn:
Thanks for the enhancement Julian!

Personally, I'm not yet sure whether I prefer starting the doc server separately or not. If "asked" I would probably recommend starting vite and docs separate unless you're testing `base_url` behavior, in which case `dev:docs --base=/testme/` is much simpler without making "permanent" changes (i.e. .env files).

The reason for `dev:nodocs` is to skip the port-change code, in case we see weird behavior and suspect this PR.  During the development of this PR vite seemed to be a bit more brittle but whether that's because I was primarily messing with vite's config or this code somehow is confusing it, or some recently-added code cause a "pre-existing" issue I don't know but the flag allows us to easily eliminate this as the source of error.

Examples (obviously not directly related to the docs):
~~~
[vite]  ws proxy error:
Error: read ECONNRESET
    at TCP.onStreamRead (node:internal/stream_base_commons:216:20)
...
[vite] ws proxy error:
AggregateError [ECONNREFUSED]: 
    at internalConnectMultiple (node:net:1134:18)
    at afterConnectMultiple (node:net:1715:7)
~~~


If I have time later today I'll look at the behavior in WSL.

webui/vite.config.ts
```diff
@@ -64,10 +152,29 @@ export default defineConfig({
...
-			},
+			[`${normalizedBase}/user-guide`]: noDocServer
+				? {
+						// forward to Express, which may show an error screen
+						target: `http://${upstreamUrl}`,
+						rewrite: (path) => path.slice(normalizedBase.length),
+					}
+				: {
+						// forward to Docusaurus
+						target: `http://${upstreamHost}:4000`,
```
Julusian: I dont think this `upstreamHost` flow is appropriate.

The intention of UPSTREAM_URL is so that I can run vite locally connecting to a companion running on another machine. useful occasionally to do some direct comparison of some changes off the same backend.

Which this will fail with, as docusuarus wont be able to bind to that address, and the proxy wont find anything running on port 4000 to connect to

Julusian:
> The reason for dev:nodocs is to skip the port-change code, in case we see weird behavior and suspect this PR.

I cant tell what you are referring to here, the only thing I can see this changing is ignoring the WEBUI_DOCS env var, and not trying to proxy the docs, but I see no reason to need to disable that now. It may collide with something else, but thats acceptable

webui/vite.config.ts
arikorn: Oh, I missed this comment! That makes sense, so we should always run docusaurus locally?
(Can fix tomorrow)

arikorn: Would the upstream server (on port 8000) be able to serve the user-guide? i.e., should I restore the no-docserver functionality to allow that?

Julusian: > so we should always run docusaurus locally?

Yes, we should always assume docusuarus is run locally (but dont actually run it unless the dev:docs mode is used)

> Would the upstream server (on port 8000) be able to serve the user-guide? 

I dont think we should expect this, it would only do this either when connecting vite to a properly built companion (which is a rare case, and if youre runnign the ui in dev, I dont know if it is better to have the prebuilt or dev docs too).

Although I suppose instead of what I did with serving a static file, forwarding the request to the upstream url would make sense, but I think the value is minimal so its not worth much time/complexity

----

Basically I am worried that while we can improve the DX for docs editing/serving in dev, I am worried that it will bring complexity that will need maintenance and will end up broken every now and then without being noticed

arikorn: OK, docusaurus is now hardcoded to localhost, with a note on each side (vite config, dev-docs script) to modify the other in the unlikely event that it changes.

arikorn:
> > The reason for dev:nodocs is to skip the port-change code, in case we see weird behavior and suspect this PR.

> I cant tell what you are referring to here

Well it's moot now but something wasn't working right with the plugin to launch docusaurus so nodocs helped with that. It turned out to be an ordering issue (order of plugins in the list) but in the end I realized that the plugin approach was a bad idea anyway. Here's the relevant commit message (fd64c01):

> Creating a plugin in vite.config was problematic and also didn't really fit the workflow, since there's really no reason to kill docusaurus each time vite restarts.
> 
> Instead the logic is now in a "tool" webui-dev-docs.mts and done using the `concurrently` package, which should also be more reliable cross-platform.
> 
> env flags are no longer used/needed and nodocs is also no longer needed, thanks to Julian's fallback code along with the lack of a plugin.
> 
> some shared code was added in a new file tools/webui-dev-utils.ts

**@Julusian this should be good to go now, pending your further comments.**

arikorn:
Thanks for the platform-fix and the merge Julian!

PR #4038, opened 2026-03-25
feat: smooth-scroll to an anchor-link in user-guide

arikorn:
...rather than jumping directly to it. It makes it a bit clearer to the user that this is the middle of the page (especially if linking from the Admin window).

Also, show a little space above the anchor, for better context.

## Summary by CodeRabbit

* **Style**
  * Improved in-page anchor navigation with smooth scrolling when reduced motion isn't requested.
  * Added responsive top offset for taller viewports so anchored headings are visible and not hidden behind fixed headers.


arikorn:
thanks Julian, as always.

PR #4039, opened 2026-03-25
chore: fix `yarn dev:docs` startup message

arikorn:
- don't say "starting vite _**and docusaurus**_" if skipping docusaurus!

## Summary by CodeRabbit

* **Chores**
  * Improved accuracy of startup status messages to reflect whether all services are being initialized.

### DIFF
--- a/tools/webui-dev-docs.mts
+++ b/tools/webui-dev-docs.mts
@@ -86,7 +86,7 @@ if (await isPortInUse(4000, docusaurusHost)) {
 	}
 }
 
-console.log(`Starting Vite and Docusaurus...`)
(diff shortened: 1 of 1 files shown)


arikorn:
Thanks Julian, so this reminds me of a persistent-but-never-important-enough question:
when is it a "fix" vs "chore"? I can see updating dependencies or maybe a simple refactor as a chore vs. fixing functionality as a fix (if it's not a new feat), but I can't help noticing that you renamed this one "chore: fix..."!

Julusian:
If its a dev thing, I usually call it chore

PR #4049, opened 2026-03-29
feat: stream expression previews

Julusian:
The aim here is to make it easier to write expressions for options fields; to easily understand the current value of the expression and to inform users when the value their expression produces does not pass validation according to the rules defined by the module.

Show a live-value of expressions next to expression input fields:

image

When hovering over the value, it shows a popover:

image

## Summary by CodeRabbit

* **New Features**
  * Added expression preview functionality that validates and displays expression values in real-time with error handling.
  * Enhanced variable value display component with expanded/collapsed states and popover support for better readability.

* **UI Improvements**
  * Improved styling and layout for expanded variable values with customized alert and popover appearance.
  * Added visual error indicators and validation feedback when displaying variable values.


arikorn:
This is a great idea, and much needed -- particularly for local variables ... and especially for local variables in triggers!

PR #4051, opened 2026-03-30
Expose http port based on ENV variable

codeitlikecody:
Now that the http port is defined by an environment variable, we need to expose the http port based on this variable as well. If there are any other reasons to keep port 8000 open, perhaps we should just add COMPANION_ADMIN_PORT to the list of ports to expose rather than replace the hardcoded port 8000?

## Summary by CodeRabbit

* **Chores**
  * Updated Docker configuration to use an environment variable for admin port mapping, enabling flexible port configuration without rebuilding the image.


codeitlikecody:
I'll leave this then. When someone changes the port, they can set the port to expose in their dockerfile?

PR #4056, opened 2026-03-31
chore: add CodeRabbit review rules for service layer

claw-explorer:
## Summary
- Adds `.coderabbit.yaml` with path-scoped review instructions for `companion/lib/Service/**`
- Enforces three patterns that came out of maintainer review on #4048:
  1. Service handlers must use `ServiceApi`, not inject domain controllers directly
  2. No optional constructor parameters without a real caller that omits them
  3. Import named types instead of computing them with `ReturnType<...>`

## Test plan
- [ ] CodeRabbit picks up the new rules on subsequent PRs touching `companion/lib/Service/`
- [ ] Verify rules don't fire false positives on existing code

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Updated internal code review configuration to establish service-layer development guidelines.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,24 @@
+reviews:
+  path_instructions:
+    - path: "companion/lib/Service/**"
+      instructions: |
+        Service layer architecture rules:
+
+        1. **Use ServiceApi, not direct controller injection.**
+           Service handlers (HttpApi, Tcp, Udp, etc.) must access backend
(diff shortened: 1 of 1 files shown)


Julusian:
I don't think this makes sense for the .coderabbit.yaml, it should be some generic agent md so that it gets picked up while writing the code.

At this stage I don't think some incomplete fragments of agent input is a good idea. I think that we would need to write out some general rules before adding refinemnt for niche edge cases like these

Issue #217, opened 2018-09-04
M32 Show control and select/solo


Gaiamus:
Is it possible to add show control buttons for M32/X32? I've got a M32r, the show control buttons are missing on the console and I cannot chose these functions as a user define key on the console. This issue would be sovled by scene buttons for "Last", "next", "undo" and "GO".

Also, it would be awesome to be able to choose "solo" and/or "select for a mix bus when doing monitors from the console. In this way I can set the console to select follows solo or solo follows select. When I press a the button I will then hear the mix in my wedge and att the same time flip the faders on the console for that mix. The dream for all monitor techs using this console!

Gaiamus:
Any news on this?

Gaiamus:
I manage to get "select a mixbus" working over generic OSC.
I found the OSC commands at this site.
https://behringerwiki.musictribe.com/index.php?title=Status_(/-stat)_data

osc: Send integer
OSC Path: /-stat/selidx
Value: 48-63

Select channel index
0-31: Ch 1-32
32-63: Ch 32-64
64-47: Aux in /USB
48-63: Bus master
64-69: Matrix 1-6
70: L/R
71: Mono/Center

PR #4065, opened 2026-04-02
feat: Add UI to actions whose definitions indicate they return a result, to optionally store that result in a local or custom variable

jswalden:
The one moderately-major thing in all this that I am uncertain about, is how to propagate a result to a subsequent concurrent action group *that contains waits*.

If I read the code correctly, Companion basically splits concurrent action groups into sets of non-wait actions, and then it performs all those non-waits of the first set at once, then does the intervening wait, then does the next set of non-waits, etc.

I think it is reasonable to propagate a previous result to all actions in a concurrent action group that doesn't contain any waits.  It is less clear that it makes sense to do so when the action group contains waits.  Or that they should propagate beyond the first set of actions?  For prototype-ful purposes I just acted like the waits weren't there.  For actual suitability for landing purposes, maybe that should change.  Don't propagate if there are any waits?  Only propagate to actions before the first wait?  The answer isn't immediately obvious to me.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Actions can now return values that are automatically stored in variables
  * Added support for storing action results in local and custom variables
  * Users can configure where action results should be saved

* **Bug Fixes**
  * Improved action validation and error handling for missing connections


companion/package.json
```diff
@@ -1,104 +1,104 @@
...
+		"@types/yazl": "^3.3.0",
+		"supertest": "^7.2.2",
+		"typescript": "~5.9.3",
+		"webpack": "5.104.1",
+		"webpack-cli": "^7.0.2"
+	},
+	"dependencies": {
+		"@companion-app/shared": "*",
+		"@companion-module/base-old": "npm:@companion-module/base@~1.14.1",
+		"@companion-module/host": "file:/home/jwalden/Christianity/Sound/github/module-base/packages/host/",
```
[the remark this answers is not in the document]
jswalden: Yeah yeah, this is bad.  Acknowledged already, doesn't matter at this phase of things, will change later if/when we get to that point.

jswalden:
> Internal action results are still dropped on this path.

I chose for internal actions to not produce their own results, and also not to propagate a prior result across them.  (Why should they?  Don't see good reason for them to, honestly.)  It makes most sense IMO for only an action that produces a result itself, to be source of one.

> Please lock down previousResult semantics across concurrent groups with waits

Flagged in the initial comment.

> Consider making previousResult optional in the interface

Explicit is better than implicit.  There's always a previous result value/variable.  It's not actually optional semantically.  (I even considered using `const previousResult: VariableValue = undefined` every place, for documentation, but didn't carry through on it.)

> Nice chaining logic — consider avoiding in-place mutation of extras

I kind of prefer less object creation churn, but Julian's the one who will have a meaningful preference here, probably.

docs/user-guide/3_config/variables.md
```diff
@@ -121,3 +121,5 @@ Additionally, in some places there are some builtin local variables under the `$
   - Variable: `this:page_name`
 - The id of the surface triggering this action
   - Variable: `this:surface_id`
+- The result of the action sequentially preceding this one
+  - Variable: `this:result`
```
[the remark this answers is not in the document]
jswalden: Brevity/clarity versus precision tradeoff.  Can tweak if someone wants it.

jswalden:
This will fix #4064

makstech:
The idea is awesome and I 100% support it but my **suggestion** to your question
> how to propagate a result to a subsequent concurrent action group that contains waits.

Why not use something like n8n does it: any node that's downstream, can access any upstream node's result using the node's name. For example, first node is called "Message", the next node can get the data directly with `$json` (in n8n) or `$(this:result)` with your implementation in Companion, but then next nodes or even the current one, can also get it with `$('Message').item.json`. This is an example that works elsewhere and is just an idea, but obviously completely out of scope for this PR.

Julusian:
I'm wondering if this should work such that for any action definition which says 'i can return a value' we let the user pick a local-variable name to write the value into.  
move the state into the existing local variable system, instead of introducing a new flow for state that the user will need to understand behaves differently.

This would make it similar to how the generic-http module (and others?) do this today, but as they predate local-variables they can only write to custom variables. Perhaps it should be able to do both, but lets limit this to local to begin with.

I see pros and cons to each:
* This will be more tolerant to multiple concurrent executions, using local variables will not
* Local variables makes it obvious how things will flow around waits
* Sometimes, the user will want to fire off multiple requests in parallel and operate on the result of all of them, local variables solves the id/reference problem
* This is simpler to use in simple scenarios, no boilerplate of creating a local variable

> $('Message').item.json.

I'm not opposed to something like this in the future, but I think what we use as the identifier needs some thought first. Something like this would be a start to addressing some other requests to reference/update properties of actions/feedbacks (some of which I am not sure is a good idea)

jswalden:
> Why not use something like n8n does it: any node that's downstream, can access any upstream node's result using the node's name. For example, first node is called "Message", the next node can get the data directly with `$json` (in n8n) or `$(this:result)` with your implementation in Companion, but then next nodes or even the current one, can also get it with `$('Message').item.json`.

My concern here is that an action that sets a result, should not have that result remain live for unbounded period of time.  (At least not unless a subsequent user of it chooses to keep it live.)  Tying results directly and only to a bounded set of sequentially-next actions does this.  Whereas if you could refer backwards to previous actions, perhaps even dynamically, then every action's result has to be kept alive "just in case".

There are plainly a bunch of ways to skin this cat.  "Use it immediately or lose it" doesn't seem like too complicated a way to grasp, to me.

jswalden:
> I'm wondering if this should work such that for any action definition which says 'i can return a value' we let the user pick a local-variable name to write the value into. move the state into the existing local variable system, instead of introducing a new flow for state that the user will need to understand behaves differently.

That makes some idiomatic sense.

> I see pros and cons to each:
> 
> * This will be more tolerant to multiple concurrent executions, using local variables will not

Yeah.  (If I were to grump on terminology, I'd name local variables as "per-entity variables", or better non-internal idiom terminology.  And then maybe have some kind of truly local variable that can be created by an internal action, set from expression [or with this enhancement from an action result], that remains live only for the duration of the sequential action group it's found in.  And would not have concurrency considerations to think about.  But I digress.)

I guess the concurrency of local variables is my biggest quirk/beef here.  Same as, but at least significantly reduced in troublesomeness, of a module variable, or of writing to a custom variable.

> * Local variables makes it obvious how things will flow around waits

I think you could arguably say that implicitly piping the result distributing it across a concurrent action group is...of uncertain intuitiveness, as mechanism to enable multiple distinct manners of consuming an action result.  So possibly the root problem isn't really the waits.

(I don't fully understand why waits in concurrent action groups are even given special meaning.  Why not just require users to create a sequential action group, then interleave waits between concurrent action groups?  But I digress even harder...)

> * Sometimes, the user will want to fire off multiple requests in parallel and operate on the result of all of them, local variables solves the id/reference problem
> * This is simpler to use in simple scenarios, no boilerplate of creating a local variable

Forcing stuff to be named doesn't feel like boilerplate to me.  What does feel like boilerplate, is having to create/name them out of the actual UI of the step flow being executed.  And having that variable exist across all steps, rather than only narrowly in the exact set of actions the computation/initialization happens in.

Julusian:
> I don't fully understand why waits in concurrent action groups are even given special meaning

it was done to make it easier to program these buttons. having a wait inside of a concurrent group otherwise makes no real sense, and this way we probably cut down the number of groups the user needs to add massively (perhaps even avoiding them altogether).

The thing to remember is that this was to replace the old behaviour where we never awaited actions, and it was possible to give each action an either relative or absolute start delay. That relative mode is what this is emulating.  

And consider that adding a delay used to be simply typing in a number. Then it moved to an action and many users couldnt find it. Making them also add a group and drag their actions around to arrange them to get the same behaviour would likely not have been well received.


> And would not have concurrency considerations to think about. 

Another solution to concurrency issues would be to add a checkbox onto the 'user value' type, so that it ignores external updates while running the stack of actions.  
Maybe that is worse for user complexity, but also maybe not.  
Maybe this 'cached' mode should be the default, with an opt-out for the few scenarios that would need it (eg, a loop waiting for another button/trigger to change a variable and break the loop)

> I think you could arguably say that implicitly piping the result distributing it across a concurrent action group is...of uncertain intuitiveness,

True, I didn't consider that. It for sure wont be able to work within the concurrent group, but the group could return a single value (from I guess the last?) to be used later? But this will make it hard to explain to users what the variable will refer to

---

I think my main concern here is just that I feel this will be a half-solution to concurrency issues over using local-variables. I expect that in many cases users will still want or need to write the result to a local-variable (even if just to make debugging/tracing the values easier).

For anything sensitive to concurrency, I do also wonder if our behaviour of allowing it to run in parallel is good enough. Not an argument against this, just that if we adopt a flow like home-assistant where they have modes:
image

---

So I am leaning towards saying that I think this needs to be done with local-variables, as this approach essentially does not work in concurrent groups.

For 4.3, I am open to including the ability to write action results to local-variables (but be aware that I had originally intended to release 4.3 by now, so that could happen any day now). And any work on better handling concurrency I would like to delay for 4.4, so it can have some time to be tested in the betas. Doesn't mean it can't be implemented now, it just won't be merged until 4.3 is finished

jswalden:
> > I think you could arguably say that implicitly piping the result distributing it across a concurrent action group is...of uncertain intuitiveness,
> 
> True, I didn't consider that. It for sure wont be able to work within the concurrent group, but the group could return a single value (from I guess the last?) to be used later? But this will make it hard to explain to users what the variable will refer to

It's entirely reasonable for a concurrent group's result to always be `undefined`, IMO.  It's basically an internal action same as every other internal action in not returning a result.

(I need to tweak slightly so an empty sequential group results in `undefined`.  Worth mentioning while it's on the mind.)

> I think my main concern here is just that I feel this will be a half-solution to concurrency issues over using local-variables. I expect that in many cases users will still want or need to write the result to a local-variable (even if just to make debugging/tracing the values easier).

It mostly is a Companion-side decision how/where to sink result values, once the base/host side of it is adequately solved.  Sink into _something_ can be done as first step.  Enabling sink into more stuff is deferrable.

> So I am leaning towards saying that I think this needs to be done with local-variables, as this approach essentially does not work in concurrent groups.

Given the name "concurrent", nobody should expect bare act of yielding a result to be usable in them.  Get a value, then set it somewhere, by nature incorporates a dependency that requires sequential operation.

> For 4.3, I am open to including the ability to write action results to local-variables (but be aware that I had originally intended to release 4.3 by now, so that could happen any day now).

...short of there being some kind of breaking API change to make, and wanting it to be in before the 2.0 API actually ships, I don't think you should delay.  I've been on the side of things you might want in a release, that nevertheless were rejected for one because not in time for it.  Logrolling near end of a cycle can end badly, and always creates precedent for more delays in the future.  It's rarely worth it for a bare enhancement.  If this is not in time, so be it.  There's always next time.

That said.

*If this is a time where breaking API changes can be made still* -- is it?  is a hypothetical 2.0.1 or most pedantically 3.0.0 out of line now, before there has been a Companion release supporting it? --  if you changed the callback return type to `undefined | Promise<undefined>` _now_ (and assuming this doesn't create linter warnings for code that uses unadorned `return` and falling off the end -- I'm not sure whether it does), then adding in `| JsonValue` in the future would not be a breaking change (for code that complies with the types, at least) requiring `hasResult` be added.  (A `hasResult` that requires gnarly tricks to support only `{ hasResult: true; callback: (...) => JsonValue | Promise<JsonValue> }` xor `{ hasResult?: false; callback: (...) => void | Promise<void> }` -- tricks that mangle Typedoc output in ways that are hard to reduce.)

> And any work on better handling concurrency I would like to delay for 4.4, so it can have some time to be tested in the betas. Doesn't mean it can't be implemented now, it just won't be merged until 4.3 is finished

I don't think an action that returns a result can ever be meaningfully usable (at least as to the result value; side effects are still usable, of course) inside a concurrent group.

And I will reiterate that I don't think you should hold anything up for this -- save that *if* we are still at a juncture where incompatibly changing the type that `CompanionActionDefinition.callback` returns is viable, we should change it from `void` to `undefined`.

companion/lib/Controls/ActionRunner.ts
```diff
@@ -32,25 +33,23 @@ export class ActionRunner {
...
-				this.#logger.silly('trying to run action on a missing instance.', action)
-			}
+			return undefined
 		}
+
+		const instance = this.#instanceController.processManager.getConnectionChild(action.connectionId)
+		if (!instance) {
+			this.#logger.silly('trying to run action on a missing instance.', action)
+			return undefined
+		}
```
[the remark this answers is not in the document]
jswalden: This is basically intentional.

jswalden:
Updated to add actions to set a local/custom variable from the result of a series of actions.

A series of actions is allowed mostly because I cribbed from `logic_if` which allows sequences.  But I can also rationalize this as allowing actions/computations that culminate in the produced result, to be enclosed within the overall variable-set action.  (If it were limited to one action, you could still do a series by using a sequential action group as the action.)

jswalden:
There's already `maximumChildren: 1` to permit restricting to just one action.  Made that change.

I think that's everything previously mooted here, so this seems ready to be taken for this release, or sat on til after release.

I'd still argue for the mini-type-fix of making callbacks return `undefined` (if breaking type changes are still in-bounds) if you delay on this.  But I found ways to make the generated typedoc results of https://github.com/bitfocus/companion-module-base/pull/217 tolerable, so forcing `hasResult` on module authors -- rather than us simply expanding the range of callback-returnable types from `undefined` to `undefined | JsonValue`, backwards-compatibly -- is a workable decision IMO.

Julusian:
I'm not keen on this requiring putting an action inside another action to be able to save the result.

I am worried about a combination of:
* User discovery. How will I know that an action I have added can return a value that I might want to use?  Users are used to having this option on the action (since before 3.0 for a few modules which use an experimental flow), and will be confused if it disappears with no indication of what to look for instead.
* Usage UX. The less clicks the better (within reason), as long as we arent bloating ui. Especially as I imagine a common flow will be to add the action, then realise they want to save the value and have to rearrange things to achieve that

I would much rather that this added an extra field in the ui to any actions which support this, handled within companion. I think the closest we have to this today is the inverted field for boolean feedbacks.


So I am wondering about a couple of auto-added fields at the bottom of the action by the ui, when the action supports this:
* Store result - a dropdown (disabled/local variable/custom variable)
* Variable name - same style field as we use today for storing to variables? should have the expression toggle. This should only be visible when 'store result' is not disabled. (may want to be different depending on which of local/custom are selected)
* Create if not exists - only be visible when 'store result' is not disabled. matches what we already have.

I think that by doing this we both solve the discovery issue with minimal ui bloat, matching a flow users are already used to (in generic-http and friends), and simplify usage by making it a couple of clicks to setup.


In either case the ui will need to be aware of which actions support this, so that it can stop the user adding an action to the  `Set to result of an action` actions which dont support returning a value. 

----

For the implementation of this, I wonder if we should have an 'internalOptions' object on each entity, for storing these values? This could be well typed too I expect. Ideally the feedback 'inverted' property could be moved into this too, but I would leave that for a follow up as that will be a bit fiddly to untangle. 


----

As a follow up to this (for me to do), I will look into what we can do to auto-migrate some existing modules across to this instead of using the old `setCustomVariable` flow that they are using today. Either approach should be possible to fixup for the modules we know about (likely with comapnion assuming more knowledge on module actions than we do today, but I think that acceptable as its only a couple)

jswalden:
> I would much rather that this added an extra field in the ui to any actions which support this, handled within companion. I think the closest we have to this today is the inverted field for boolean feedbacks.
> 
> So I am wondering about a couple of auto-added fields at the bottom of the action by the ui, when the action supports this:
> 
>     * Store result - a dropdown (disabled/local variable/custom variable)
> 
>     * Variable name - same style field as we use today for storing to variables? should have the expression toggle. This should only be visible when 'store result' is not disabled. (may want to be different depending on which of local/custom are selected)
> 
>     * Create if not exists - only be visible when 'store result' is not disabled. matches what we already have.

This seems like solidly better UI than my approach.

> In either case the ui will need to be aware of which actions support this, so that it can stop the user adding an action to the `Set to result of an action` actions which dont support returning a value.

The `hasResult` boolean in the action definition should be sufficient to know this, if I'm not underthinking it.

> For the implementation of this, I wonder if we should have an 'internalOptions' object on each entity, for storing these values? This could be well typed too I expect. Ideally the feedback 'inverted' property could be moved into this too, but I would leave that for a follow up as that will be a bit fiddly to untangle.

On no more than two minutes' thought that seems plausible.  Need to noodle a little more, not at EOD, more before being certain.

I'll try to get back on this in the next few days.

Julusian:
> The hasResult boolean in the action definition should be sufficient to know this, if I'm not underthinking it.

yes it should be. defining that requirement in the generic-ness of the entity child lists requires a little bit more glue to handle. but hopefully not necessary now

> I'll try to get back on this in the next few days.

no rush :)

phillipivan:
Just piping in here, based on my experience working with the value feedbacks, it would make life considerable easier for users if the current value or last value (if that makes more conceptual sense here) of the returned variable was brought to the UI along with a copy button for the variable name. Before these were added in 4.3 I was frequently having to drop local variables on buttons just to confirm values while configuring an action/feedback.

jswalden:
Okay, for the most part this is all "ready" to go.  At least in terms of it not being a waste to look at, if nothing else.

The extent of change has grown a fair bit from the "just add some actions that can do all the stuff" previous state.  The final two commits here are not for committing, ignore them and/or realize they would be excluded.  That done, we have:

> 59dda5ab1 refactor: Extract local variable manipulation out of the `InternalVariables` module fragment into a standalone controller.

This is the bit about local-variable manipulation only being possible through the variables fragment now, and this creating a controller that `ActionRunner` can later in the stack use in addition to the fragment.

> cc3998093 refactor: Convert internal action handling-by-`definitionId` to use `switch` statements instead of `if`/else if`/`else`.
> 64abcba86 refactor: Add scaffolding for internal actions to return a result (that is for the moment ignored).

In order for internal actions to be able to return results -- none do now, but I feel like I ought be making it _possible_ for them to, so when you eventually want such an action you don't have to do the plumbing yourself -- I change `executeAction` from returning true (handled) or false (not), to returning `null` (not handled) or `{ result: JsonValue | undefined }` (perhaps behind a promise).  IMO it is nice to avoid typing out `{ result: undefined }` everywhere, so making all these use `switch` and then `break` to a single trailing `return { result: undefined }` is aesthetically pleasing.  Feel free to disagree.

> 6b1de48a6 refactor: Tweak action-running code for simpler flow-through once actions return results.
> 825a32270 refactor: Compute the result of an action when running it, then for the moment simply discard it.

This just gets the internal action `executeAction` and non-internal `actionRunn` flows into place where they're both returning a result value from the action -- which will be `undefined` if the action doesn't have a result.  Then the result value gets visibly dropped, to be adjusted in subsequent patching.

> 502740746 refactor: Refactor `isInverted` expression evaluation/invalidation code into a system that allows additional kinds of special expressions to be handled.

`EntityIsInvertedManager` is all about tracking is-inverted expressions, because those expressions exist outside of action/feedback options where they would otherwise be automatically parsed/tracked.

What this PR aims to do, is allow for another field similar to `isInverted?: ExpressionOrValue<boolean>` in the entity model -- `storeResultTarget?: StoreResultToLocalVariable | StoreResultToCustomVariable`.  These are `{ target: <discriminator-literal>; variableName: ExpressionOrValue<string> }` combined with either `{ createIfNotExists: boolean }` or `{ location: ExpressionOrValue<string> }` depending whether custom or local.

It seems to me that the simplest/clearest way to do this, is to generalize is-inverted handling so that it's just one instance of "special expression", that exists outside of option values.  So in this rev I do that -- defining a record of special expression name to expression value, which is initially just `isInverted => boolean`.  A similarly structured object is added to the special-expressions manager, with contents of each key being special expression-specific handling functions (like to compute a new value of the special expression, be updated on these new values as `updateIsInvertedValues` does now).  Then most functions in the manager just loop over `Object.values(this.#specialExpressions)` to do their thing.

Updating is still debounced acros all updates.  Entities are tracked by the specific special expression at issue.  Entities are _forgotten_ across all special expressions at once -- but it would be easy to add a later optimization to forget only for specific cases.

> 5a82d4f4d refactor: Rename special-expressions files now that edits to them are done.

I didn't rename any files for the new class names in the previous rev.  This renames them, separately so as not to break readable rev history.

> dc388e309 feat: Add UI to actions that return a result, to permit storing that result in a local or custom variable.

This is the biggest meat of change.  There is not a lot you could do to trim it down.  I guess the webui changes could have been split into their own rev, and the actionrunner changes could be a separate rev, and and and.

I tried to simply copy pretty heavily the `isInverted` handling when devising similar-yet-different code for action result handling.  It's totally possible I screwed up something somewhere in it, that I'm not aware of.

I added no tests for any of this, just worked through the UI manually.  I can likely be guilted into adding tests, but this probably needs a skim before I discover I've gone astray somewhere that would invalidate any test-writing.

The names for "storeActionResultTarget", "storeResultTarget", etc. are under-thought-out and mostly just named to have a name to refer to at all.  I did not have ideas I really liked on this.  Maybe `storeResultTo` is my current idea, but you might have a better one.

I don't really understand the logic of `ControllEntityInstance#subscribe`, and the `trackEntity` I added there might be off-base in some way.  Also the `this.subscribe(false)` in `CEI#setStoreActionResultTarget` maybe should pass in `EntityModelType.Action`?  But is-inverted doesn't pass in feedback and I was copying.

Most of the scattered `updateStoreActionResultTargetValues` logic is copied.  You would understand better what entity sources do not need recurring into to find all actions that must be updated.

I _assume_ the `reportChange` argument in `entitySetStoreActionResultTarget` is correct but really have no idea.

This PR is not developed against `main` but against a version of it that's slightly older.  When I rebase the PR onto `main` -- it does so without error -- trying to switch an action result target from local variable to `<discard>` triggers a tRPC error about the setStoreActionResultTarget mutation not processing argument right.  It's fantastically bad luck, I think there's a chance that the update you did a few days back to `zod` ([4.4.2->4.4.3](https://github.com/colinhacks/zod/compare/v4.4.2...v4.4.3)) caused my definition of `zodStoreActionResultTarget` to trigger errors.  Or I just don't understand what I'm doing enough.  But somehow this PR works all fine, but a rebased version of it onto `main` does not.

My React skills are meager, and I almost can't believe my `useRef` usage actually worked exactly as I meant it to the first time I got all the code to compile.  The webui changes could easily have things wrong in them that I just don't know I did wrong!

Oh, and the final patch that actually makes webui changes, that depends on #217 landing and an updated `@companion-module/host` to be available to rely on.  But all changes _before_ that patch, do not require any changes to `@companion-module/host` (or base).

I think that's a summary of everything here that is worth flagging.  There are things I could do to split up that final patch (e.g.  move the action option-commoning into shared-lib into a separate rev), but those aren't really the parts of it that are questionable.

jswalden:
@coderabbitai review

companion/lib/Internal/BuildingBlocks.ts
```diff
@@ -249,95 +250,103 @@ export class InternalBuildingBlocks
...
+					}
+					return { result: undefined }
+				})
+			}
+			default:
+				return null
 		}
+
+		return { result: undefined }
 	}
```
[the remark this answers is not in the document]
jswalden: That's literally the point of returning `null` or `{ result: <value> }`, it forces callers to handle the returned-`null` case.  This isn't an issue at all.

shared-lib/lib/Model/EntityDefinitionModel.ts
```diff
@@ -21,6 +21,8 @@ export interface ClientEntityDefinition {
 	learnTimeout: number | undefined
 	showInvert: boolean
 
+	hasResult?: boolean
```
Julusian: perhaps this should be `actionHasResult` to clarify this is for actions?

A comment block to clarify this might be good too (see the properties below)

webui/src/Controls/Components/EntityCommonCells.tsx
```diff
@@ -70,6 +72,13 @@ export const EntityCommonCells = observer(function EntityCommonCells({
 		[service]
 	)
 
+	const setRawStoreResult = useCallback(
```
Julusian: I think instead of this callback wrapper, `service.setRawStoreResult` could be passed into the child component directly

webui/src/Services/Controls/ControlActionsService.ts
```diff
@@ -73,6 +74,10 @@ export function useActionRecorderActionService(sessionId: string): IEntityEditor
 			setEnabled: undefined,
 			setHeadline: undefined,
 
+			setRawStoreResult: (_entityId: string, _target: RawStoreResult | undefined) => {
```
Julusian: I suspect this will be a slight problem in the action recorder, where it will have the ui to set these fields, but it wont work.
I think it makes sense here to simply hide this ui, other features do this by the methods being optional (the few `undefined`s above)

companion/lib/Instance/Connection/Thread/Entrypoint.ts
```diff
@@ -103,10 +103,19 @@ const ipcWrapper = new IpcWrapper<ModuleToHostEventsNew, HostToModuleEventsNew>(
...
 			const res = await instance.executeAction(msg.action, msg.surfaceId)
-			return {
-				success: res.success,
-				errorMessage: res.errorMessage,
-			}
+			return res.success
+				? {
+						success: true,
+						result:
+							// This can be `res.result` after the `@companion-module/host`
```
Julusian: should probably do this in the simple way assuming the lib has been updated (it will be by the time this is merged)

companion/lib/Controls/Entities/Types.ts
```diff
@@ -22,8 +27,16 @@ export interface NewFeedbackValue {
 	value: FeedbackValue
 }
 
-export interface NewIsInvertedValue {
```
Julusian, line 25: should this have been removed?

companion/lib/Controls/Entities/EntityListPoolExpressionVariable.ts
```diff
@@ -96,4 +97,22 @@ export class EntityListPoolExpressionVariable extends ControlEntityListPoolBase
...
 		this.tryTriggerLocalVariablesChanged(...changedVariableEntities)
 	}
+
+	/**
+	 * Update the storeResult values on the control with new calculated
+	 * storeResult values
+	 * @param newValues The new storeResult values
+	 */
+	override updateStoreResultValues(newValues: ReadonlyMap<string, NewSpecialExpressionValue<'storeResult'>>): void {
+		const changedVariableEntities = this.#localVariables.updateStoreResultValues(newValues)
```
Julusian: I think this method should be a no-op for an expression variable, as none of the entities inside of it can be actions

companion/lib/Controls/Entities/EntityListPoolTrigger.ts
```diff
@@ -119,4 +120,24 @@ export class ControlEntityListPoolTrigger extends ControlEntityListPoolBase {
...
+
+	/**
+	 * Update the storeResult values on the control with new calculated
+	 * storeResult values
+	 * @param newValues The new storeResult values
+	 */
+	updateStoreResultValues(newValues: ReadonlyMap<string, NewSpecialExpressionValue<'storeResult'>>): void {
+		this.#actions.updateStoreResultValues(newValues)
+
+		const changedVariableEntities = this.#localVariables.updateStoreResultValues(newValues)
```
Julusian: Similarly, I think only the `this.#actions.updateStoreResultValues(newValues)` is needed here

companion/test/Controls/Entities/EntitySpecialExpressionManager.test.ts
Julusian, line 1: some new tests for the new flows would be great if possible

companion/lib/Controls/Entities/EntityInstance.ts
```diff
@@ -663,6 +706,12 @@ export class ControlEntityInstance {
 				newPropsData.styleOverrides && newPropsData.styleOverrides.length > 0
 					? newPropsData.styleOverrides
 					: feedbackData.styleOverrides
+		} else if (this.#data.type === EntityModelType.Action) {
```
Julusian: I don't think we need this.

This method was written for module upgrade scripts to perform some updates, but upgrade scripts dont need to manipulate this.

companion/lib/Controls/Entities/EntityInstance.ts
```diff
@@ -837,6 +888,62 @@ export class ControlEntityInstance {
...
+	updateStoreResultValues(
+		newValues: ReadonlyMap<string, NewSpecialExpressionValue<'storeResult'>>
+	): ControlEntityInstance[] {
+		const changed: ControlEntityInstance[] = []
+
+		const newValue = newValues.get(this.#data.id)
+
+		if (this.type === EntityModelType.Action && newValue) {
+			const cachedStoreResultValue = this.#cachedStoreResult
+			let thisChanged: boolean
```
Julusian: perhaps this could do a `isEqual(a,b)` instead of breaking it out manually?

It wouldn't be a terrible if it invalidated slightly too often by accident, so the benefit of simpler code might be better

companion/lib/Controls/Entities/EntityListPoolBase.ts
```diff
@@ -645,6 +650,18 @@ export abstract class ControlEntityListPoolBase {
...
 	}
 
+	/** Set where this action's result will be stored (if at all). */
+	entitySetRawStoreResult(listId: SomeSocketEntityLocation, id: string, target: RawStoreResult | undefined): boolean {
+		const entity = this.getEntityList(listId)?.findById(id)
+		if (!entity) return false
+
+		entity.setRawStoreResult(target)
+
+		this.reportChange({ redraw: true })
```
Julusian: I dont think we need to redraw here, this shoudlnt affect any drawing

Julusian (approved):
Sorry for taking so long to properly look at this.  
It looks good!

I have a handful of small niggles, but nothing major

Julusian:
@jswalden I have updated the module-api after merging your pr there, so there are a couple of small merge conflicts here now. This does mean that once you have resolved it the tests and types should be fully happy though

companion/lib/Controls/Entities/Types.ts
jswalden: It was only used for the one-off `updateIsInvertedValues(newValues: ReadonlyMap<string, NewIsInvertedValue>)` functions before.  As those have been supplanted by the special expression-based `updateIsInvertedValues(newValues: ReadonlyMap<string, NewSpecialExpressionValue<'isInverted'>>)` form -- that is to say, this interface is now replaced by `NewSpecialExpressionValue<Expression>` -- this interface is no longer necessary or used.

shared-lib/lib/Model/EntityDefinitionModel.ts
jswalden: Good call, I hadn't considered the `feedbackAffectedProperties` precedent.

Also for consistency with that, I'm going to make this non-optional, while allowing it to be optional in `InternalActionDefinition` derived from it.  Because additional places get touched in the change to making it required, I'll separate the adjustments along these lines into a separate rev for easier examination.

companion/lib/Controls/Entities/EntityInstance.ts
jswalden: I'd looked at callers and gotten that sense.  Figured better to make an unnecessary change for perusal/possible rejection than to omit it.

webui/src/Services/Controls/ControlActionsService.ts
jswalden: Making this optional extended changes into a couple more places, so also relegated these adjustments to their own rev for examination.

jswalden:
Okay, responded to all review comments...plus a touch more.  Running through the notable adjustments:

### New special expression values can be computed in standalone functions

After the generalize-`isInverted`-handling patch, I added:

* refactor: Move `#computeIsInverted` into a standalone function, as on second look it doesn't use `this`.

for the reason in the summary.  I did this because it seemed like the change _ought_ be in the initial refactor, not tacked atop handling for `storeResult` as well.  Unfortunately this meant rebasing it through the addition of `storeResult` handling -- so the original "Add UI to actions that return a result" necessarily had to address those conflicts, and not be fixed in a clear followup patch.

### Don't track `storeResult` for disabled actions

I noticed that `isInverted` tracking doesn't happen for disabled feedbacks, so I adjusted the code path so the same is done for `storeResult`.

(It would be really nice if we could recognize special expressions that contain no variables and simply _never_ track them, to keep the entity maps minimally sized given that most feedbacks don't have `isInverted` and most actions don't have `storeResult`, even without any expressions/parseVariables strings in them.  This seems like potential followup material.)

### Changes made that debatably could use a quick skim

I addressed the `ClientEntityDefinition.actionHasResult` in a separate rev because of the slightly further changes required there.  Same for making `IEntityEditorService.setRawStoreResult` also allow `undefined`.  The bulk of other review comments were menial so I dumped them in the "chore: Respond to some review comments." rev.

### `storeResult` tests

I did in fact add a boatload of `storeResult` tests.  Perfectly possible some of it could be more cleanly done, but I personally have always tried to tolerate a little more code fugly in testing code because it, unlike actual functional code, in principle is ignorable as long as it passes, and polishing it is IMO not an ideal tradeoff in use of time.

I recommend squashing the "move compute-isInverted code to a standalone function" into its preceding rev, and I recommend squashing everything from "Add UI to actions that return a result" to end into one patch.  The splitting is intended to be helpful for reviewing of incremental modifications, but I don't think it helps future readers to see the intermediate states directly.

PR #4068, opened 2026-04-03
chore: prevent EMFILE error when starting watchers

bryce-seifert:
Prevent some potential errors when starting the watchers by limiting the paths more explicitly, and also waiting to start the module watcher until main is started

## Summary by CodeRabbit

* **Chores**
  * Enhanced development file watcher configuration to target specific paths and defer secondary watcher startup, improving initialization performance and reducing resource usage.

### DIFF
--- a/tools/dev.mts
+++ b/tools/dev.mts
@@ -105,26 +105,19 @@ concurrently([
 
 const cachedDebounces = {} as Record<string, any>
 
-chokidar
-	.watch('..', {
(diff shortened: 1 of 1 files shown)


Julusian:
Makes sense to me

I'm now wondering if maybe we should try using the 'new' `node --watch` mode instead of using chokidar for this.. I have been putting it off because the current script worked fine, but as it apparently doesnt maybe it is time to try it

PR #4069, opened 2026-04-03
fix: incorrect/missing `htmlFor` linkages

arikorn:
NOTE: I am posting this "early" for comments before I continue down this path...

`htmlFor` in a form label needs to be linked to the `id` property of the input element in order to link them for accessibility (such as screen readers), UI behavior (for example, click on the label sets focus on the input) and certain other features.

In many cases in the current codebase, the `htmlFor` is set but the corresponding `id` is missing (or only `name` is set to that value, which is the same thing).

This PR starts fixing it. (it is not yet complete:  I am posting this "early" for comments before I continue down this path...)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed form field naming inconsistency in grid configuration.
  * Adjusted minimum grid dimensions to require at least 1 row and column.

* **Improvements**
  * Enhanced form field accessibility throughout the application by ensuring labels are properly linked to their corresponding input elements, improving keyboard navigation and screen reader support.


arikorn:
@Julusian - I'm converting to draft (a) because it's only partly done (b) I'm waiting for your feedback before proceeding and (c) it will conflict with PR #4070 which should be merged first.

Julusian:
Do we still need the name fields as well as the id, or should those be removed at the same time?

I don't have a problem with this. It does sound like its going to be sprawling (especailly the useId() pattern), but I suppose there is a need for this change

arikorn:
Thanks for the feedback and go-ahead, Julian.

>Do we still need the name fields as well as the id, or should those be removed at the same time?

Probably should be kept. `name` is used for the form (it's what is sent on submit, if you're serializing, I believe...or something like that), so at the very least, it can't hurt to have both. Technically they're completely independent: id is used to link the input and the label, name is for submit. That's why in some places I changed `htmlFor/id` values to be more securely unique but left `name` alone, so there's no chance of breaking any putative submit behavior.

>I don't have a problem with this. It does sound like its going to be sprawling (especailly the useId() pattern), but I suppose there is a need for this change

Yeah, my intention in this pass is probably to fix only the unlinked `htmlFor` props, i.e. `htmlFor` is there but `id` is missing. For whatever reason, the devTool classifies those as errors ("incorrect use") whereas it gives only "improvement advice" if it's missing entirely. Even though the effect is the same either way!  What you see here is around halfway through, maybe a bit more.

Theoretcially, it might have been simpler if label + input was always in a component (then you just link them in the component) but I suspect that would be an even more massive refactor!

Just to restate from my previous comment: I'll be waiting on this until the DOM errors PR is resolved, so I can resolve the conflict I created/avoid even more of them!

Julusian:
>  if you're serializing, I believe...or something like that)

I dont think we ever truely submit any of these forms, so the name probably isnt of any actual value to us. Not even sure if there are any submit handlers..

> That's why in some places I changed htmlFor/id values to be more securely unique but left name alone, so there's no chance of breaking any putative submit behavior.

Im open to using useId everywhere so that we can be sure there wont be collisions. up to you

> Theoretcially, it might have been simpler if label + input was always in a component

Yeah, not easy though as there are a few different ways we want to have the elements arranged

----

A side thought; if you want to explore any unit tests for ui code to avoid regressions in anything, that would be good (but perhaps an epicly large task to do meaningfully). Even a couple of small tests is a start. I think I've used https://testing-library.com/docs/react-testing-library/intro/ for a few minimal dom only tests (wiihtout rendering)

arikorn:
>I dont think we ever truely submit any of these forms, so the name probably isnt of any actual value to us. Not even sure if there are any submit handlers..

I thought I saw some serialization related to tRPC, but regardless it never hurts to keep the name as-is, and just reduces the chances of me accidentally breaking something!

> Im open to using useId everywhere so that we can be sure there wont be collisions. up to you

We could do that, though it seems a bit of a sledgehammer, so my preference is to use "hierarchical" names when it makes sense to do so, i.e. ComponentName_ElementName, which should ensure uniqueness, have no code or processing overhead, and even maybe be a bit easier on the eye?

---

3) Let me think about/look into unit tests. Thanks for the pointer -- it looks like a fair bit of reading! 😆

Are you thinking of something like making sure label-input linkages don't get broken?

Julusian:
An idea for this, perhaps the `id` field on our custom components should be `id: string | null` property, so that it has to be defined (but still optional) as that will make it a requirement to at least think about setting it (rather than accidentally forgetting)

And/or maybe there is an eslint plugin that can enforce setting these for basic components, and maybe that rule is configurable to allow more element types to consider.

Just thinking of ways to minimise the risk of new code forgetting to do this

PR #4070, opened 2026-04-03
chore: fix html DOM errors

arikorn:
In Dev mode, the console lists numerous errors about `<div> can't be nested within <p> elements` and this is also true for other block-level elements inside `<p>`. This PR fixes all the ones I've found so far (using a regex search in VS Code). 

There's also issues with nested `<form>` elements addressed here, though maybe not exhaustively, since the nesting isn't searchable.

Note that these error messages are suppressed in production mode, but the problem still is there.

original: very minor html error: `li` should be in `ul`, not `p`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the disconnected-companion help text to render as a proper list for better readability and accessibility.
  * Modal close now clears focus to avoid lingering focus/interaction issues after closing.

* **Improvements**
  * Standardized spacing and markup around buttons, labels and descriptive blocks for more consistent layout.
  * Adjusted recording label/control spacing for cleaner alignment.
  * Stabilized input initialization and modal/grid-size editing flow to prevent input glitches.
  * Added safe link attributes for external fixture downloads.


arikorn:
note updated title and OP: I've increased the scope of issues resolved here. 

Nested forms may be a bit trickier. ~~Will add them soon.~~ All done now.

arikorn:
@Julusian , this PR is ready for review and/or merging now. Thanks!

webui/src/UserConfig/Sections/GridConfig.tsx
```diff
@@ -80,25 +81,25 @@ const GridSizeModal = observer<object, GridSizeModalRef>(
 		const doClose = useCallback(() => {
 			setShow(false)
 
-			// Delay clearing the data so the modal can animate out
+			// Delay clearing the data so the modal can animate out (otherwise React complains about changing a controlled input to be uncontrolled)
```
Julusian: the purpose of this sleep was purely to avoid visual 'flicker' (data unloading) while the user can still see the form. otherwise it looks messy having the new values be forgotten briefly

arikorn:
@Julusian I had to downgrade eslint in order to merge on my side (I was doing it the same time as you did it here). Not sure why it's barfing since `styleText` should be there. (fnm --list shows: * v22.22.2 default, lts-jod)

> TypeError: util.styleText is not a function
    at util.styleText.validateStream (E:\ari\Javascript\companion\node_modules\eslint\lib\cli-engine\formatters\stylish.js:22:9)
    at E:\ari\Javascript\companion\node_modules\eslint\lib\cli-engine\formatters\stylish.js:67:16
    at Array.forEach (<anonymous>)
    at module.exports (E:\ari\Javascript\companion\node_modules\eslint\lib\cli-engine\formatters\stylish.js:55:10)
    at Object.format (E:\ari\Javascript\companion\node_modules\eslint\lib\eslint\eslint.js:1284:12)
    at printResults (E:\ari\Javascript\companion\node_modules\eslint\lib\cli.js:118:33)
    at async Object.execute (E:\ari\Javascript\companion\node_modules\eslint\lib\cli.js:478:4)
    at async main (E:\ari\Javascript\companion\node_modules\eslint\bin\eslint.js:175:19)
husky - pre-commit script failed (code 1)

webui/src/UserConfig/Sections/GridConfig.tsx
arikorn: Ah... this is the consequence of rabbit suggesting I be more efficient.
The React problem happens when you click Save and was originally put there, but the rabbit suggested that I "reuse" the close code. I'll clarify the comment. (Done.)

Would it be better to separate the code out, if the needs diverge?  (Marking resolved for now.)

arikorn:
working on the windows lint issue. should be able to fix shortly

arikorn:
It turns out I had manually installed node 18 before eventually switching to fnm. This masked a configuration omission until eslint 10 was installed, which requires node 20+, or so.  The problem was that husky doesn't use the path in the Powershell environment  One solution is to add fnm's location to the system path. A less intrusive solution is to add something to husky's init file either in ~/.config/husky or in the repo's .husky. A simple two line solution for the former:

(On windows, Open Git Bash)
```bash
mkdir -p ~/.config/husky
echo 'eval "$(fnm env --shell bash)"' >> ~/.config/husky/init.sh
```

On Windows this creates something like: `C:\Users\yourusername\.config\husky\init.sh`

Not sure whether manually adding to the path is better (more universal) or adding specifically for husky is better... Oh well, more stuff to add to the developers' docs once we resolve my outstanding PR there (presumably after 4.3 is released).

webui/src/UserConfig/Sections/GridConfig.tsx
```diff
@@ -79,26 +80,23 @@ const GridSizeModal = observer<object, GridSizeModalRef>(
 
 		const doClose = useCallback(() => {
 			setShow(false)
-
-			// Delay clearing the data so the modal can animate out
-			setTimeout(() => {
-				setNewGridSize(null)
-			}, 1500)
```
arikorn, line 86: Revisiting this topic:  After trying to reproduce the React errors I had an epiphany...(a) the rabbit "inadvertently" fixed the React problem in a better way than I had and (b) there's really no need to be resetting `newGridSize`

Details: on the one hand, it turns out that adding `?? ''` to all those `value:` fields ensures we get no React un/controlled type-change errors.  On the other hand, looking at the code, I really don't see why we need to clear `newGridSize` at all. There's already a `useEffect`, below, which fires on `show`, so just adjusting the `if` condition there ensures that the dialog is always up-to-date.

With that, there's no need to clear `newGridSize` and therefore no need for the `setTimeout` code (and no need for my comments, which are out of date already!).  So check out my latest commit: [07157e9](https://github.com/bitfocus/companion/pull/4070/commits/07157e9cb20c497300fc691d473f8b935923ebe3) (or just look here)

It passed my tests and keeps things simple. (Testing: open dialog, close it. Change the grid size in the Wizard or grid view. Now reopen grid dialog and confirm that it's up-to-date.)

webui/src/UserConfig/Sections/GridConfig.tsx
```diff
@@ -79,26 +80,23 @@ const GridSizeModal = observer<object, GridSizeModalRef>(
...
 				setShow(false)
-				setNewGridSize(null)
-
-				if (!newGridSize) return
+				if (!newGridSize) return // this should be impossible in the callback!
 
-				console.log('set gridSize', newGridSize)
+				if (import.meta.env.DEV) {
+					// note: newGridSize is a proxy object and doesn't print much on its own, so stringify it.
+					console.log('set gridSize', JSON.stringify(newGridSize))
```
Julusian: this whole if can probably be removed.

but for future reference, to deal with the proxies, there is a `toJS(` helper in mobx that will convert it to a plain object

PR #4084, opened 2026-04-11
fix: default destination page to match source on import

tonypiper:
## Summary

- When importing, the destination page now defaults to match the source page number rather than always defaulting to page 1
- For single-page imports: uses `snapshot.oldPageNumber` if valid, otherwise defaults to "new page" (-1)
- For full imports: destination syncs to the selected source page, updating as the user changes source page
- Renamed `importPageNumber`/`setImportPageNumber`/`changeImportPage` to `sourcePageNumber`/`setSourcePageNumber`/`changeSourcePage` for clarity

## Why

The previous default of page 1 was almost always wrong — if you export page 21 and re-import it, you want to restore it to page 21, not overwrite page 1. Defaulting to "new page" when the source page doesn't exist in the destination is safer than silently overwriting page 1. The same logic applies to full imports where the destination should follow the source page selection.

## Reviewer feedback addressed

Per @arikorn's review:
1. Extended destination-follows-source to full imports via `useEffect`
2. Renamed ambiguous `importPageNumber` vars to `sourcePageNumber`
3. Removed unnecessary `useMemo` and `!isSinglePage` guard from `defaultDestinationPage`

## Test plan

- [ ] Export a page that is not page 1 (e.g. page 21)
- [ ] Import the file → verify destination defaults to page 21 (or "new page" if page 21 doesn't exist)
- [ ] Verify the destination page can still be changed freely
- [ ] Import a full config → verify destination matches source page 1
- [ ] Change source page on full import → verify destination follows
- [ ] Import a full config, select a source page beyond the destination page count → verify destination defaults to "new page"

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/webui/src/ImportExport/Import/Page.tsx
+++ b/webui/src/ImportExport/Import/Page.tsx
@@ -66,12 +66,14 @@ export const ImportPageWizard = observer(function ImportPageWizard({
 		}
 	}, [snapshot.pages, snapshot.page, isSinglePage])
 
-	const { pageNumber, setPageNumber, changePage } = usePagePicker(pages.data.length, 1)
 	const {
 		pageNumber: importPageNumber,
 		setPageNumber: setImportPageNumber,
 		changePage: changeImportPage,
 	} = usePagePicker(pageCount, 1)
+
+	const defaultDestinationPage = isSinglePage ? (snapshot.oldPageNumber ?? 1) : importPageNumber
+	const { pageNumber, setPageNumber, changePage } = usePagePicker(pages.data.length, defaultDestinationPage)
 
 	const setConnectionRemap2 = useCallback(
 		(fromId: string, toId: string) => {
--- a/webui/src/ImportExport/Import/Page.tsx
+++ b/webui/src/ImportExport/Import/Page.tsx
@@ -72,7 +72,14 @@ export const ImportPageWizard = observer(function ImportPageWizard({
 		changePage: changeImportPage,
 	} = usePagePicker(pageCount, 1)
 
-	const defaultDestinationPage = isSinglePage ? (snapshot.oldPageNumber ?? 1) : importPageNumber
+	const defaultDestinationPage = useMemo(() => {
+		if (!isSinglePage) return importPageNumber
+		const oldPage = snapshot.oldPageNumber
+		if (typeof oldPage === 'number' && Number.isInteger(oldPage) && oldPage >= 1 && oldPage <= pages.data.length) {
+			return oldPage
+		}
+		return -1
+	}, [isSinglePage, importPageNumber, snapshot.oldPageNumber, pages.data.length])
 	const { pageNumber, setPageNumber, changePage } = usePagePicker(pages.data.length, defaultDestinationPage)
 
 	const setConnectionRemap2 = useCallback(
--- a/webui/src/ImportExport/Import/Page.tsx
+++ b/webui/src/ImportExport/Import/Page.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useContext, useMemo, useRef } from 'react'
+import { useCallback, useContext, useEffect, useMemo, useRef } from 'react'
 import { CButton, CCol, CRow, CFormSelect, CCallout } from '@coreui/react'
 import { MyErrorBoundary } from '~/Resources/Error'
 import { ButtonGridHeader, PageNumberPicker, type PageNumberOption } from '~/Buttons/ButtonGridHeader.js'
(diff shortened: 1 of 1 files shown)


webui/src/ImportExport/Import/Page.tsx
```diff
@@ -66,13 +66,22 @@ export const ImportPageWizard = observer(function ImportPageWizard({
...
 		changePage: changeImportPage,
 	} = usePagePicker(pageCount, 1)
 
+	const defaultDestinationPage = useMemo(() => {
+		if (!isSinglePage) return importPageNumber
+		const oldPage = snapshot.oldPageNumber
+		if (typeof oldPage === 'number' && Number.isInteger(oldPage) && oldPage >= 1 && oldPage <= pages.data.length) {
+			return oldPage
+		}
+		return -1
```
arikorn: Nice suggestion @tonypiper ! Some comments

1. This will default to a new page (-1) rather than the first page (1), as was the pre-existing behavior. Is that your intention? This actually makes sense when `oldPage` is beyond `pages.data.length`, but otherwise it should probably default to 1.
2. `useMemo` here seems a bit of overkill: you're just doing a couple of logical tests and all it's doing is setting a default, so it's a no-op if any of the dependents change, anyway. Suggestion: either it should be done only on mount or just take it out of a useMemo.
3. Technically the `!isSinglePage` test isn't necessary, since full won't have an oldPageNumber, and really, if there is an oldPageNumber it should probably be used.

arikorn:
@tonypiper two general comments:

1. It seems to me that your suggestion would be just as useful on "full" imports as well (either way it only affects the "Page" tab). You just need to add a call to `setPageNumber` to set it either to `importPageNumber` or -1 immediately after your `usePagePicker` line. So if the user chooses page 3 for the source page, it would change destination page to the same (or -1 if beyond the end).
2. As long as you're touching this file, I'd recommend changing importPageNumber/setImportPageNumber/changeImportPage to use "sourcePage" for clarity, since "import" is ambiguous (is it the page you're importing from or importing to?!).  In the source code just refactor using F2  (if using VS Code) on lines 71-73 of the original file. (maybe 70-72 now?). Or ask Claude.

webui/src/ImportExport/Import/Page.tsx
tonypiper: Thank you for the quick and helpful feedback. Defaulting to a new page was intentional - original motivation was that I have important buttons on page 1 and they kept getting overwritten. It seems to me that it's safer to create a new page (and then move it manually if need be) than overwrite an existing page if we can't be sure that it's the right page to overwrite.

webui/src/ImportExport/Import/Page.tsx
```diff
@@ -66,13 +66,24 @@ export const ImportPageWizard = observer(function ImportPageWizard({
...
+	const pageOrNew = (page: number | undefined): number =>
+		typeof page === 'number' && Number.isInteger(page) && page >= 1 && page <= pages.data.length ? page : -1
+
+	const validOldPage = pageOrNew(snapshot.oldPageNumber)
+	const defaultDestinationPage = validOldPage !== -1 ? validOldPage : pageOrNew(sourcePageNumber)
+	const { pageNumber, setPageNumber, changePage } = usePagePicker(pages.data.length, defaultDestinationPage)
+
+	useEffect(() => {
+		if (isSinglePage) return
+		setPageNumber(pageOrNew(sourcePageNumber))
```
arikorn, line 84: It's looking much better now, @tonypiper. One comment here: `useEffect` here is in inefficient since it adds a  render cycle and it somewhat hides the intent, which is that it's specifically in response to a user action. Better  to wrap `setSourcePageNumber` and `changeSourcePage` in `useCallback`s for use by lines 136-137,  which set both source and dest. (yeah, sorry my original suggestion wasn't quite right) Also, again, you don't need to protect for `isSinglePage`.

Another thing, which is mainly for clarity: instead of `usePagePicker(pageCount, 1)` online 73, determine `validOldPage` first and set the default page to `Math.abs(validOldPage)` or if that horrifies you `validOldPage === -1 ? 1 : validOldPage`, then you can get rid of `defaultDestinationPage` and just use `sourcePageNumber` in the second `usePagePicker`, which makes both of the calls clearer (to me).

Anyway, think about that second suggestion and question it, if you disagree!

Julusian:
I'm not 100% convinced by some of the behaviour here. While the auto page changing will be useful in some cases, in others it will be very frustrating.

eg, if you want to import 1,2,3 to 15,16,17, then it requires jumping all the way through the list of pages after each one instead of simply hitting next on both.

sometimes I pick the destination page first, before the source page too.

Maybe this is solvable by having the destination page picker track that the user has changed it manually, and stop auto-adjusting when that happens? might need a way to indicate that and put it back to auto, not sure where that should go though

PR #4094, opened 2026-04-14
feat: convert value when toggling between expression and raw mode #4087

### LINKED ISSUE #4087: [BUG] Toggling expression mode on dropdowns with string values should preserve string value

### Describe the bug

When toggling to expression mode for a dropdown with a string value, the `id` value of the current selection becomes the expression value. However because that value becomes `value` and not `'value'` actions or feedbacks that are expecting specific key values break until it is edited to include the ''. At that point toggling the option out of expression mode returns an error for a value not found in the dropdown list. I think this could be cleaner and auto handled by companion.

### Steps To Reproduce

1. Feedback with a dropdown selection:

Image
2. Toggle to expression mode, now the feedback can't return a value because it is looking for a known key:

Image
3. Edit the expression to wrap the value in ' ':

Image

4. Now toggle out of expression mode, and the value can't be found in the drop down contents:

Image

### Expected Behavior

I think this makes for a clumsy experience when moving between non-expression and expression mode, and ideally companion would handle this behind the scenes better.

```markdown
- Companion Version: 4.3.0+9093
- OS: Win 11
- Browser: Firefox
```

### Additional context

Module using:
- Companion Module Base: 2.0.3

For reference the feedback callback from the above:

> 		callback: (feedback) => {
> 			sub('versions', feedback.id, self)
> 			if (!self.api) return null
> 			switch (feedback.options.field) {
> 				case 'major':
> 					return self.api.versions.major
> 				case 'minor':
> 					return self.api.versions.minor
> 				case 'muProcessor':
> 					return self.api.versions.muProcessor
> 				case 'type':
> 					return MatVersionType[self.api.versions.type]
> 				default: {
> 					const _exhaustiveCheck: never = feedback.options.field
> 					return _exhaustiveCheck
> 				}
> 			}
> 		},

Julusian:
closes #4087

When toggling a field between expression and value mode, this will now explicitly convert it between the expression string and the value.

If it encounters a plain value, the field will be switched quietly, but if it contains an expression doing anything more than just a literal value/object, then it will show a modal to require confirmation before replacing the expression the user has written

Not 100% happy with the styling of the modal, but its better than nothing:

image

## Summary by CodeRabbit

* **New Features**
  * Added an expression-to-value conversion modal that evaluates expressions and lets users apply the computed value when switching from expression to value mode.
  * Field editing now attempts to auto-convert plain expression literals to values and prompts when needed.

* **Tests**
  * Added tests covering extraction of plain expression values and cases where conversion is not possible.

### DIFF
--- a/shared-lib/lib/Expression/ExpressionParse.ts
+++ b/shared-lib/lib/Expression/ExpressionParse.ts
@@ -5,6 +5,7 @@ import jsepTemplateLiteral, { type TemplateLiteral } from '@jsep-plugin/template
 import jsepComments from '@jsep-plugin/comment'
 import { CompanionVariablesPlugin, type CompanionVariableExpression } from './Plugins/CompanionVariables.js'
 import { AssignmentPlugin, type AssignmentExpression, type UpdateExpression } from './Plugins/Assignment.js'
+import type { JsonValue } from 'type-fest'
 
 // setup plugins
 jsep.plugins.register(jsepNumbers)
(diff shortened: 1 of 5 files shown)


phillipivan:
Should the modal have 3rd option beyond cancel and use computed value? I feel like there should be; like a select / enter alternate value. But I don't know of the top of my head when one might want to use this.

thedist:
> I feel like there should be; like a select / enter alternate value

Wouldn't that just be cancelling and the user entering a different value?

Julusian:
I think this needs some more thought, and trial and error and I dont want to delay 4.3.0 for it. hopefully it can be done for 4.3.1

After looking into this more, there is some ambiguity on how it should behave:
* for a textinput of `abc $(abc:test)` going to an expression it is reasonable to expect `"abc $(abc:test)"`, but perhaps it would be expected to remain `abc $(abc:test)` (which is how the button text behaves today). 
* doing the reverse for a text field, maybe the user started typing before realising they were in expression mode
* For a number/checkbox, this very much does make sense
* For a dropdown without `allowCustom: true`, this would also make sense to do; but perhaps it should be a bit looser so that either `abc` or `"abc"` in the expression field will become `abc`
* For a dropdown with `allowCustom: true`, maybe different in some way?
* maybe I overlooked some types, perhaps for internal field types
* have I handled the case where the expression itself is invalid?

And I am not entirely happy with how this modal is looking, I am not sure how to present the bits of data in a friendly way. it should contain some amount of:
* the expression
* if the computed value is 'invalid'
  * the reason it is invalid
  * the invalid value?
  * the value that will be used when clicking 'use default value'
* if the computed value is 'valid'
  * any warnings returned by the validator
  * the sanitised value that will be used 

So I am thinking that a first iteration of this could skip touching textinput fields, but there is still some unknowns and testing that needs to be tackled for the rest and I don't want to delay 4.3.0 any longer than it already is

If anyone feels like doing some sketches/patches for the layout of this modal, I welcome the input; that is the part that I will feel blocked on shortly.

Julusian:
A note for me to return to; this was on develop for some graphics field and should be used for inspiration on transforms here

```
// Fetch the property wrapper
		const elementEntry = (element as any)[key] as ExpressionOrValue<JsonValue | undefined>
		if (!elementEntry) return false

		if (!elementEntry.isExpression && value) {
			// Make sure the value is expression safe
			if (element.type === 'text' && key === 'text') {
				// Skip, this is very hard to fixup perfectly
			} else if (typeof elementEntry.value === 'string') {
				// If its a string, it will need to be wrapped in quotes
				// This is not always good enough, but is better than nothing
				elementEntry.value = `'${elementEntry.value}'`
			} else if (typeof elementEntry.value === 'number' && key === 'color') {
				// If its a color number, it is nicer to have it as a hex string
				elementEntry.value = '0x' + elementEntry.value.toString(16)
			} else if (typeof elementEntry.value === 'boolean') {
				elementEntry.value = elementEntry.value ? 'true' : 'false'
			}
			// Future: this may want more cases
		} else if (elementEntry.isExpression && !value) {
			// Preserve current resolved value
			const lastDrawStyle = this.getLastDrawStyle()
			const lastDrawElement = lastDrawStyle?.elements.find((el) => el.id === id)

			if (key === 'enabled' && !lastDrawElement) {
				// Special case, element was presumably disabled
				elementEntry.value = false as any
			} else if (lastDrawElement) {
				// The element was found, copy the value
				elementEntry.value = lastDrawElement[key as keyof typeof lastDrawElement]
			}
		}
```

Julusian:
> Should the modal have 3rd option beyond cancel and use computed value? I feel like there should be; like a select / enter alternate value. But I don't know of the top of my head when one might want to use this.

Maybe there should also be an 'ignore' option skip any conversion and behave like today?

PR #4108, opened 2026-04-22
fix: docusaurus aspect ratio on large images

arikorn:
When viewing in Docusaurus (help), large images are limited in width but the aspect-ratio is not honored (this is apparently a CSS quirk, since Docusaurus specifies the height explicitly, the `aspect-ratio` style-prop is ignored).

This PR preserves the aspect ratio.

Example: whatsnew/v4-3-0.md  `New surfaces UI`

(Note the problem is specific to Docusaurus -- it looks fine with react-markdown.)

## Summary by CodeRabbit

* **Style**
  * Documentation images now preserve their original aspect ratios when scaled, preventing vertical distortion.
  * Centered images on their own lines continue to be supported; no other visual behaviors were changed.

### DIFF
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -308,13 +308,13 @@
 	font-size: 1.25em;
 }
 
-/* Docusaurus limits max-width to 100% but without the follow rule, the image will lose its aspect ratio if it is too wide */
+/* Docusaurus limits max-width to 100% but without the following rule, the image will lose its aspect ratio if it is too wide */
 .markdown img {
 	height: auto;
 }
 
 /* The following rule centers images on their own line. 
-    The reason for ":last-child" is to allow us to "inline" the image by placing an element after the image in the markdown text.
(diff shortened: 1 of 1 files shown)


Julusian:
strange, I cant see this issue. what browser are you using?

arikorn:
You're right. It must be something related to my unsubmitted video code.  Let me look into it and I'll close this once I figure it out*.  Inspecting the same image in a docusaurus page using a dev-build w/o that branch I see:

```css
.img_ev3q {
    height: auto;
}
```

which is exactly what this PR is doing more generally.

(*OTOH, my additional comment on the `img:last-child` rule is still useful --I could either just delete the unneeded rule or move the comment to the other branch...)

arikorn:
I was hoping to just rename the  branch but I guess there's no way to do it. (Renaming in github just deleted it)

arikorn:
Moved the remaining CSS comment to #4114

PR #4135, opened 2026-04-26
feat: let user choose what right-click does on sidebar

arikorn:
https://github.com/bitfocus/companion/discussions/4122

Default is current behavior: right-click gets the context-menu and modifier-right-click gets the system menu. This option allows the user to reverse that behavior by activating "Prefer System Menu".

image

## Summary by CodeRabbit

* **New Features**
  * Added a "Prefer System Menu" preference option. Users can now toggle between using the custom context menu and the system context menu, with their choice persisted across sessions. The toggle is accessible from the sidebar settings.

### DIFF
--- a/webui/src/Components/useContextMenuProps.tsx
+++ b/webui/src/Components/useContextMenuProps.tsx
@@ -21,6 +21,9 @@ export interface ContextMenuProps {
 	It also manages positioning and visibility, and must be passed to the ContextMenu component.
 
 	* menuItems: the menu description
+	* preferSystemMenu: default, false: right-click brings up the context-menu; holding down a modifier key on click brings up the system menu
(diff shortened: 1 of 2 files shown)


arikorn:
coderabbit wrote:
>The tooltip currently says "ctrl-right-click", but the hook actually accepts any of alt / ctrl / meta / shift as the modifier. On macOS, users may instinctively reach for ⌘ (cmd) — so it'd be friendlier to phrase it more generically. Totally optional, but might save a few support questions.

Perhaps someone could comment on the rabbit's second nit: would people understand "modfier" keys? It seems to be standard in technical circles, but for mere mortals?

Julusian:
As mentioned in #4151, I don't think this is a user friendly way to go.

It will leave users confused by the 'highjacking' of the context menu, and if toggled to 'prefer system menu' mode will make this pretty much undiscoverable unless users know how to find it

PR #4151, opened 2026-05-02
Fix: Limit scope of sidebar context menu

thedist:
This PR adjusts the new context menu added in #3938 by limiting the scope of the custom context menu to only appear when right-clicking the sidebar-toggler element.

Right-clicks anywhere else in the UI now return to the default browser context menu which has been the well-established behaviour for all versions prior to 4.3. With the vast majority of the Companion userbase (~85% at the time of writing this) still on versions <4.3 this fix will allow to maintain a consistent experience with this when they upgrade.

This addresses the some of usability concerns raised in discussion #4122 and gives us time to consider more suitable solutions such as if we want to includes the context menu on other elements, have settings options to control right click behaviour and if so where is appropriate for those settings to be, and any other discussion that may need to be had.

### DIFF
--- a/webui/src/Components/useContextMenuProps.tsx
+++ b/webui/src/Components/useContextMenuProps.tsx
@@ -48,7 +48,9 @@ export function useContextMenuState(menuItems: MenuItemProps[]): ContextMenuProp
 			setVisible(false)
 		}
 
-		const handleInsideClick = (e?: Event) => {
+		const handleInsideClick = (e: MouseEvent) => {
+			if (e.button === 2) return // Ignore right click, to avoid the trigger click closing itself
+
 			if (e && menuRef.current && menuRef.current.contains(e.target as Node)) {
 				// allow propagation just to be safe so actions will trigger regardless of calling order.
 				setVisible(false)
--- a/webui/src/Layout/Sidebar.tsx
+++ b/webui/src/Layout/Sidebar.tsx
@@ -608,7 +608,11 @@ const UnfoldTogglerAndVersion = observer(function UnfoldTogglerAndVersion({
 
 	return (
 		<div className="nav-link sidebar-header-toggler2">
-			<span className={classNames('nav-icon-wrapper', mobileMode ? 'd-none' : 'd-flex')} onClick={toggleUnfoldable} onContextMenu={onContextMenu}>
+			<span
+				className={classNames('nav-icon-wrapper', mobileMode ? 'd-none' : 'd-flex')}
+				onClick={toggleUnfoldable}
+				onContextMenu={onContextMenu}
+			>
 				<span className="nav-icon sidebar-toggler"></span>
 			</span>
 
--- a/webui/src/Layout/Sidebar.tsx
+++ b/webui/src/Layout/Sidebar.tsx
@@ -374,7 +374,6 @@ export const MySidebar = memo(function MySidebar() {
 				unfoldable={unfoldable}
 				narrow={tempNarrow || narrowMode}
 				setNarrow={narrowMode ? DontSetOrUnset : setTempNarrow}
-				onContextMenu={contextState.onContextMenu}
 			>
 				<ContextMenu {...contextState} />
 				<CSidebarHeader className="brand">
@@ -504,7 +503,7 @@ export const MySidebar = memo(function MySidebar() {
 				)}
 				{!narrowMode && (
 					<CSidebarHeader className="border-top d-flex sidebar-header-toggler">
-						<UnfoldTogglerAndVersion toggleUnfoldable={toggleUnfoldable} />
+						<UnfoldTogglerAndVersion toggleUnfoldable={toggleUnfoldable} onContextMenu={contextState.onContextMenu} />
 					</CSidebarHeader>
 				)}
 			</CSidebar>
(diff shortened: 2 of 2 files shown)


Julusian:
Not sure if its just me, but sometimes I am unable to get the menu to remain open.

It does seem to work if I hit just to the right of the icon, but not on the icon.  
Also a minor thing, but it seems that this doesnt work for the full area where the icon changes, some of that gets the browser menu

https://github.com/user-attachments/assets/8e8ab9bc-f216-4cf0-a2e3-0ca02e596df1

thedist:
I'm not able to reproduce your issue, for me in any of the elements area (not including margin) I'm able to right click and get the context menu
chrome_OZmY0yhOmm

I've just moved the context menu up to its parent so check to see if this area resolves your issue,
chrome_3AGcWyG8bs

If not we could always go the next parent but that would start to include things like the version number, and I'm not sure if that's really for the best right now.
chrome_rjgnweNFIv

arikorn:
I honestly see no merit in this PR. It hides the context-menu behind the very element that inspired making the context menu as a work-around in the first place -- i.e., the sidebar toggler, which is very hard to see.

More importantly, a simple solution was already introduced a week ago in PR #4135, which is much more compatible with the current behavior. (If one objects to offering a choice of preference, just delete the menu-item and set `preferSystemMenu` to true.)

As I've already said in the discussion, I do not agree that any action is currently needed but if the consensus is against me, let's at least do it in a way that minimizes inconvenience to the people who want to use the new context menu, especially considering that such a solution has already been offered.

thedist:
> More importantly, a simple solution was already introduced a week ago in PR https://github.com/bitfocus/companion/pull/4135, which is much more compatible with the current behavior.

A feature update would require v4.4.0, or wait until v5.0.0, with the former not something we've even discussed if we want a v4.4, and v5 only recently hit beta so could be some time away, as such as I don't think waiting to release such a PR to stable is a viable option.

Additionally it still has the issue of you're changing the default behaviour that has already been well established that right click brings up the standard browser right click menu. The PR may have added an option to change back to that but it's an opt-out solution rather than opt-in, meaning for the vast amount of users who are still on pre-4.3 even in your new PR you would still be changing the functionality they're used to and requiring extra user interaction to opt-out and revert back to how they had it.

> let's at least do it in a way that minimizes inconvenience to the people who want to use the new context menu

You keep trying to use the inconvenience to the people who use the context menu as an excuse to push your way, but seem to completely disregard the much larger userbase that still has right click being the browsers default menu that you're disrupting.

This PR I've put forward is a patch, it resolve some of the issues with the context menu and restores the functionality users are used to, it keeps the context menu available for those who wish to opt-in to its use without forcing it on anyone or requiring users to through extra steps to opt-out, As a patch it can be rolled out to stable soon and minimize the impact to those who both use v4.3 (already a limited set of users) and also utilize that menu, unlike delaying for a feature patch which would increase the chance of impacting those users.

This also gives time to find a more reasonable solution that resolves the issues brought up (which the PR just to add an opt-out button still does not resolve) and build that more complete solution for v5.0.0 which is in beta.

Julusian:
> Additionally it still has the issue of you're changing the default behaviour that has already been well established that right click brings up the standard browser right click menu

yeah this is what worries me. I know that I at times am doing some visual pattern matching when using things that I am familiar with. I will eventually read the menu options, but only after spending some time being confused or thinking I must have hit a special area and to try clicking elsewhere.

>  let's at least do it in a way that minimizes inconvenience to the people who want to use the new context menu,

As I think I said elsewhere, the problem I see is leaving this as-is inconveniences people who dont want/expect it.

> If one objects to offering a choice of preference, just delete the menu-item and set preferSystemMenu to true.

The problem here is that it makes this borderline undiscoverable. I would not think to ctrl/alt click to get a different context menu, especially when this is the only place it does it and is new.  
After using the prefer system toggle, I probably wouldnt manage to find the custom menu without reading the docs to figure out how to enable it (if I remember it exists)

That said I probably wouldnt discover this menu, but a low-effort solution to that is to make the whole expand toggle button in 5.0 be just for this menu (I do wonder if anyone actively uses that mode toggle enough to be bothered by that.  
Or maybe some more work needs to go into figuring this out, but this pr avoids reverting it and gives a non-confusing middle ground until we can figure out what the best solution is

Julusian:
> I'm not able to reproduce your issue, for me in any of the elements area (not including margin) I'm able to right click and get the context menu

I found my bug, the mouseup handler was firing on the newly opened menu.

actually looks like an existing bug, doesnt happen if the cursor sits in a rounded corner

arikorn:
@Julusian responded
> > If one objects to offering a choice of preference, just delete the menu-item and set preferSystemMenu to true.

> The problem here is that it makes this borderline undiscoverable. 

Hmm... so replace it with something that is truly undiscoverable (as you yourself admit regarding toggler use)? That makes no sense. The behavior in 4.3 that you're removing here is fully documented. If the only difference, as I've been proposing, is that now the user holds down a modifier to get it, that's easy to add to the docs and perhaps not even a bad change (i.e. it may well be the behavior we stay with in the end to maintain browser compatibility).

In addition, this PR completely removes the context-menu behavior for narrow windows (mobile-mode) which doesn't have a toggler. So not only is it now undiscoverable, it's also actually unavailable for those users.

Adding "prefer system menu" to advanced settings (and defaulting to true) is simple and preserves choice, making it opt-in. Even just hardcoding "prefer system menu" to true is a better option for the user and more discoverable than this PR, given the documentation, and better ease of use.

PR #4185, opened 2026-05-24
fix: OSC press command example

ApolloFops:
The example was implying that, when you use `/press` on a location over OSC, it would press and then hold down the button, when in reality it presses and then immediately releases the button (`/down` presses and holds down the button).

## Summary by CodeRabbit

* **Documentation**
  * Updated OSC control docs to clarify that the `/location/<page>/<row>/<column>/press` command performs an immediate press-and-release (no hold).
* **Documentation**
  * Updated in-app OSC protocol examples/descriptions to match the press-and-release semantics for button commands.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bitfocus/companion/pull/4185?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/user-guide/5_remote-control/osc-control.md
+++ b/docs/user-guide/5_remote-control/osc-control.md
@@ -39,7 +39,7 @@ Remote triggering can be done by sending OSC commands to port `12321` (the port 
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


Julusian:
Thanks!

PR #4191, opened 2026-05-26
feat: lazy render buttons

### LINKED ISSUE #4008: Optionally suppress button text expression evaluation if a page isn't being rendered

### Describe the feature

Having dug through issues, I perceive that there isn't one that aligns with the intent of this feature request.

In companion, my use case (one of near infinite, I acknowledge, hence feature vs. bug flag), I am implementing a surface that controls a VoiceMeeter instance on a headless server.  In my case, it's VM Potato (8x8).  I've a page for each source and bus and, thereon each, I have all of the others selectable to enable bouncing from one strip to another in an easy manner.  I've elected to express current levels on each on all of the strip/bus buttons.  The net result is some 304 buttons spread across some 20+ pages of 10x10 surface buttons (thus far).  With the example button text expression depicted below (admittedly expensive), with no conditionals, Companion ends up consuming an obnoxious amount of CPU cycles (N of 60% (an older (circa 2015-ish) box, admittedly)).  After some dialog with our AI brethren, it was suggested that I wrap the expensive the expressions in a conditional to see if the expression conditional would short circuit and suppress evaluation of the false branch.  Thankfully, the expression evaluation engine did so and resource consumption is again reasonable, deterministic (and intentionally controllable).

In the course of this endeavor, I perceive to have determined that all button text expressions are being evaluated even if they are on pages that aren't being rendered.  I find that an option to disable expression engine evaluation if a page isn't being rendered might be useful.  This is the topic of my feature request.

Expensive Expression wrapped in a conditional:

```
$(internal:surface_emulator_JmQSxl782YQvxzenPry_4_page) == $(this:page) ?
//expensive expression start
concat(
  getVariable(
    'internal',
    concat('page_number_', $(custom:VmpStripGrp)*10 + $(this:row)-1, '_name')
  ),
  "\n",
  round(parseVariables(concat('$(VMP:strip_', $(this:row)-1, '_l1)')), 0),
  "|",
  round(parseVariables(concat('$(VMP:strip_', $(this:row)-1, '_l2)')), 0),
  "\n"
)
// expensive expression end
: ""
```

(As an aside, is there a more broadly applicable (elegant) way to accomplish the conditional test (removing the specific reference to a surface)?  Not being a dev, and new to companion, I'm no where near competent enough...)

Thanks!

### Usecases

Being able to deterministically suppress evaluation of non-rendered button text expressions would be of potential use to keep host resource consumption under control on the hosting pc.

The potential set of use cases is broad (from my ignorant perspective... :)).  

Image

Image

Image

Julusian:
Closes #4008

Currently, whenever a button invalidates we re-render the png data-url for that button (unless we get a cache hit). This render gets used for the webui, emulator and cloud. If none of those pull that render, it goes unused.
For many systems, which are left running a streamdeck without the webui open, or the webui looking at a different page, this results in a lot of unnecessary button renders.
In 5.0, the streamdeck/surface bitmaps are rendered lazily due to not knowing the needed resolutions up front, so this makes the data-url flow consistent with that.

This PR avoids that cost by deferring the rendering of this data-url until it is first wanted.  
As a consequence, it does mean that the first use of it will take a little longer to load as it has to be rendered instead of being pre-renderer. But as these are async requests maybe a little more delay is fine and not really noticable? 

I do not yet know if this is a good compromise, there is still a chance I will change my mind on this (perhaps even after merging)

----

For comparison with a very small config on my dev machine, primarily with 2 flashing buttons.
Before: CPU: 0.7%, memory ~700mb
After: CPU: 0.1%, memory ~400mb (increases to 700mb once I open the button view

So it does appear to have a measurable performance improvement.

Opening the button grid view has always had a delay before things render (perhaps becasue I'm in dev mode), so I cant say whether it feels any slower

## Summary by CodeRabbit

* **Refactor**
  * Image rendering is now lazy and asynchronous to reduce blocking and improve responsiveness.
  * Bank, preview, and emulator surface updates use queued, debounced pipelines for smoother, more reliable image updates.
  * Image data URLs are generated on demand, lowering startup work, memory churn, and improving update consistency.

### DIFF
--- a/companion/lib/Cloud/Controller.ts
+++ b/companion/lib/Cloud/Controller.ts
@@ -3,7 +3,7 @@ import isEqual from 'fast-deep-equal'
 import nodeMachineId from 'node-machine-id'
 import { v4 } from 'uuid'
 import z from 'zod'
-import { xyToOldBankIndex } from '@companion-app/shared/ControlId.js'
+import { formatLocation, xyToOldBankIndex } from '@companion-app/shared/ControlId.js'
 import { CloudRegionState } from '@companion-app/shared/Model/Cloud.js'
 import type { ControlLocation } from '@companion-app/shared/Model/Common.js'
 import { stringifyError } from '@companion-app/shared/Stringify.js'
@@ -16,6 +16,7 @@ import type { ImageResult } from '../Graphics/ImageResult.js'
 import LogController from '../Log/Controller.js'
 import type { IPageStore } from '../Page/Store.js'
 import type { AppInfo } from '../Registry.js'
+import { ImageWriteQueue } from '../Resources/ImageWriteQueue.js'
 import { delay } from '../Resources/Util.js'
 import { publicProcedure, router, toIterable } from '../UI/TRPC.js'
 import { CloudRegion, RegionInfo } from './Region.js'
@@ -104,6 +105,8 @@ export class CloudController {
 	 * The initialized region handlers
 	 */
 	#regionInstances: { [region: string]: CloudRegion } = {}
+
+	readonly #updateBankQueue: ImageWriteQueue<string, [ControlLocation, ImageResult]>
 	/**
 	 * The state object for the UI
 	 */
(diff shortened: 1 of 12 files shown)


Julusian:
I'm thinking now of merging this, it can always be reverted if it turns out to be a bad idea

Julusian:
@coderabbitai 

> Performance: invalidation still eagerly builds buttonStyle via getDrawStyle() before the lazy/render-cache boundary

This is intentional. This should be cheap and we need to do this to get a snapshot of the render.

Q) how would we know what variables should cause a re-render if we havent computed the drawstyle and in the process determined what variables the drawing depends upon?

PR #4193, opened 2026-05-27
fix: Stop tracking special expressions that don't refer to any variables after their values have been computed

jswalden:
This is a followup optimization after #4065.

The value of a special expression only changes if the value of a variable it depends upon changes.  A special expression that refers to no variables, can never change.  Therefore, there's no need to keep tracking such expression to recompute it as variables change value.

Before this change, it looks like every single feedback was being continually tracked for `isInverted` special expressions, for all time, and same now for all actions for `storeResult` as of that feature landing.  This will minimize the former to exactly those feedbacks containing an `isInverted` that depends upon a variable, and the latter to only those actions that depend upon a variable -- the least tracking possible.

## Summary by CodeRabbit

* **Refactor**
  * Improved special-expression tracking and recomputation to reduce unnecessary work, speed up updates, and make lifecycle behavior more predictable.
* **Tests**
  * Added focused tests validating lifecycle state transitions and timed recomputation, ensuring expression evaluation and update notifications occur reliably and only once when expected.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bitfocus/companion/pull/4193?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/companion/lib/Controls/Entities/EntitySpecialExpressionManager.ts
+++ b/companion/lib/Controls/Entities/EntitySpecialExpressionManager.ts
@@ -12,24 +12,42 @@ import type {
 	SpecialExpressions,
 	UpdateSpecialExpressionValuesFn,
 } from './SpecialExpressions.js'
-import type { StoreResult } from './Types.js'
 
 export type CreateVariablesAndExpressionParser = (
 	overrideVariableValues: VariableValues | null
 ) => VariablesAndExpressionParser
 
 interface EntityWrapper {
+	/**
+	 * The entity containing the special expression being tracked.
+	 */
 	readonly entity: WeakRef<ControlEntityInstance>
-	needsProcessing: boolean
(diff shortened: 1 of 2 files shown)


Julusian (approved):
It would be great to have a test or two to verify that this works as expected.
Probably requires adding a method to be able to query the internal state of this class a little, but that is an acceptable tradeoff

jswalden:
> Probably requires adding a method to be able to query the internal state of this class a little, but that is an acceptable tradeoff

Turns out no, indexed access can evade TypeScript access restrictions so just needed to make `this.#specialExpressions` private rather than a private field.

Added miscellaneous tests, didn't want to be too overly tied to the exact timer dispatch so I had to do some manual looping in poor-man's state tracking.  Original patch is the same, all adjustments are in the second patch -- and aside from the rename all is in test code.

PR #4216, opened 2026-06-02
refactor: Look up variables by `${connectionLabel}:${variableId}` without enclosing `$()`

jswalden:
Having to look up variables using identifiers enclosed in variable syntax creates an impedance mismatch, because callers may naturally have access to `${connectionLabel}:${variableId}` forcing them to go out of their way to tack on `$()`.  Everything should just use the un-enclosed form.

I did my best to work through the actual callers when doing all this, to ensure I got everything, but there's enough sprawl to create a small chance I missed something in the process.

I presume you don't have future plans for those `injectedVariableValues` trailing arguments when all callers inject no variables -- but kept those changes separate for readability and just in case you actually do.

I expect this to be a prerequisite of building these builtin variables directly into parsing so they don't have to be fully computed in parser creation, but I haven't completed that subsequent work yet to be certain.  Seems like this mild increase in simplicity is desirable regardless.

## Summary by CodeRabbit

* **Refactor**
  * Normalized internal variable reference format and simplified expression-evaluation interfaces across the app; parsing and preview behavior remain unchanged for end users.

* **Tests**
  * Updated and added tests to validate the revised variable parsing, override semantics, and expression evaluation to ensure compatibility and prevent regressions.

### DIFF
--- a/companion/lib/Variables/Util.ts
+++ b/companion/lib/Variables/Util.ts
@@ -34,7 +34,7 @@ import LogController from '../Log/Controller.js'
 import type { VariablesBlinker } from './VariablesBlinker.js'
 
 // Everybody stand back. I know regular expressions. - xckd #208 /ck/kc/
-const VARIABLE_REGEX = /\$\(([^:$)]+):([^)$]+)\)/
+const VARIABLE_REGEX = /\$\((([^:$)]+):([^)$]+))\)/
 // Like VARIABLE_REGEX but allows $ in the variable name part, to support nested variables
 const REPLACE_VARIABLE_REGEX = /\$\(([^:$)]+):([^)]+)\)/
 
@@ -138,9 +138,10 @@ export function parseVariablesInString(
 			break
 		}
 
-		const fullId = matches[0]
-		let connectionLabel = matches[1]
-		let variableId = matches[2]
+		const fullReference = matches[0]
+		const fullId = matches[1]
+		let connectionLabel = matches[2]
+		let variableId = matches[3]
 
 		if (connectionLabel === 'internal' && variableId.substring(0, 7) === 'custom_') {
 			connectionLabel = 'custom'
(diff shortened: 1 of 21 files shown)


companion/lib/Variables/Util.ts
```diff
@@ -212,7 +213,7 @@ export function replaceAllVariables(string: string, newLabel: string, preserveLa
 			// ensure we don't try and match the same thing again
 			fromIndex = matches.index + fromIndex + 1
 
-			if (matches[2] !== undefined && !preserveLabels.has(matches[1])) {
+			if (!preserveLabels.has(matches[1])) {
```
jswalden, line 216: Technically this first conditional removal is not necessary, if the index is updated correctly.  But the reality is the condition is impossible.  Matches can only be `undefined` if the match conditionally doesn't exist when the capture is within an ignored conditional, e.g. `/a(b)?c/` when matching against `"ac"`.  But here, nothing about this match is conditional -- the overall regular expression applying is, but that's handled by `matches !== null`, and if `matches` then the variable-name part must match and must be a string.

companion/lib/Variables/Util.ts
jswalden, line 166: The bot is unhappy about using `internal:custom_*` here rather than `custom:*`.  It's possible it's not wrong about the concern; I sort of think it might not be wrong.  But I'm not certain enough to actively change this preexisting behavior.

companion/test/Variables/parse.test.ts
```diff
@@ -158,4 +158,41 @@ describe('variable parsing', () => {
 			variableIds: new Set(['abc:third', 'abc:nope']),
 		})
 	})
+
+	test('internal:custom_* -> custom:* mapping and rewriting', () => {
```
jswalden, line 162: I added this because I accidentally goofed it by introducing an invalid simplification when writing the patch.  😅  Figure there should be a test to detect that error.

Julusian (approved):
I'm not a big fan of removing the injestVariableValues parameters from everything, those are really there for some future proofing for when something needs it (perhaps some did in the past).  
But equally, I dont have a good argument for keeping them, so maybe this is fine

peternewman:
I assume this is just for internal code, rather than expression syntax type stuff? Even so, is there a scenario internally where you can now no-longer use a colon between two things/how does it decide whether to look them up as a connection-variable pair, or as two things separated by a colon?

jswalden:
It's entirely internal.  Unless I goofed something, which I don't think I have, nothing the user can see would be affected by this except minimal differences in performance for various things (like, in milliseconds or so, if that).

PR #4219, opened 2026-06-03
feat: add copy increment button action

Christoph111:
## Summary
- Add a Copy +N button-grid action for copying a source button to multiple targets with incrementing values.
- Add a settings modal to choose which numeric/text fields should be incremented and configure the step size.
- Add TRPC support and a shared control increment utility to discover and apply incrementable button fields.
- Cover the increment utility with focused tests for action options, feedback options, local variables, button labels, zero-padded values, and skipped expressions/colors.

## Verification
- `git diff --check`
- `yarn test --run companion/test/Controls/ControlIncrementUtil.test.ts`
- `yarn check-types`
- Commit hook: `eslint`, `prettier --check`

## Summary by CodeRabbit

* **New Features**
  * Added "Copy +N" to the control grid for copying controls while incrementing selected numeric fields with a configurable step, per-paste offset, and on-screen guidance.
  * New settings modal to choose which fields to increment, set the step (validated range), and select/deselect all.

* **Tests**
  * Added tests verifying discovery of incrementable fields, correct increment behavior (including preserved zero-padding), and that non-numeric/color/layout fields are ignored.

### DIFF
--- a/companion/lib/Controls/ControlIncrementUtil.ts
+++ b/companion/lib/Controls/ControlIncrementUtil.ts
@@ -1,0 +1,256 @@
+import type { SomeControlModel } from '@companion-app/shared/Model/Controls.js'
+
+type IncrementPathSegment = string | number
+type IncrementPath = IncrementPathSegment[]
+
+export interface ControlIncrementOption {
+	id: string
+	label: string
+	currentValue: string
+	valueType: 'number' | 'text'
+}
+
+interface InternalControlIncrementOption extends ControlIncrementOption {
+	path: IncrementPath
+}
+
+const numericTextRegex = /-?\d+(?:\.\d+)?/
+const numericTextGlobalRegex = /-?\d+(?:\.\d+)?/g
+const hexColorRegex = /^#?[0-9a-f]{6,8}$/i
+
+export function getControlIncrementOptions(controlJson: SomeControlModel): ControlIncrementOption[] {
+	return collectControlIncrementOptions(controlJson).map(({ path: _path, ...field }) => field)
+}
+
+export function incrementControlModelFields(
+	controlJson: SomeControlModel,
+	selectedFieldIds: string[],
+	incrementBy: number
+): SomeControlModel {
+	if (!selectedFieldIds.length || incrementBy === 0) return controlJson
+
+	const selectedFields = new Set(selectedFieldIds)
+	const candidates = collectControlIncrementOptions(controlJson)
+
+	for (const candidate of candidates) {
+		if (!selectedFields.has(candidate.id)) continue
+
+		const target = getTargetByPath(controlJson, candidate.path)
+		if (!target) continue
+
+		const currentValue = getContainerValue(target.container, target.key)
+		if (typeof currentValue === 'number') {
+			setContainerValue(target.container, target.key, currentValue + incrementBy)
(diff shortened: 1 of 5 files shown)


companion/lib/Controls/ControlsTrpcRouter.ts
```diff
@@ -176,6 +177,78 @@ export function createControlsTrpcRouter(
...
+
+				const fromControl = controlsMap.get(fromControlId)
+				if (!fromControl) return false
+				const controlJson = incrementControlModelFields(
+					fromControl.toJSON(true),
+					input.incrementFieldIds,
+					input.incrementBy
+				)
+
+				// Delete the control at the destination
```
Julusian, line 231: Here onwards could probably be replaced with `controlsController.importControl(toLocation, controlJson, true)` and check the return was a string for success.

webui/src/scss/_button-grid.scss
Julusian, line 1: I feel like some of this css shouldnt be necessary, it would be better to use existing util when possible. For the grid/layout stuff that should be possible, probably not for the badge

companion/lib/Controls/ControlIncrementUtil.ts
```diff
@@ -0,0 +1,256 @@
...
+		default:
+			return `Step ${String(setId ?? '?')}`
+	}
+}
+
+function formatOptionKey(key: string): string {
+	return key.replace(/^opt:/, '').replace(/_/g, ' ')
+}
+
+function isExpressionValueObject(value: unknown): value is { value: unknown; isExpression: boolean } {
```
Julusian, line 250: there is already a util for this

companion/lib/Controls/ControlIncrementUtil.ts
Julusian, line 1: I'm not liking how generically this is searching. As some of the label generation shows, this loses all the context on what everything is, which is resulting in the functions for generating the label based on the paths, and leaves parts of the labels as abstract/internal names that don't mean anything to the user.  
And I worry that as we adjust terminology or add functionality elsewhere, it will be very easy to forget about the label generation here resulting in them diverging

In my modules, a bunch of the options fields are named very poorly as they were chosen years ago but they arent user facing so there has been no reason to go through the effort of renaming them.

Julusian (changes requested):
Thanks for the PR, I like the idea behind this!

From a UX side:
* I feel like the placement of the button and calling it '+1' is not the most clear on what it does
* This should be added to the context menu that was recently added to the grid
* There needs to be a confirm step before replacing an existing button; I think it is too easy to forget it is still active and to accidentally replace a button you dont mean to.

image

* I think it would be good to have a quick explainer block of text at the top to explain what this does
* I would prefer the `Local variable 1 (me)` to be `Local varible: me` or perhaps it should be `$(local:me)` to show it in a familar way?
* It looks like it has picked up colour feedback values, which is weird. It is hard to tell what these feedback values are.
* As this is doing replacements in strings, it would be good to show a preview of what those values will become. Should it be incrementing every number or just the last found number?


I am wondering whether this should support adjusting any feedback/action/style option, or if it should be limited to just local variables. While I can see value in it being able to operate on anything, I think its going to either add a lot of complexity to be able to accurately filter and label what it is showing. 

This is how it shows for one of my pretty complex buttons:
image

PR #4240, opened 2026-06-14
chore: move module watcher from chokidar to fs

bryce-seifert:
This prevents EMFILE errors (same situation as) https://github.com/bitfocus/companion/pull/4068, that remerged after chokidar version update.

## Summary by CodeRabbit

* **Chores**
  * Updated developer-mode extra-module watching to use native Node.js file watching instead of an external dependency.
  * Improved reload triggering by filtering watched changes to relevant file types, ignoring `node_modules`, and debouncing per-module reloads to reduce noisy updates.
  * Added clearer error handling for module watching so watcher failures are reported more reliably.

### DIFF
--- a/launcher/main.js
+++ b/launcher/main.js
@@ -3,7 +3,6 @@ import os from 'node:os'
 import path from 'node:path'
 import url, { fileURLToPath } from 'node:url'
 import { getCurrentScope, init } from '@sentry/electron/main'
-import chokidar from 'chokidar'
 import debounceFn from 'debounce-fn'
 import electron, { app, BrowserWindow, dialog, ipcMain } from 'electron'
 import Store from 'electron-store'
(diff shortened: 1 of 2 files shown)


Julusian:
There are quite a few documented caveats for this api, some could maybe be a problem? 
https://nodejs.org/docs/latest/api/fs.html#caveats

In particular I am wondering about:
> If the underlying functionality is not available for some reason, then fs.watch() will not be able to function and may throw an exception. For example, [...] or host file systems when using virtualization software such as Vagrant or Docker.

I wonder what that means for WSL, as that is a virtual machine these days, or if anyone is using something like devcontainers on macos

bryce-seifert:
Good points / questions. I'll need to do some reading on those use cases, as I'm not entirely sure / have not tested outside of a Mac dev environment.

PR #4259, opened 2026-06-18
feat: REST API Prototype

yannisgu:
This is the first prototype for #4101. The idea of this PR is to implement one resource to finalize the open discussion points. 

## Implementation Scope
- All endpoints related to `connections`
- Authentication is mock implementation with 4 hardcoded tokens for testing (`cpn_admin`, `cpn_write`, `cpn_read`, `cpn_secrets`)
- The REST API is currently only enabled if not packed (e.g. in development), there is no UI yet to enable it in the packed version (reason for that is that the PR can be merged, even when the feature is not "done" yet)
- REST API Docs are created when running `yarn run build:rest-api-docs`. It is not integrated into the CI build, so that the docs are not published yet, but they can be viewed locally

## How to test

Open [http://localhost:8000/api/docs/](http://localhost:8000/api/docs/) where API calls can be executed interactively

### How should we handle versioning

**Option A:** 
By resource, e.g. `connections/v1`, `buttons/v1. Whenever there is a breaking change in the API of a resource, we would bump the version of this resource only. New resources always start with V1. 
Note that this is related to the API only. So when we introduce `connections/v2`, we could keep `connections/v1` for compatibility.

A similar approach could be achieved with `Accept: application/vnd.companion.v2+json` headers, but making it less transparent I think.  

**Option B**
Include the Companion major version in the api path, e.g. `api/v4/connections`. 
Personally I'm not a fan of this option, as every major companion release would break API paths, despite most endpoints probably stay the same

**Option C**
Global API versioning, so start with `api/v1`, so that when one day we want to change the structure of the api completly (imaginge in a few years REST is not "best practice" anymore and we want to completly change the API structure). 
I guess in this case we would imply that API endpoints with the same path can be breaking across major companion version

**Option D**

Completely omit versioning for now. 

### Format of "one" resource

In the discussions there were voices to not have endpoints for an individual resources (e.g. `connections/:id`), instead having `connections` endpoint which can be filtered by id, always returning a collection. Personally I don't like that, I feels wrong, not how HTTP works,  but eventually it is just a matter of taste. The Companion Core team should reach a decision in this matter and I will implement it accordingly.

## Summary by CodeRabbit

* **New Features**
  * Added a Bearer-token, scope-protected REST API with interactive docs (`/api/docs`) and OpenAPI spec (`/api/openapi.json`).
  * Added `/api/connections/v1` endpoints for connection management, including config/secrets handling and restart.
* **Behavior Changes**
  * REST API availability is now controlled by `rest_api_enabled` and is enabled only when not running in a packaged build.
  * Connection PATCH can partially update config keys without replacing the full config.
* **Documentation**
  * Added REST API documentation pages (authentication and scopes, including secrets access).
* **Tests**
  * Added end-to-end REST API, OpenAPI, and UIExpress routing tests.


companion/lib/Service/RestApi/RestApiService.ts
```diff
@@ -0,0 +1,36 @@
...
+	readonly #logger = LogController.createLogger('Service/RestApi')
+	readonly tokenStore: RestApiTokenStoreMemory
+
+	constructor(instanceController: InstanceController, userconfigController: DataUserConfig, express: UIExpress) {
+		this.tokenStore = new RestApiTokenStoreMemory()
+
+		if (!userconfigController.getKey('rest_api_enabled')) {
+			this.#logger.info('REST API is disabled (set rest_api_enabled and restart to enable)')
+			return
+		}
```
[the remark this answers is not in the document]
yannisgu: I would keep it like that, for the time REST API is in active development

companion/lib/Service/RestApi/RestApiTokenStore.ts
```diff
@@ -0,0 +1,40 @@
...
+		token: 'cpn_secrets',
+		scopes: ['read', 'write', 'secrets'],
+	},
+	cpn_admin: {
+		id: 'static-admin',
+		name: 'Dev Admin Token',
+		token: 'cpn_admin',
+		scopes: ['admin'],
+	},
+}
```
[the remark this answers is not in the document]
yannisgu: Will add it if we add execute endpoints before creating UIs to manage tokens

tools/rest-api-docs/_category-intro.md
```diff
@@ -0,0 +1,5 @@
+The new REST API replaces the old [HTTP Remote Control](../http-remote-control.md) with a proper API that exposes most functions that can be performed in the UI.
```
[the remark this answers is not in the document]
yannisgu: As documentation will only be published when REST API is done, I would leave it like this

tools/rest-api-docs/authentication.md
```diff
@@ -0,0 +1,24 @@
...
+
+# Authentication
+
+The REST API uses **Bearer token** authentication. Include your token in every request:
+
+```
+Authorization: Bearer <token>
+```
+
+Tokens can be created in the admin UI (in development).
```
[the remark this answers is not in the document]
yannisgu: Let's keep the documentation for a future state, as the docs are only published when REST API will be fully implemented.

I added the secrets scope now manually

tools/rest-api-docs/authentication.md
```diff
@@ -0,0 +1,24 @@
...
+description: How to authenticate with the REST API
+---
+
+# Authentication
+
+The REST API uses **Bearer token** authentication. Include your token in every request:
+
+```
+Authorization: Bearer <token>
+```
```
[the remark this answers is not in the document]
yannisgu: done

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
```diff
@@ -0,0 +1,591 @@
...
+		if (config || secrets) {
+			const validationResult = await validateConfigAndSecrets(instanceController, connectionId, config, secrets)
+			if (validationResult.status === 'unavailable') {
+				next(RestApiError.conflict(validationResult.message))
+				return
+			}
+			if (validationResult.status === 'invalid') {
+				next(RestApiError.badRequest('Config validation failed', validationResult.errors))
+				return
+			}
```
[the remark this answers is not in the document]
yannisgu: Fixed by first updating the version

companion/lib/Service/RestApi/middleware/errorHandler.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Global error handler for the REST API.
+ * Converts RestApiError instances to structured JSON responses.
+ */
+export function restApiErrorHandler(
+	err: Error,
+	_req: Express.Request,
+	res: Express.Response,
+	_next: Express.NextFunction
+): void {
```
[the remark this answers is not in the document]
yannisgu: applied patch

companion/lib/UI/Express.ts
```diff
@@ -136,12 +137,19 @@ export class UIExpress {
...
+		// Use the router #restApiRouter for the new REST API, mounted before the legacy /api routes
+		this.app.use('/api', async (r, s, n) => this.#restApiRouter(r, s, n))
+
 		// Use the router #apiRouter to add API routes dynamically, this router can be redefined at runtime with setter
 		this.app.use('/api', async (r, s, n) => this.#apiRouter(r, s, n))
 
 		// Use the router #legacyApiRouter to add API routes dynamically, this router can be redefined at runtime with setter
 		this.app.use(async (r, s, n) => this.#legacyApiRouter(r, s, n))
 
+		this.app.use(['/api', '/trpc', '/socket.io'], (_req, res) => {
```
Julusian: arent we handling this already in some way?
at least  if I try to access /api in a browser I get a basic 404

docs/.gitignore
```diff
@@ -7,6 +7,7 @@
 # Generated files
 .docusaurus
 .cache-loader
+/user-guide/5_remote-control/rest-api/
```
Julusian, line 10: This will be a little problematic, as we auto-sync the /user-guide folder to the website, so it will be missing this folder

companion/package.json
```diff
@@ -86,6 +88,7 @@
     "semver": "^7.8.4",
     "socketcluster-client": "^20.0.1",
     "supports-color": "^10.2.2",
+    "swagger-ui-express": "^5.0.1",
```
Julusian: a note for myself: I don't like how unmaintained this is. Its a pretty thin wrapper around swagger-ui-dist, so it might be good to build our own wrapper.

I suspect it will need some fiddling to work in our esbuild packaging anyway

companion/lib/Service/RestApi/openapi.ts
```diff
@@ -0,0 +1,29 @@
+import { createRequire } from 'module'
+import { OpenApiGeneratorV3 } from '@asteasolutions/zod-to-openapi'
+import { registry } from './registry.js'
+import { registerConnectionPaths } from './routes/ConnectionsRouter.js'
+
+const require = createRequire(import.meta.url)
+const { version } = require('../../../package.json') as { version: string }
```
Julusian: this should use the appBuild or appVersion off `AppInfo` instead of reading the package.json

companion/lib/Service/RestApi/registry.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Global OpenAPI registry that collects all route and schema definitions.
+ * Each router module registers its paths here; the spec is generated from this at startup.
+ */
+export const registry = new OpenAPIRegistry()
+
+// Register the Bearer token security scheme
+registry.registerComponent('securitySchemes', 'bearerAuth', {
+	type: 'http',
+	scheme: 'bearer',
+	description: 'API token (e.g. cpn_admin, cpn_read)',
```
Julusian: probably shouldnt have examples here, as these are temporary values

companion/lib/Service/RestApi/RestApiRouter.ts
```diff
@@ -0,0 +1,45 @@
...
+		'/docs',
+		swaggerUi.serve,
+		swaggerUi.setup(openApiDocument, { swaggerOptions: { persistAuthorization: true } })
+	)
+
+	// Mount resource routers — each versioned independently
+	router.use(
+		'/connections/v1',
+		createAuthMiddleware(logger, tokenStore),
+		createConnectionsRouter(logger, instanceController)
```
Julusian: a discussion on arranging things; should we put the routers all together under the rest api, or put them next to each area to keep them msotly self-contained (like we have with TRPC)

Im sure it can be argued each way, I don't have a strong preference at this stage

companion/lib/Service/RestApi/schemas/connections.ts
```diff
@@ -0,0 +1,286 @@
...
+				'Module version id used by this connection, such as "1.2.0". Null means the newest compatible stable version is selected automatically, and may be returned while that version is being installed.'
+			)
+			.meta({ example: '1.2.0' }),
+		updatePolicy: z
+			.enum(InstanceVersionUpdatePolicy)
+			.describe('Module version update policy for this connection.')
+			.meta({ example: InstanceVersionUpdatePolicy.Stable }),
+		enabled: z.boolean().describe('Whether the connection is enabled and allowed to run.').meta({ example: true }),
+		sortOrder: z.number().describe('Sort position of the connection in the Companion UI.').meta({ example: 1 }),
+		collectionId: z.string().nullable().describe('Connection collection id, or null when not in a collection.'),
```
Julusian: Should these be in this response, or maybe there should be an endpoint which returns the structure/arrangement for all connections?

also worth noting that the sortOrder is within the collection, so not that useful without the full structure

same question probably goes on updating these values too. I've never really been sure on what is a good way to handle managing order

companion/lib/Service/RestApi/schemas/connections.ts
```diff
@@ -0,0 +1,286 @@
...
+			.optional()
+			.describe('Module version update policy to apply.')
+			.meta({ example: InstanceVersionUpdatePolicy.Manual }),
+		versionId: VersionIdSchema.nullable()
+			.optional()
+			.describe(
+				'Specific module version to use, such as "1.2.0". Omit to leave the current version unchanged. Use null to switch to the newest compatible stable version.'
+			)
+			.meta({ example: '1.2.0' }),
+		collectionId: z
```
Julusian: same thought as above. interestingly the sortOrder is missing in this schema

companion/lib/Service/RestApi/schemas/connections.ts
```diff
@@ -0,0 +1,286 @@
...
+	{
+		id: 'enableCameraControl',
+		type: 'checkbox',
+		label: 'Enable Camera Control',
+		default: false,
+	},
+]
+
+/** Schema for a raw config field definition in API responses */
+export const ConfigFieldResponseSchema = z
```
Julusian: This needs some work, to make it more understandable. I think that it should be the same style of type union we use internally.

Part of me wonders if this should be moved into shared-lib and replace the current interfaces with interfaces derived from zod schemas to avoid needing to duplicate this and keep it in sync. That would remove the isolation of it though, and perhaps make it a little harder to version?

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
```diff
@@ -0,0 +1,596 @@
...
+ * Create the connections router for /api/connections/v1
+ */
+export function createConnectionsRouter(logger: Logger, instanceController: InstanceController): Express.Router {
+	const router = Express.Router()
+
+	/**
+	 * GET /connections — List all connections with config + status
+	 */
+	router.get('/', requireScope('read'), (req, res, next) => {
+		const token = (req as any).apiToken as ApiToken
```
Julusian: would be nice to do this in a way typescript is happy with

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
```diff
@@ -0,0 +1,596 @@
...
+		const includeConfig = req.query.include_config === 'true'
+		const includeSecrets = req.query.include_secrets === 'true'
+
+		if (includeSecrets && !includeConfig) {
+			next(RestApiError.badRequest("Query parameter 'include_secrets' requires 'include_config=true'"))
+			return
+		}
+
+		if (includeSecrets && !hasScope(token.scopes, 'secrets')) {
```
Julusian: I think we should have a scope for accessing any connection config, as not all that many use secrets currently.

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
```diff
@@ -0,0 +1,596 @@
...
+
+		const connections = Object.entries(clientConnections).map(([id, config]) => {
+			const status = instanceController.getInstanceStatus(id)
+			const instanceConfig = includeConfig
+				? instanceController.getInstanceConfigOfType(id, ModuleInstanceType.Connection)
+				: undefined
+			return buildConnectionResponse(id, config, status, instanceConfig, includeSecrets)
+		})
+
+		res.json(collectionResponse(connections, { total: connections.length, limit: connections.length, offset: 0 }))
```
Julusian: should this endpoint include the full connection object, or just the bare minimum, and require `router.get('/:connectionId'` to get the detailed object?

either way is valid, just wanted to qeustion it

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
```diff
@@ -0,0 +1,596 @@
...
+			})
+
+			// Re-fetch from client JSON so response goes through the same path as GET
+			const clientConnections = instanceController.getConnectionClientJson(false)
+			const config = clientConnections[id]
+			const status = instanceController.getInstanceStatus(id)
+			const instanceConfig = instanceController.getInstanceConfigOfType(id, ModuleInstanceType.Connection)
+			const response = buildConnectionResponse(id, config, status, instanceConfig)
+
+			logger.info(`REST API: Created connection "${label}" (${id})`)
```
Julusian: Ideally this logger would be created with a prefix/path/namespace of soemthing like 'Service/Rest/connection/v1' which would allow for this line to become
```
logger.info(`Created connection "${label}" (${id})`)
```

I assume the same is true for other log lines, I won't comment on any more

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
```diff
@@ -0,0 +1,596 @@
...
+		if (secrets) {
+			const token = (req as any).apiToken as ApiToken
+			if (!hasScope(token.scopes, 'secrets')) {
+				next(RestApiError.forbidden("Insufficient scope: requires 'secrets'"))
+				return
+			}
+		}
+
+		if (versionId) {
+			const versionInfo =
```
Julusian: Some of this seems quite long and like some helpers are missing in the classes being referenced. This is true of a few methods here.

Particualrly I am worried about drift when the trpc changes and this api needs to be updated to have the same behaviour

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
```diff
@@ -0,0 +1,596 @@
...
+
+		let fields: SomeCompanionInputField[]
+		try {
+			fields = await instance.requestConfigFields()
+		} catch {
+			next(RestApiError.conflict('Failed to retrieve config fields from module'))
+			return
+		}
+
+		const configFields = fields.filter((field) => field.type !== 'static-text')
```
Julusian: We shouldnt do this; this will hide useful data if this list is shown to users

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
```diff
@@ -0,0 +1,596 @@
...
+		res.json(successResponse(response))
+	})
+
+	/**
+	 * GET /connections/:connectionId/config-fields — Get module config field definitions
+	 */
+	router.get('/:connectionId/config-fields', requireScope('read'), async (req, res, next) => {
+		const { connectionId } = req.params
+
+		const clientConnections = instanceController.getConnectionClientJson(true)
```
Julusian: yeah this feels very repetetive of the trpc api..

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
```diff
@@ -0,0 +1,596 @@
...
+/**
+ * Create the connections router for /api/connections/v1
+ */
+export function createConnectionsRouter(logger: Logger, instanceController: InstanceController): Express.Router {
+	const router = Express.Router()
+
+	/**
+	 * GET /connections — List all connections with config + status
+	 */
+	router.get('/', requireScope('read'), (req, res, next) => {
```
Julusian: hmm.. is this type safe against the openapi?

Im guessing not seeing as the schema looks to be built separately down below.  
I would really like for these handlers to be typed fully

Maybe we could use or take inspiration from https://github.com/bitfocusas/api? That auto generates a json version of the schema just from the data when building the router (eg https://updates.companion.free/docs/json https://github.com/bitfocus/updates-api/)

Probably a bit more effort to generate the schema without spinning up the whole of companion, but Im sure its solvable

companion/test/Service/RestApi/OpenApiSpec.test.ts
```diff
@@ -0,0 +1,289 @@
...
+				scheme: 'bearer',
+			})
+		)
+	})
+
+	test('has global security requirement', () => {
+		expect(doc.security).toEqual(expect.arrayContaining([{ bearerAuth: [] }]))
+	})
+
+	describe('connection paths', () => {
```
Julusian: from a skim I cant really tell what these are testing

companion/test/Service/RestApi/OpenApiSpec.test.ts
```diff
@@ -0,0 +1,289 @@
...
+const require = createRequire(import.meta.url)
+const { version } = require('../../../package.json') as { version: string }
+
+const mockOptions = {
+	fallbackMockImplementation: () => {
+		throw new Error('not mocked')
+	},
+}
+
+describe('OpenAPI Spec Generation', () => {
```
Julusian: are there tests confirming that changes to the api follow semver? 
so that when we add new fields it ensures they are optional so that older clients still work

Julusian (commented):
I'm happy with the general shape of this.

I think there are a few design things to address, and a handful of tiny things to deal with before this gets merged

Oh, and I think Option A. But I am starting to wonder about C. I do worry that A will be too granular, and be a bit confusing on that front, but I also worry about potential churn with C as we won't coordinate breaking changes across releases well. I feel like too granular is the better compromise though.

docs/.gitignore
yannisgu: The idea would be to generate those auto-generated docs with `yarn run build:rest-api-docs` before publishing them to the website. 
Is that not feasable?

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
yannisgu: Would you propose to have this scope instead of the `secrets` scope?

companion/lib/Service/RestApi/openapi.ts
yannisgu: fixed

companion/package.json
yannisgu: refactored to use `swagger-ui-dist` directly

companion/lib/Service/RestApi/registry.ts
yannisgu: fixed

companion/lib/Service/RestApi/RestApiRouter.ts
yannisgu: Good input, makes kinda sense to split it by domain, is now re-organized

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
yannisgu: Refactored the implementation so that it is now typesafe

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
yannisgu: fixed

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
yannisgu: fixed

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
yannisgu: I think it can be handy to get all connections with all the details . 

If you want to have smaller answers , it is already possible with include_config=false

companion/test/Service/RestApi/OpenApiSpec.test.ts
yannisgu: implemented now. Tested it by removing a parameter, it gives such an error:

```
AssertionError: Breaking REST API contract changes:
- Removed parameter: /paths/~1connections~1v1/get/parameters/query:include_secrets: expected [ Array(1) ] to deeply equal []

- Expected
+ Received

- []
+ [
+   "Removed parameter: /paths/~1connections~1v1/get/parameters/query:include_secrets",
+ ]
```

companion/test/Service/RestApi/OpenApiSpec.test.ts
yannisgu: removed all these tests and replace it with a test that checks for backwards compatiblity

companion/lib/UI/Express.ts
yannisgu: yeah, sorry, that is completely unnecessary

companion/lib/Service/RestApi/schemas/connections.ts
yannisgu: now implemented as union type zod types, not touching the existing interfaces. Seems more stable to me, and with the current setup, typescript should yield errors when the schema and interfaces drift apart.

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
yannisgu: fixed

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
yannisgu: Tried to refactor in 28100efb6, seems anyway some "dancing" to have this layer in-between (`ConnectionOperations`) which works for TRPC and REST-API.

companion/lib/Service/RestApi/schemas/connections.ts
yannisgu: I really struggle to find a proper approach for this. 

For `GET` I guess the approach would be to have : 

- `/api/connections/v1` - flat list without collectionId or sortOrder, sorted by id?
- `/api/connections/v1/tree` - returns connections grouped by collection and sorted by sort order

`POST` would maybe accept `collectionId` and a `position` (position index inside  collection) or `beforeConnection` or `afterConnection`?

`PATCH` would not accept `collectionId` or `sortOrder`

Instead we could need `PATCH /api/connections/v1/move` that accepts `collectionId`  (optionaly) and something like `position` (position index inside  collection) or `beforeConnection` or `afterConnection`? Now that I'm writing this I question if it is worth to have `move`, or we can anyway incldue that in `PATCH`.


What do you think?

companion/test/Service/RestApi/OpenApiSpec.test.ts
yannisgu: Maybe I should mention, that when doing a release, where the REST API changed the `update-rest-api-contract` github workflow needs to be triggered manually to, put the reference openapi schema into the repo.

docs/.gitignore
Julusian: oh yeah I suppose that would be. something would need to go in https://github.com/bitfocus/companion/blob/main/.github/workflows/sync-docs.yml to do that.

that would need to be done now as the sync happens for every commit to main (under the beta version on the website)

or maybe something commented out? I agree it would be good to not publish until enabled in the builds, but Im worried about the docs system failing to build the website if the docs are missing. 
we could also wait and see until it is merged to see what breaks

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
Julusian: actually yeah. 

long term a secrets scope is probably a good idea, short-term it feels a bit flawed. so yeah lets not for now, it can be reintroduced later once there is a better reason to

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
Julusian: Thinking out loud to myself; maybe the best approach to keeping this 'simple' is to treat the rest api as the one api.  
Maybe with some muddling, the rest api endpoints could all be registered over trpc too.  

Very much not something to do in this pr, thatll be a fairly chunky refactor.

.github/workflows/update-rest-api-contract.yaml
```diff
@@ -0,0 +1,61 @@
...
+          yarn config set httpTimeout 100000
+          yarn install --immutable
+
+      - name: Update stable contract
+        run: |
+          yarn rest-api:update-stable-contract
+
+      - name: Run compatibility test
+        run: |
+          yarn vitest run companion/test/Service/RestApi/OpenApiCompatibility.test.ts --project companion
```
Julusian: I dont think this is needed. 

I think it better to have a pr opened which will report the failing tests, as that will be more visible than a manually triggered ci run

tools/dev.mts
```diff
@@ -29,6 +29,13 @@ if (!semver.satisfies(process.versions.node, nodeJsValidRange)) {
 
 let node: ChildProcess | null = null
 const nodeArgs: string[] = []
+const swaggerUiSourcePath = path.join(import.meta.dirname, '../assets/swagger-ui')
+const swaggerUiDevPath = path.join(import.meta.dirname, '../companion/dist/Service/RestApi/assets/swagger-ui')
```
Julusian: instead of this copying and watching, could the place which serves it use a check of `isPackaged()` and produce a different path?

companion/lib/UI/Express.ts
```diff
@@ -147,6 +148,9 @@ export class UIExpress {
 			res.redirect(301, `/instance${req.url}`)
 		})
 
+		// Use the router #restApiRouter for the new REST API, mounted before the legacy /api routes
+		this.app.use('/api', async (r, s, n) => this.#restApiRouter(r, s, n))
```
Julusian: I'm wondering if /api is the best place to mount this actually. I'm a little worried about it clashing with existing endpoints. 

It looks like we are currnetly using:
* custom-variable
* location
* surfaces
* variable
* connections

Some of which will clash with this new api, unless we explicitly try to avoid that. I'm not sure how best to resovle this though.  
I partly wonder if putting all this new stuff under `/api/v2`, which will give some weird paths `/api/v2/connections/v1/...`, but perhaps that is fine? And leaves us scope to completely rebuild this from scratch in the future too.

companion/lib/Registry.ts
```diff
@@ -310,7 +310,8 @@ export class Registry {
 			pageStore,
 			this.instance,
 			this.ui.io,
-			this.ui.express
+			this.ui.express,
+			this
```
Julusian, line 340: I don't like the `this` being passed in whole like this, I tried to untangle all of that not too long ago.

Leave it for now, I expect there is no 'good' option for this so itll be better for me to play and find the one I dislike the least

companion/lib/Service/RestApi/SwaggerUi.ts
Julusian, line 1: A follow up note for myself

companion/lib/Instance/Connection/ConnectionsRestApi.ts
```diff
@@ -0,0 +1,738 @@
...
+		response.config = instanceConfig.config ?? {}
+		if (includeSecrets) {
+			response.secrets = instanceConfig.secrets ?? {}
+		}
+	}
+
+	return ConnectionResponseSchema.parse(response)
+}
+
+const CONNECTIONS_API_BASE_PATH = '/connections/v1'
```
Julusian: I think I saw the place binding this router wasnt using this constant. probably should be

companion/lib/Instance/Connection/ConnectionsRestApi.ts
```diff
@@ -0,0 +1,738 @@
...
+	z.ZodType | undefined,
+	z.ZodType | undefined
+>
+
+type AnyRestEndpointDefinition = RestEndpointDefinition<
+	z.ZodType | undefined,
+	z.ZodType | undefined,
+	z.ZodType | undefined,
+	z.ZodType | undefined
+>
```
Julusian: seems like some common util that should be put in a utils file somewher?

companion/lib/Instance/Connection/ConnectionsRestApi.ts
```diff
@@ -0,0 +1,738 @@
...
+		params: connectionIdParam,
+	},
+	response: {
+		status: 204,
+		description: 'Connection deleted',
+	},
+	errorResponses,
+})
+
+const restartConnectionEndpoint = defineRestEndpointContract({
```
Julusian: maybe these should be inlined to their usages? to keep them closer to where they are used.

companion/lib/Instance/Connection/ConnectionsRestApi.ts
Julusian, line 1: perhaps for a follow up, it might be nice to break this into more files. This file feels pretty long already, and its only going to get longer. Not important to do in this pr

companion/lib/Service/RestApi/schemas/connections.ts
Julusian: I think a move endpoint like that is reasonable. I feel that gives a bit of a separation between the simple properties, and the positioning that needs more context awareness. 

actually perhaps that move should take an array of move operations, so that the reordering can be done in bulk? that gives more argument for it being separate (and of course complicates it to handle that implementation, especially regarding error behaviour)

Julusian (commented):
I think this was all pretty nit-picky, nothing significant to tackle.

Now that 5.0 is released, I am happy to merge this in the betas (what will be 5.1) pretty soon

companion/lib/Instance/Connection/ConnectionsRestApi.ts
yannisgu: Done

companion/lib/UI/Express.ts
yannisgu: done, moved to api/v2 - makes sense to me, hope for everyone also as well ;)

companion/lib/Instance/Connection/ConnectionsRestApi.ts
yannisgu: fixed

tools/dev.mts
yannisgu: fixed

.github/workflows/update-rest-api-contract.yaml
yannisgu: ok, removed it and added a docs file

companion/lib/Instance/Connection/ConnectionsRestApi.ts
yannisgu: Done

companion/lib/Service/RestApi/schemas/connections.ts
yannisgu: Ok , i implemented new `/tree` and `/move` endpoints

companion/lib/Service/RestApi/routes/ConnectionsRouter.ts
yannisgu: Removed `secrets` scope and added `connections` scope, that must be set additionaly to read or write

docs/.gitignore
yannisgu: From my understanding, the scripts just copies the files, that exists, so for now I would do nothing, when we are ready, we can add `yarn run build:rest-api-docs` to `sync-docs.yml`

yannisgu:
@Julusian all points should be addressed/fixed  now

Julusian:
I've merged this and adjusted when it is enabled, it now requires a `EXPERIMENTAL_ENABLE_REST_API=1` env var to be set (this can be done in a .env file when running from source), to make it a bit easier to enable/disable

PR #4260, opened 2026-06-18
feat: more expression math utils

Julusian:
I'm not 100% about any of this, so input is appreciated

## Release Notes

* **New Features**
  * Extended `log(v, [base])` to support optional base (defaults to natural logarithm)
  * Added `exp(v)`, `sqrt(v)`, and `pow(base, exponent)` expression functions
  * Added built-in `PI` constant support in expressions

* **Documentation**
  * Updated the expressions functions guide to reflect `log(v, [base])`, including `log10(v)`, and added references for `exp()` and `sqrt()`

### DIFF
--- a/docs/user-guide/4_expressions/functions.md
+++ b/docs/user-guide/4_expressions/functions.md
@@ -114,6 +114,12 @@ Calculate the square root of a number.
 
 eg `sqrt(9)` gives `3`
 
+**pow(base, exponent)**
+
+Calculate the base raised to the power of the exponent. Equivalent to the `**` operator, but available as a function for clarity and for those used to other languages.
+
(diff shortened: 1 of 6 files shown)


thedist (commented):
My only feedback is if it's worth considering a `pow(base, exponent)` Expression function too? It is somewhat redundant as we do have the `**` operator, but as that's not an operator native to all programming languages (some use `**`, some `^`, and some have no operator and require using a `Math.Pow` function or similar, such as C#), so I'm wondering if adding a `pow` function would make it easier for non-JS users to work with exponents?

Other than that question, this LGTM, and would be everything that should be needed when working on math involving volumes, or signal strengths, and things like that.

Issue #3284, opened 2025-02-13
Set brightness from a variable and not just %


dudman:
### Is this a feature relevant to companion itself, and not a module?

- [x] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the feature

I'd like to use a $variable as the % to set surface Brightness, this dosn't seem possible ?

### Usecases

I'd use this for changing the brightness nigh/day, and as part of screen saver.
Thanks

torowi:
It would be great if the current brightness value were also available as a variable.

dnmeid:
related #2345

Julusian:
This is supported in 4.3

Issue #3046, opened 2024-09-23
Variables for surface locked status


GarthDB:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

<img width="1506" alt="image" src="https://github.com/user-attachments/assets/bfb6f95e-c25c-4a21-b0b2-b993a6ecd4db">

I want a variable that shows whether a surface is locked (with a pin).

### Usecases

I have a USB relay that I use to lock my keyboard, and I'd like to turn it on and off when the companion surface is locked or unlocked.

If there was a surface lock status variable, I could use a trigger to turn on/off the USB relay or check the status from my home automation hub.

thedist:
> Additionally, it would be helpful to have an option to select the startup state of any particular surface (locked/unlocked).

This would be more suited to a separate feature request

ukiews:
Here you go: https://github.com/bitfocus/companion/issues/3405

Issue #1964, opened 2022-03-18
Ajust Brightnesse by TCP and/or by an automatic trigger


pierr975:
### Is this a feature relevant to companion itself, and not a module?

- [X] I believe this to be a feature for companion, not a module

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

Hi Everyone,
I usually use Stream Deck with windows, but as soon as I discovered the possibility to use Stream Deck with Raspberry Pi I started using it seen that it's much cheaper and companion has a lot more of possibilities, but a possibility that I couldn't find was Turning off the keys after a no time use, so is it possible to implement in the raspberry pi companio version a brightness control by TCP (or HTTP) and use it as an internal feature that could be added to those interval time events?
Sorry for any english mistakes, I'm a Brazilian Automation Engineer.

Julusian:
There already is an action to set the brightness of a streamdeck, so you can do it on a timer, I dont think it is possible to do over TCP/UDP/HTTP/OSC, but we will happily accept a PR which adds that

pierr975:
I couldn't find the action to set the brightness after a certain time as I would like to do, I'm not sure if this is available for the Raspberry Pi Version 2.2.0 or I'm just looking on the wrong page.
I trying to configure it in "Triggers" and then "Add a new Trigger" and I guess it should be available on the "Action" list, am I right?

lonebaggie:
Create two triggers 
First trigger is a timer after x seconds set brightness to a set level zero for off
second trigger after any key press resets the first trigger ( therefore resets the timer) and change brightness back to a set level say 100 

Therefore  after no activity for x seconds the brightness will dim

PR #4266, opened 2026-06-21
feat: Variable for surface brightness

FusselTV:
Adds a `surface_<id>_brightness` internal variable, alongside the existing
  name/locked/location/page ones, so a surface's current brightness is available
  to expressions and feedbacks.

  Use cases:
  - Show the current brightness on a button.
  - Day/night or screensaver style automations that react to brightness.
  - Pairs with a future relative "adjust brightness" action (e.g. a Stream Deck+
    dial to nudge it up and down), where this variable tracks the level.

  I also made the "Surface: Set to brightness" action persist its value. Before,
  it only pushed to the hardware, so the surfaces page (and this variable) showed
  a stale number after using the action. Now they stay in sync however brightness
  gets changed.

  Mostly mirrors the existing `surface_<id>_locked` variable (#3046). Live updates
  reuse the `surface-config` event since brightness is persisted config.

  Related to #1964 and #3284.

## Summary by CodeRabbit

* **New Features**
  * Added brightness variable for each surface, available for dynamic control and integration.

* **Bug Fixes**
  * Surface brightness settings are now properly saved and persisted when updated.

### DIFF
--- a/companion/lib/Internal/Surface.ts
+++ b/companion/lib/Internal/Surface.ts
@@ -109,6 +109,7 @@ export class InternalSurface extends EventEmitter<InternalModuleFragmentEvents> 
 		})
 		this.#surfaceController.on('group_page', () => debounceUpdateVariables())
 		this.#surfaceController.on('surface_locked', () => debounceUpdateVariables())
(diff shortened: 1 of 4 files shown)


Julusian:
Thanks!

PR #4274, opened 2026-06-22
feat: advanced expressions

### LINKED ISSUE #4187: foreach(), map(), for()/range(), while() function in expression

### Describe the feature

Did you thinks it's possible to have a `foreach()` function in expression système, like that:
```js
foreach(Element, (e, i) => {})
```
if Element is list, e is element of list and i is index, else if Element is string, e is every seperated chars of string and i is index too, else if Element is Object ( like "{key: "value"}" ), e is value and i is key, otherhise, foreach mad an error.

In same idea, the function `map()` can be made, like:
```js
let variable = map(Element, (e, i) => {return "anything"})
```
Same rules of `foreach(),` but can be return value to create a list.

Always in same idea, the function `for():`
```js
for(startValue, objectiveValue, step, (i) => {})
for(startValue, objectiveValue, (i) => {}) // step is 1 by default
for(objectiveValue, (i) => {}) // startValue is 0 by default
```
if step === 0 then error
if step < 0 && startValue < objectiveValue then error
if step > 0 && startValue > objectiveValue then error

That can be replace by a `range(),` because he can use like that: `foreach(range(...), ...)`
```js
range(startValue, objectiveValue, step)
range(startValue, objectiveValue) // step is 1 by default
range(objectiveValue) // startValue is 0 by default
```
Same rule of `for()` function, the advantage of `range()` functions: you can create list of number easier.

And for finish, the `while()` function:
```js
while(condition, () => {}) // check before loop
while(() => {}, condition) // check after loop
while(() => {return condition}) // internal condition, usefull for use variable defined inside loop in condition, check after loop
```
condition is just a boolean.

To use all of that, that would say add the possibility to made function in variable, like:
```js
let variable = () => {}
```
but that can create problem if is returned in expression, so I suggest to just add new format, and do not use all of that with function, example:
```js
for (let e, i in element) { // replace foreach()
    // content here
}
```
```js
for (let i = start; condition; incrementator) { // replace for()
    // content here
}
```
```js
while (condition) { // replace while() with check before
    // content here
}
while { // replace while() with check after
    // content here
} do (condition /* can be use variable defined inside while bracket ? */)
```

Except for function `range(),` I think its the only function need to keep that form

### Usecases

To many usage to personalize stream deck, I can speak about all of use.. Example:
in VDO ninja, you can have mutiple connection, you have 8 scene, to check if minimum of 1 person is connected in scene n°3 for example, use a foreach loop in every person, and if one is detected as connected in scene, made a variable to true (this variable is false as default). In that case, while is needed, because no variable with list of person connected exist, but in futur version or in other connections, if a variable exist with list of **anything**, a foreach loop can be use, and in another without list of **anything**, but with a variable with quantity of **anything** for can be used

In general, that really usefull for every connections customisation, not really for companion without connections

Julusian:
closes #4187

This aims to solve the requst of loops and proper if statements in expressions.
In doing this, I quickly hit the limit of what jsep is capable of. It can do arrow functions (to a limited degree), but everything would have had to be custom plugins.  
So before doing that, I looked to see what other parsers could bring. I settled on `acorn` for this.

As a consequence, there are some small changes in the parser behaviour, as the parser is now stricter and expecting true javascript.
1) `1 2 3` was valid before, with the result being `3`. This will now error as multiple statements on one line is not valid js

I think that is the main one that could catch someone out.  
I tried to avoid any known breaking changes, but this one was unavoidable.

----

Because of the potential for infinite loops, expressions now have an operation limit. After reaching this limit the expression will fail. This may want tuning to ensure it is high enough.

Some other js syntax is now supported which wasnt before. Nothing major, just newer language features (spread, optional chaining)

## Summary by CodeRabbit

* **New Features**
  * Added collection helper functions: `arrayMap`, `arrayFilter`, `arrayReduce`, `arrayForEach`, `arrayFind`, `arrayFindIndex`, `arraySome`, `arrayEvery`, `arraySort`, `arrayReverse`, plus `objectKeys` and `objectValues`.
  * Expanded expression scripting support (control flow, closures, and `let`/`const`), including optional chaining (`?.`) and nullish coalescing (`??`).

* **Documentation**
  * Added an “Advanced expressions” guide and expanded the expressions functions reference (including array iteration operations).

* **Bug Fixes**
  * Tightened execution budgeting for expression evaluation in visibility and surface matching, improving safety against runaway expressions.

* **UI**
  * Updated expression editor completions/keywords for the new built-in helpers and language constructs.


Julusian:
cc @thedist 

I havent tested this at all yet, but this is my pitch on how to handle loops, ifs and fleshing out expressions to be more of a language beyond simple calculations.

There are some potentially breaking changes though, due to the stricter parser. Hopefully they wont be a problem?

Any thoughts on any of this? (not the biggest rush to merge this, but this week would be good)

Julusian:
Some terrible demo of loops and statements:
https://github.com/user-attachments/assets/5ffbc8b2-87bc-48b5-a044-be68f3581998


Proving that infinite loops are 'safe':
image

thedist:
Looks good so far from my testing.

I did a little testing with loops and being able to go through a loop 1,000,000 times successfully seems reasonable to me and I would never come close in any scenario I can think of, but I have no clue the sort of silly things some users would be trying to do lol but I think the current limit is fine.

It breaking some existing expressions I'm not overly concerned with as I think generally it's uncommon to have multiple statements on the same line, especially without using a `;` to end each statement. The benefits of this are worth the inconvenience to users who unfortunately used poor programming practices, and we can do what we can when releasing the version this'll be included with to explain how to handle these changes if their expressions now error.

Julusian:
> go through a loop 1,000,000 times successfully seems reasonable 

perhaps it should be lowered? its an arbitrary number, hopefully causing rogue loops to not be a noticable performance drain.

I should probably mention that isVisibleExpressions are able to use loops too, but they have an much tighter 1000 limit

>  I think generally it's uncommon to have multiple statements on the same line

cool. that is my thinking too. it doesnt seem useful to do that, but could have been done by accident.


Then I'll give this a bit more of a test and merge it

PR #4309, opened 2026-07-12
Fix: deep cloning of preset to prevent erroneous changes

### LINKED ISSUE #4306: [BUG] Editing a preset on the grid also modifies the source preset in the Module Preset Library

### Describe the bug

When a preset is dragged from the Module Preset Library onto the grid and then modified (e.g. color or text), the change is also applied back to the preset entry in the Module Preset Library itself — instead of only affecting the button instance on the grid.
This becomes especially problematic when the same preset is dragged onto the grid a second time and modified again: the style-preview of the first button (Button A) then unexpectedly changes as well, even though it was never touched again.

### Steps To Reproduce

1. Drag a preset from the Module Preset Library onto the grid (Button A).
2. Change something on Button A, e.g. color or text.
→ The preset entry in the Module Preset Library now takes on the style from Button A.
3. Drag the same preset from the library onto the grid again (Button B).
4. Change something on Button B, e.g. color or text.
5. Look at the Style tab of Button A again

### Expected Behavior

Changing a button on the grid should only affect that specific button instance. The source preset in the Module Preset Library should remain unchanged, and previously placed buttons (like Button A) should not be affected by later changes made to other buttons (like Button B).

```markdown
- Companion Version: 5.0-stable
- OS: macOS 26.5.2
- Browser: Safari 26.5.2
```

### Additional context

_No response_

thedist:
Resolves #4306 

Simply wraps the `definition.model` in a `structuredClone` to prevent any modification of the button after a preset gets dragged on to it from impacting the original preset object that's being referenced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved preset handling by returning and generating preset-derived control models as independent deep copies.
  * Prevented overridden variable/local values from unintentionally mutating the original preset-backed model.

### DIFF
--- a/companion/lib/Instance/Definitions.ts
+++ b/companion/lib/Instance/Definitions.ts
@@ -404,10 +404,7 @@ export class InstanceDefinitions extends EventEmitter<InstanceDefinitionsEvents>
 
 		if (!variableValues) return structuredClone(definition.model)
 
-		const model: LayeredButtonModel = {
-			...structuredClone(definition.model),
-			localVariables: structuredClone(definition.model.localVariables),
-		}
+		const model: LayeredButtonModel = structuredClone(definition.model)
 
 		injectOverriddenLocalVariableValues(model.localVariables, variableValues)
 
--- a/companion/lib/Instance/Definitions.ts
+++ b/companion/lib/Instance/Definitions.ts
@@ -402,10 +402,10 @@ export class InstanceDefinitions extends EventEmitter<InstanceDefinitionsEvents>
 		const definition = this.#presetDefinitions[connectionId]?.get(presetId)
 		if (!definition || definition.type !== 'button') return null
 
-		if (!variableValues) return definition.model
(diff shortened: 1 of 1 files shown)


Julusian:
I dont think this is the right solution to this. 
I've been trying to avoid adding the clone here, because we end up doing a 'clone' during the ControlButtonLayered constructor.

I suspect theres another bug hidden around here, where importing a page twice will have the same issue as we expect that constructor to do the clone

I have a commit for this locally, didnt get around to pushing it. just cloning the value passed to `this.drawing.loadElements(` in two places

thedist:
I'll close this then as you're much more familiar than me in this area. I just wrote it as it was a quick fix to clone the model right after `const definition = this.#presetDefinitions[connectionId]?.get(presetId)` so that it's a deep clone at what I thought was the earliest point and so would be safe for any downstream functions that make use of it.

PR #4327, opened 2026-07-16
chore: ignore additional directories to avoid maxing out watched files

bryce-seifert:
Alternative approach from https://github.com/bitfocus/companion/pull/4240

In my testing being more restrictive on the ignored directories has fixed up the EMFILE watcher errors with large amount of modules, while keeping chokidar in place.

## Summary by CodeRabbit

- **Improvements**
  - Development file watching is now more focused on relevant build outputs, generated content, and key configuration/documentation files.
  - Irrelevant directories and unsupported file types are ignored more consistently, reducing unnecessary reloads.
  - Multiple development watchers now share the same filtering behavior for more stable, reliable development workflows.


Julusian:
I will look at this, but an idea I have been wondering about recently is maybe trying to ditch the `tsc` step and invoke `tsx watch main.ts` instead.

That would mean being able to stop the `tsc --watch` invocation, chokidar to handling just the dev-modules, and leaving tsx watch which I believe works based off the import graph rather than everyhthing on disk.

But this likely requires some tweaks to many build things to get it all happy (keeping shared-lib importable could be interesting)

Feel free to look into this, or I can look at this PR as it stands, Im happy either way

tools/dev.mts
```diff
@@ -119,25 +119,58 @@ concurrently([
...
+		[
+			// Only the compiled output can trigger a restart, so don't pay to watch the typescript sources
+			'../companion/dist',
+			'../companion/generated',
+			'../companion/package.json',
+			'../companion/tsconfig.json',
+			'../shared-lib/dist',
+			'../shared-lib/package.json',
+			'../shared-lib/tsconfig.json',
+			'../docs',
```
Julusian:
```suggestion
```
pretty sure we dont need the docs here, thats its own watcher/process

PR #4328, opened 2026-07-16
feat: expression oscillator function

### LINKED ISSUE #3961: expression oscillator function

### Describe the feature

As suggested in https://github.com/bitfocus/companion/pull/3831, it could be very beneficial to have an oscillate function in expressions.

I propose that we pick and run at some fixed update rate, perhaps 10hz. This should be a global clock, so that everything updates aligned.   
This will want to be done outside of the variables system, likely its own dedicated clock that triggers a check for all controls/entities that have an interest in this global clock, which internally may want to do some extra routing, depending on where we allow this.

Some thought will want to be put into how this should behave in feedbacks, as we risk spamming modules with a lot of updates to handle this. Perhaps we start out by not supporting it; if we do allow it then we will want to rework the variable update flow as otherwise this could be used in an expression variable, and introduce many more rapid variable invalidations.  

Does this even make sense for actions? It will certainly flood any modules using subscribe/unsubscribe and probably wont be useful?

### Usecases

Paired with the graphics overhaul, it could be used to achieve some more dynamic/self updating drawing more easily, and allow for some other cool things.

As suggested in #3831, this will probably want a bunch of parameters to choose the shape, speed and more.

Julusian:
Closes #3961

Github or my screen recording software makes this look more stuttery
https://github.com/user-attachments/assets/cacfda32-7110-4e31-b2a9-33e1ca91c653

## Summary by CodeRabbit

* **New Features**
  * Added the `oscillate()` expression function with configurable period, waveform, and phase.
  * Clock-sensitive expressions now update automatically in previews, graphics, feedback, and entity-based controls.
  * Added editor completion and documentation support for `oscillate()`.

* **Bug Fixes**
  * Improved caching and refresh behavior for dynamic expressions.
  * Fixed gauge markers rendering outside their gauge bounds.
  * Expressions used in one-time action options now reject clock-dependent functions.


Julusian:
@coderabbitai  review

PR #4341, opened 2026-07-20
feat: wysiwyg toolbar for basic edits to a button's style

bryce-seifert:
This adds a toolbar with some basic features to help speed up the styling process, without having to dive into the options / do math.
The toolbar contains:
- Center horizontally
- Center vertically
- Scale to fill
- Linked scaling
- Snapping (to edges, center, and other elements)
- Move to front
- Move to back

Right now, if an expression is enabled for any of the position/size elements, it disables the toolbar.

https://github.com/user-attachments/assets/cbb1f802-4de4-4cea-a706-88a119f7bac0

## Summary by CodeRabbit

* **New Features**
  * Added bulk layered-style option updates for multi-field editing.
  * Enhanced layered-button editing with auto-sizing, hit-testing, move/resize controls, snapping, and linked aspect-ratio locking.
  * Added quick actions for positioning, resizing, snapping, and z-order management.
  * Improved line editing with endpoint selection and manipulation.
* **Bug Fixes**
  * Improved update consistency by reducing intermediate changes.
* **Tests**
  * Expanded coverage for hit-testing, canvas sizing, bounds conversion, and snapping.


Julusian:
ooh nice!
I have also found the positioning to be a bit tedious and was thinking about looking at a subset of this.

Before I look at the code (or try to run it), my only thought is maybe this toolbar should be vertically along the left edge?   
Just that vertical space is at a bit of a premium in this page I find, and there is almost always deadspace on either side of the preview render
(Maybe that preview aspect ratio should move too for the same reason, but maybe not to the same toolbar. Not something for this PR; thinking out loud to myself)

bryce-seifert:
@Julusian Good thinking. I changed the toolbar to vertical. And, after that, the whole are started to all look a little off balance, so I did some experimenting with the aspect ratio selection + the simple/advanced toggle. Let me know what you think!

I also added a tooltip about why the tools get disabled, for various reasons.

https://github.com/user-attachments/assets/48f90580-fbc4-44bd-8d5d-07281147600b

Julusian:
I like it!

Only one concern with the aspect ratio, that we do want to allow custom aspect ratios, for testing for those other surface types that we dont have a builtin predefined button for. Maybe that could be a 4th button that opens a popover? 

Something vaguely like the grid zoom popover. Might be enough to have a dropdown in it, and nothing else?
image

----

For that basic/advanced toggle, perhaps instead of advanced it should be 'all properties' or something? I dont mind the current naming, just thinking out loud.

Julusian (approved):
Im happy with the code of this

Giving it a quick try, some UX thoughts (which may want to be follow ups):
* Lines dont appear to have drag markers for the start+end. Could they?
* The hitboxes are correct for the bounds, but make for slightly confusing selecting. for example, having a 100% height text box with a single line of text, it feels like I should be able to click around the text to select the background. But I expect this to be non-trivial to do some drawing aware hitbox detection.

bryce-seifert:
> For that basic/advanced toggle, perhaps instead of advanced it should be 'all properties' or something? I dont mind the current naming, just thinking out loud.

I like that naming better! I changed it to that, and also made the buttons not have conflicting colors with the separator:
Screenshot 2026-08-02 at 10 33 18 AM

> Only one concern with the aspect ratio, that we do want to allow custom aspect ratios, for testing for those other surface types that we dont have a builtin predefined button for. Maybe that could be a 4th button that opens a popover?

Good point. Added a custom icon, that accepts entering custom values for w/h.

Screenshot 2026-08-02 at 10 33 08 AM

> Lines dont appear to have drag markers for the start+end. Could they?

Yeah I had kinda punted on lines. But it feels right to include them. I added separate overlay indicators since they have different behaviors. Let me know what you think.

Screenshot 2026-08-02 at 10 32 52 AM

> The hitboxes are correct for the bounds, but make for slightly confusing selecting. for example, having a 100% height text box with a single line of text, it feels like I should be able to click around the text to select the background.

Yeah, I agree with this. I'm not quite sure how to tackle it yet though. Might something for a followup, or just needs some more brainstorming.

Julusian:
I've fixed the merge conflicts and was about ready to merge it, but I've found that the markers arent following rotation:
image

I also discovered `buildElementRects` which I didnt realise before, which is kind of replicating the bounds composition calculation logic the renderer is doing. Maybe the return value of `GraphicsLayeredButtonRenderer.draw` should be modified to return these instead? Hopefully plumbing those in to the logic that needs the rects will be simple enough (although perhaps a draw behind, which is hopefully fine)

----

Another thing to think about is #4381 which wants to add an aspect ratio option for the neo. I am not sure how to handle that, I dont want to have too many options on the bar for everyone and this may want a bigger solution out of scope of this pr (surface driven presets shown within the popover?)

peternewman:
> I've fixed the merge conflicts and was about ready to merge it, but I've found that the markers arent following rotation: image

I couldn't obviously see the code for this bit in the PR, but do the current aspect ratios have tooltips to say what they're for? Actually I've just spotted it looks like they're built from the current ASPECT_RATIO_OPTIONS data.

> Another thing to think about is #4381 which wants to add an aspect ratio option for the neo. I am not sure how to handle that, I dont want to have too many options on the bar for everyone and this may want a bigger solution out of scope of this pr (surface driven presets shown within the popover?)

I'd agree we don't want a never-ending list of options on the bar, and personally I'm not convinced having more they'd probably stop being clear what they're for, but could there be some sort of pop-up/drop-down to let you see the original/full list (what does the pencil currently do for example) as from a quick look the Loupedeck should be/can be 90x60, so that's yet another ratio too and I'm sure there are more already I'm not aware of, and more to come in the future...

Edit I clearly need to read better, could the pencil also have the dropdown to populate the custom field (or custom as another option in that or something)?

On a related note, which I suspect wants a separate feature request, it also strikes me you probably want the aspect ratio to be sticky per grid location, e.g. if I'm doing a SD+, I want the touch buttons on each page to preview correctly for that by default, but the other buttons to be "normal", I don't want to have to keep switching ideally, and this gets more and more important as the aspect gets further and further away from 1:1.

Julusian:
> I'd agree we don't want a never-ending list of options on the bar, and personally I'm not convinced having more they'd probably stop being clear what they're for, but could there be some sort of pop-up/drop-down to let you see the original/full list (what does the pencil currently do for example) as from a quick look the Loupedeck should be/can be 90x60, so that's yet another ratio too and I'm sure there are more already I'm not aware of, and more to come in the future...

Yeah I think my ideal solution would be to start storing the sizes/schema of each surface that is connected, which we can then use to populate more options here. So then once you have connected a neo, the option would appear. That would give us the most futureproof behaviour here, and avoids baking in details about every surface type.

But I dont want to do that in this PR, I would like to get this merged before it gets more conflicts.

> On a related note, which I suspect wants a separate feature request, it also strikes me you probably want the aspect ratio to be sticky per grid location, e.g. if I'm doing a SD+, I want the touch buttons on each page to preview correctly for that by default, but the other buttons to be "normal", I don't want to have to keep switching ideally, and this gets more and more important as the aspect gets further and further away from 1:1.

I think it depends. If you are programming a SDS, you wouldnt want to have to choose that for all 32 buttons. And for a plus-xl, I think it could be argued either way.

If we had #2945, then it would make sense to tie to whatever surface type was selected perhaps with an auto-mode. But in the current setup, I'm not sure what is best

PR #4363, opened 2026-07-23
fix: graphics fixes

Julusian:
some misc graphics fixes that need more testing or thinking about


Julusian:
Its been too long since 5.0 for me to like the idea of adding a 'legacy' mode like this. especially as it will only help users/modules which havent yet 'fixed' themselves to work nicely in 5.0.  
And as noone else has given input on it being a good or bad idea, I'm taking that to mean noone cares enough.

The font size change will go into 5.1 now, as I think it still has value but is too late to go into a 5.0 patch

PR #4368, opened 2026-07-26
fix(actions): make Wait a barrier for concurrent action groups

ag-linden:
## Summary

In concurrent action execution, `ActionRunner` started the actions after a Wait as soon as the Wait completed, without awaiting the concurrent group before it. This change awaits each group after launching its actions in parallel, so Wait remains a true barrier while preserving concurrency within each group.

## Verification

- `corepack yarn vitest run companion/test/Controls/ActionRunner.test.ts` — 40 tests passed
- `corepack yarn build:ts` — passed
- `corepack yarn eslint companion/lib/Controls/ActionRunner.ts companion/test/Controls/ActionRunner.test.ts` — passed
- Pre-commit lint-staged ran ESLint and TypeScript checks successfully

The change is limited to first-party action sequencing and has no schema, protocol, or user-data migration.

### DIFF
--- a/companion/lib/Controls/ActionRunner.ts
+++ b/companion/lib/Controls/ActionRunner.ts
@@ -148,8 +148,6 @@ export class ActionRunner {
 		} else {
 			const groupedActions = this.#splitActionsAroundWaits(actions)
 
-			const ps: Promise<void>[] = []
-
 			for (const { waitAction, actions } of groupedActions) {
 				if (extras.abortDelayed.aborted) break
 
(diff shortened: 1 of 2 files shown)


Julusian:
Thanks for the PR, but this behaviour is currently intentional.

While the behaviour of waits in concurrent groups is a little weird generally, it was done this way to match previous behaviour where everything was always run concurrently and each action had its own delay.

Also, aa4a51c474849b6ac01504e0d3cb24620500fd04 is a fix which is pretty much the exact opposite of this PR, showing that this was an explicit choice

ag-linden:
Thanks for the context and the pointer to aa4a51c—that makes the intended scheduling model clear.

I see now that, in a concurrent group, Wait is meant to delay scheduling of later actions without waiting for earlier actions to complete, preserving the previous per-action delay behavior. My change would reverse that compatibility decision by making the wait relative to completion of the preceding group, so closing this makes sense.

The existing test name and comment led me to interpret Wait as a completion barrier. If useful, I’d be happy to submit a small follow-up clarifying that it delays subsequent scheduling but does not await earlier actions.

Issue #4372, opened 2026-07-26
[BUG] Button Text Shrink to fit not working


ChrisSW58:
### Make sure you're on the latest stable or beta build

- [x] I have tested this on the latest stable or beta release

### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion and not a specific module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

If a button has a Feedback with a button text change, the shrink to fit text option does not work as it should and does as in 5.0.2. The text wraps around within the button. Also, you need to specify the text size explicitly within the feedback rather than it picking it up from the Style.

### Steps To Reproduce

Create a feedback based on a button step with different text to what is stated in the Style. 
Make the text size big enough to wrap around

### Expected Behavior

Operate as in 5.0.2 an wrap around based on the text attributes in Style.

### Environment (please complete the following information)

```markdown
- Companion Version: 5.1.0+9736
- OS: MacOS
- Browser: Safari
```

Julusian:
can you give an example? it seems to be behaving for me

ChrisSW58:
Hi there  - so here are the screen shots and code in 5.0.2 - working correctly. 

<img width="643" height="326" alt="Image" src="https://github.com/user-attachments/assets/d9450650-3932-408e-a936-17a53fac1617" />
<img width="643" height="323" alt="Image" src="https://github.com/user-attachments/assets/550b8f21-3807-4efe-8499-78c868ea9912" />
<img width="1110" height="969" alt="Image" src="https://github.com/user-attachments/assets/9fa2381b-881c-4687-8a52-6d457b4059e3" />
<img width="1105" height="693" alt="Image" src="https://github.com/user-attachments/assets/93f5a98d-b95d-4f4a-ba15-9439fe5394a0" />

The code is exactly the same code, imported into the 5.1.0 beta. Here is how it looks on the screen - you can see a number of buttons with wraparound rather than resizing

<img width="639" height="326" alt="Image" src="https://github.com/user-attachments/assets/7e902918-b274-46b2-a20b-5863dd31eef2" />

<img width="639" height="323" alt="Image" src="https://github.com/user-attachments/assets/b8a437f3-b41b-4dde-9ca4-ce1aef7e1a6b" />

Julusian:
[...]

Issue #4359, opened 2026-07-23
[BUG] Various Style topics in v5.0.2


dnmeid:
### Make sure you're on the latest stable or beta build

- [x] I have tested this on the latest stable or beta release

### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion and not a specific module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

I've been testing with v5.0.2 on Windows / Firefox. I think one of this topics is already treated in main. This is a collection of minor bugs and improvements that I hope are not controversial:

- [x] Style: the currently selected element is shared between buttons. When I e.g. select the second box and open another button with two boxes, also the second one will be selected. ~~Each button should remember its selection state and present the last selected element (only per browser of course and not stored in the database)~~ Majority convinced to keep it

- [x] Style: newly created elements should be automatically added above the currently selected element and not on top of the list
 
- [ ] Style: Canvas is fixed on bottom of the list, but drawn on top. I think this is misleading. Actually with this pseudo-element you adjust the decorations and not the canvas. It only indirectly affects the canvas size. Shouldn’t this pseudo-element be called decorations and sit always on top of the list?

- [x] All style elements: there is a section "Layer" with the enabled and the opacity property. I'd say rename this to "Drawing" as layer can be misleading and we already have the first users speaking of layers. Also this section is a good candidate for a draw mode property.

- [x] Feedbacks: the wording "Layered Styles Overrides" is a little misleading because you easily think of layered styles instead of layered overrides. Maybe better "Stacked Style Overrides"

- [x] All style elements: the opacity doesn’t work correct against the canvas. Use a box as the only element. In the preview there is the transparency checkerboard and there everything works as intended. On the real button rendering any opacity greater 0 will show the element with full intensity. There is a slightly change, but that’s not like expected. The expected effect is that the button background itself is determined by the surface and all elements draw with correct opacity on top. For the rendering on a Stream Deck or LCD screen, the background should be assumed black. This background should not be provided automatically by the control, a control should be able to export graphics with transparency. 

- [ ] All style elements: collapsed state of property sections is linked. When I collapse or open e.g. the Border section, it will be done for all elements. In my opinion they should all default to opened, it feels somehow redundant to have the simple layout and then switching to non-simple and then having to open all the sections. This is way to tedious. Also I noticed, that at occasions some sections seem to collapse again. I haven’t found the exact reason yet, maybe when a different element gets added and it’s default state is collapsed.

[...]

Issue #4343, opened 2026-07-21
[BUG] Button preview in style tab is not consistent with grid view


AKmaster1347:
### Make sure you're on the latest stable or beta build

- [x] I have tested this on the latest stable or beta release

### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion and not a specific module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Depending on text lengths and characters, the button preview in the style tab can incorrectly render the button's graphics, moving some text to the next line.

<img width="1674" height="851" alt="Image" src="https://github.com/user-attachments/assets/ce3389b2-1f65-4f1d-87a4-5cfa8bb272b9" />

In this case the emoji style is difference and the text box's width appears to be less than actual.

<img width="744" height="231" alt="Image" src="https://github.com/user-attachments/assets/8442bba9-9e34-4d99-897b-a681f6ecd34d" />

Expanding the width using a different Preview Aspect Ratio expands the width past the cut off point. Note that "\n" are used. Emoji styles remain different.

<img width="1281" height="641" alt="Image" src="https://github.com/user-attachments/assets/cae15833-045b-4959-ad17-a85836e32d91" />

It appears incorrect rendering occurs more frequently when emojis are used.

<img width="886" height="257" alt="Image" src="https://github.com/user-attachments/assets/9a7241e7-cac9-443d-a380-3f9cc524d3f2" />

Plain numbers are unaffected.

### Steps To Reproduce

1. Create a button and add the following text ` "☑️"    : "☒")` to a text element
2. Click on the styles tab and view the generated preview

### Expected Behavior

Style preview and grid display should be the same

### Environment (please complete the following information)

```markdown
Companion: v5.0.2
OS: Windows 11 Home (v10.0.26200; x64)
Browser: Chrome v150.0.7871.128 running OS: Windows 11 (Build 19.0.0) on: desktop
```

Issue #4305, opened 2026-07-12
[BUG] Font Size unit change from Px to % breaks text/glyphs in module presets (5.0)


mm2706:
### Make sure you're on the latest stable or beta build

- [x] I have tested this on the latest stable or beta release

### Is this a bug in companion itself or a module?

- [x] I believe this to be a bug in companion and not a specific module

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Since updating to Companion 5.0, Font Size was changed from Px to percent. The problem: for presets that are defined by modules with a fixed Px size, the conversion doesn't work correctly – the text/glyph is rendered way too large and becomes completely invisible because it overflows the button.
What makes it worse: "Shrink to Fit" is disabled by default, so this doesn't just affect edge cases, it affects basically every impacted preset right after the update. Manually enabling "Shrink to Fit" fixes the display again.
The real issue here: I can't just recreate or adjust the preset myself, since it comes from the module (module-defined preset). So while I can enable "Shrink to Fit" on a per-preset basis, the actual conversion itself needs to be fixed.

### Steps To Reproduce

1. Module Preset with glyph ⏵, Font Size 60 (Px) 
2. In Companion 4.3: displays normally
3. After updating to 5.0/ Or new Instance on 5.0: Font Size becomes 126 (%), glyph disappears completely
4. Only after manually enabling "Shrink to Fit" or adjusting the size by hand does the glyph show up again

### Expected Behavior

The Px to % conversion should render the text at roughly the same visual size as before or "Shrink to Fit" on by Default.

### Environment (please complete the following information)

```markdown
- Companion Version: 5.0-stable
- OS: macOS 26.5.2
- Browser: Safari 26.5.2
```

Julusian:
I think I need some examples to look at for this. Which module? hopefully I dont need to be connected to anything to get the presets? Alternatively, an exported page from 4.3 and a screenshot of that page will suffice

Throwing together a test module, in 4.3.4 I have:

<img width="1263" height="582" alt="Image" src="https://github.com/user-attachments/assets/4b7dc47b-459b-49aa-8cda-b59ae323f27f" />

and 5.0.0:

<img width="1263" height="582" alt="Image" src="https://github.com/user-attachments/assets/9136758f-f3fe-4475-9124-3f08c84f702d" />

for some context, 60pt text is the last one that draws the text and it is only doing so when the topbar is hidden

[...]

PR #4377, opened 2026-07-30
Fix/line breaking

dnmeid:
There seem to be various issues with the line breaking and shrinking and text handling that have accumulated over many versions or are fresh.
Related issues: #3671, #4305, #4343, #4359, #4372, 
I'm trying to get this straight asap.

Some of the bugs seem to be there for quite a while so the question is if this is now still a fix or a change. I guess some presets and styles had been created with the last breaking behavior. I don't know if it is better to say: "now the old stuff works again as expected and you in-between guys are out of luck" or "the existing bug is now the new normal and the old stuff has been broken so long that it doesn't matter anymore".  
Overall it should affect only some rather seldom cases.

- [X] Line breaking will now again first shrink and only break inside a word if there is no shrinking option any more. This will affect a lot of dynamic text in a positive way but should not affect presets or styles that have been carefully tweeked.  (actually that was bugging me for quite a while and I'm so happy to fix this)

- [X] Second difference is the alignment when the test overflows the space. Since quite a while it seems that in case of overflow always the beginning of the text was fitted into the available space and then aligned. This now changes it back to the original concept that also uses the bottom lines or the center lines for bottom or center alignment. I think both variations are not 100% great, the last behavior was maybe 50% nice and the original behavior 60% nice. Given that this shouldn't be a feature that people use on purpose, I think it is acceptable to go back to the original behavior and maybe even improve this a little in the future (not part of the fix).

- [X] Line height measurement is rather exact and line height of 100% still gives no visual shift when alignment changes

- [X] unicode chars are now again preserved when breaking a line

- [x] some unicode chars are not rendered

- [x] style preview and button rendering are using different emoji fonts

- [ ] line height is still fixed to 1.0 and not 1.1 like originally. I think this change is quite old too and we should keep it for now and make this a configurable property eventually

- [ ] a lot of tests are failing now because my cheapo free AI wasn't able to fix them in acceptable time. Also some tests are looking useless to me, had they been checked in deph?

Unfortunately I have based that branch on main. I don't know if it is easy to merge into 5.0 stable


dnmeid:
Ok, here is the current pitfall:
For Color emojis we are using basically Noto Color Emoji font, but not the original version provided by Google because it is not compatible with webkit. Instead we are using a version that has been designed as a replacement for the Apple emoji font. This one is compatible with webkit and as far as we know also with chromium *but not gecko*. It has some more quirks:

- it is using bitmaps (sbix format), that are quite large, don't look nice and scale poorly
- it is overall quite heavy with 40MB
- it is somehow working on Chrome but not recommended
- major quirk: as a replacement for Apple emoji it uses the same typographic measurements as Apple emoji and this leads to a vertical offset of the glyphs (known from #3351) and also a smaller size than Noto

So what are the options?

- Continue using this font: vertical offset of glyphs, no color emoji preview on firefox 
- Adjust the font metrics in the font (my font editor can do that but not export the sbix layers properly): still no preview on firefox
- Generate our own sbix font derived from the official Noto color emoji: will have correct position but breaks sizing as the glyphs are larger and still no preview on firefox
- Generate our own sbix font derived from the official Noto and reduce size: will have correct position and roughly the same glyph size as today but still bad look and big filesize and still no preview on firefox
- Use official Noto: will give best visual quality at chromium and gecko at smaller filesize, but no preview rendering of color emojis in Safari, glyph position is good but glyph size is larger than today
- Use adjusted Noto: will give best visual quality at chromium and gecko at smaller filesize, but no preview rendering of color emojis in Safari, glyph position is good and glyph size is similar to today
- Use different fonts for different engines: everything can work but not a quick fix

Additionally I have to mention that even if we use Noto in the COLR1 version for chromium and gecko, there will be slightly differences because different versions of chromium support different font features and e.g. our skia-canvas does not support black and white variation codes but the chrome browser does (#3671)

What to do?

Julusian:
> Some of the bugs seem to be there for quite a while so the question is if this is now still a fix or a change. I guess some presets and styles had been created with the last breaking behavior. 

yeah im not sure. Cosnidering its 3 weeks after 5.0.0, and 17% of users have upgraded, our aim should be to match 4.3. But changing anything now does worry me about user frustration, but leaving it to 5.1 to fix things that broke in 5.0 will be just as frustrating. So I dont know what is best, other than asap if theyre going to be in 5.0

> Since quite a while it seems that in case of overflow always the beginning of the text was fitted into the available space and then aligned. 

Did I break this when optimising the parsing so that it wouldnt try and chunk a 10_000 character long string into 1000s or lines, to then only draw 3 of them?

Also, if top and middle align show the first lines, should bottom align show the last or first lines? 

> Unfortunately I have based that branch on main. I don't know if it is easy to merge into 5.0 stable

Probably is (cherry pick once merged). Thats how I've been applying things, and this area should still match almost identically.


> Use different fonts for different engines:

This is appealing to me. If we can have chrome and the backend working well and the same for 5.0 then I think that is good enough for today, getting other browsers to work can be the next patch or 5.1 

> Also some tests are looking useless to me, had they been checked in deph?

Yeah a lot of the tests I have mostly left to claude. The ones I have read looked fine, but I havent read all of it (way too time consuming and tedious) and figured that its better than not having any tests.

dnmeid:
>> Use different fonts for different engines

> This is appealing to me. If we can have chrome and the backend working well and the same for 5.0 then I think that is good enough for today

Ok, I'll go ahead with providing a current but size-adjusted font that works for all but Safari. We'll leave this broken for Safari and see what happens.

> Also, if top and middle align show the first lines, should bottom align show the last or first lines?

I did fix this in the first round, but now have decided to stay with the behaviour of 4.3. It seems that it is like that for a long time and maybe the change was intentional. At least I found it a little bit irritating because our default alignment is center/center that the shown part of the text was changing and it is not easy to understand what part is shown.

So the font and test updates will be the last part of this fix.

dnmeid:
After thinking about this now for a day and some experimentation, I have a slightly different pitch:
The color emoji font is super hard to adjust because of the different color layers and their incompatibilities but is provided quite well. It is from the Noto font-family and we are using a lot of Noto fonts because they have a super complete range of glyphs for all script systems. The notably exception is plain latin text where we use Arimo because it is much more compact and has a good readability at the small sizes we are usually dealing with.

We now have a Monospaced font (also Noto) and maybe want to support more fonts in the future. Actually it would be best to not tweak the fonts, but tweak the rendering in the canvas at a font based level. Unfortunately this is not easy to do and also questionable if we should do it for a tool like Companion where typography is not the top priority. Anyhow different fonts should match when used in the same line.

So my current go is to adjust the Arimo font to match all of the other used Noto fonts. That way we always can add and update Noto without tweaking and Arimo should stay quite fixed. To compensate for the wider lines of Noto, I'll introduce a factor that reduces the resulting metrics to the current metrics of Arimo (roughly 0.85). We'll call that our new line height 1.0.
As I have written earlier, to get back the line height from our early versions, we'd need 1.1 and for roughly the original line height of Noto 1.2.

I think that fits nicely in the future strategy, fixes some existing font problems and doesn't break too much.
This will in deed lead to all the Noto fonts being rendered ~6% smaller than today. For Users of the non-latin fonts this could mean a different word wrap of their texts, but they won't loose visible text.

dnmeid:
Always when you think, you're done...

Actually finished with my list with the now new known issue of problematic emoji rendering in preview renderer in WebKit browsers.

Another little thing I noticed but am willing to ignore for now: The preview now shows a tiny bit of shift when changing the alignment of 100% text. The canvas in core does not. I assume some rounding errors.

Not to ignore: The text shrinking doesn't work smooth when the maximum text size is close to 100% but not exactly 100%.

dnmeid:
Getting close, now a bug seems to have (re-)appeared that the browser canvas works different than the node canvas.

Julusian:
@dnmeid hows this going? any portions that are self-contained and worth pulling out and merging before the final bits are done?

I am thinking about looking at the smooth shrink sizes thing (binary search through valid size space), and perhaps a toggle to disable auto line-breaks, but I expect both to majorly conflict with these changes, so I am hoping to wait for this to be done first

Julusian:
I've pulled out I think all of the line breaking and unicode fixes into main now, I don't know how complete the rest is so I shall leave that until this PR is marked as ready.

PR #4381, opened 2026-08-01
Add new aspect ratio option for Stream Deck Neo

peternewman:
Currently entirely untested, but pretty trivial...

## Summary by CodeRabbit

* **New Features**
  * Added a 124:29 aspect-ratio option for Stream Deck Neo previews.


Julusian:
FYI #4341 reworks the area this touches, and I'm not sure how to fit it into the new layout.

I am hesistant to add it as another button in the row as I dont expect this one to be very popular, so am not certain what to do with it

peternewman:
> I am hesistant to add it as another button in the row as I dont expect this one to be very popular, so am not certain what to do with it

I've joined the discussion in the other PR, but it feels like we need to handle other (more niche) aspect ratios ideally, not push the effort onto the user to find out what size a button is etc.

Julusian:
Following #4341 I started playing with this:

image

To drive this we are caching the layout of each surface that has been connected, and using that to influence what options we show here. So as long as companion knows you have a sd-neo, it will offer this size.  
And this means that companion isnt any more aware of these surfaces, and will behave the same for any unofficial/newer surfaces

Julusian:
#4450

PR #4388, opened 2026-08-03
feat: new maths functions

peternewman:
Sin, cos, tan, asin, acos, atan, asinh, acosh, atanh and cbrt functions

## Summary by CodeRabbit

* **New Features**
  * Added autocomplete suggestions for cube-root, trigonometric, inverse trigonometric, and inverse hyperbolic functions.
  * Retained autocomplete support for power functions.
* **Documentation**
  * Added user guide entries with descriptions and usage examples for cube-root, trigonometric, inverse trigonometric, and inverse hyperbolic functions.


Julusian:
could you update the lib too?

I realise that coderabbit here has a point, I didnt think to check whether this worked in degrees or radians. I dont mind which as long as it is clearly stated. 
Given that most of our users are not programmers, degrees probably makes more sense though? And maybe there should be a couple of utils to make the conversion each way trivial?

peternewman:
> could you update the lib too?

I assume you mean pull in v1 of the lib into shared-libs?
 
> I realise that coderabbit here has a point, I didnt think to check whether this worked in degrees or radians. I dont mind which as long as it is clearly stated. Given that most of our users are not programmers, degrees probably makes more sense though?

Given it seems the JS functions work in radians, and to avoid even more confusion for some users, aren't we better off leaving the functions as-is and then adding a helper function to convert either way, rather than making these non-standard to simplify it for some users?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sin

> And maybe there should be a couple of utils to make the conversion each way trivial?

To me making it clear on the units and having these functions feels safer, as that will also be the way those functions work on a calculator etc.

peternewman:
Shall we just use these functions and naming?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math#converting_between_degrees_and_radians

Julusian:
> I assume you mean pull in v1 of the lib into shared-libs?

yes

> Given it seems the JS functions work in radians, and to avoid even more confusion for some users, aren't we better off leaving the functions as-is and then adding a helper function to convert either way, rather than making these non-standard to simplify it for some users?

:shrug: 

All I know right now is that for element angles, we are using degrees, and the other functions are js inspired, but not strictly js. 

Another idea is perhaps these should gain a second parameter to specify which unit. but that might be too with a conversion util..

PR #4390, opened 2026-08-03
feat: streaming json parsing/generation

### LINKED ISSUE #3497: Rework exports to avoid issues with max string length

### Describe the feature

There have been issues for some users in extreme cases where their [config was too large to store as a strin](https://github.com/bitfocus/companion/issues/2980)g, causing companion to fail to open it. To mitigate this, the config was converted to be a sqlite database

This resolves the issue for configs, but a similar problem could still occur for config exports, we should start thinking about how we want to handle this, to try to have a solution in place before users hit this issue.  
With the graphics overhaul, I think it likely that users will start adding larger images, and potentially more of them, making it more likely they will hit this issue. 

I see a couple of options:
1) Change the export format to be a sqlite database. 
    This has the downside of making it less user friendly to manipulate, and it will be a binary format.
2) Change the export to be a zipped folder of files
    This keeps it more user editable, with the risk of potentially hitting this issue if the structure is not thought about carefully
3) Make the full export be a full copy of the db
    A different approach, but means that making an export is simpler and might make it easier to ensure we preserve everything? Im not sure if this is a good idea or not.

Both of these will mean reworking the formats we offer for exports, but to differing degrees.  
But in both cases we will avoid the string length issue. For the zipped folder approach, I would be tempted to do it so that roughly each row we have in the sqlite db currently translates to its own file. This will mean hundreds or thousands of files in the zip, and some hassle to work with, but it ensures we wont have the length issue.

### Usecases

_No response_

Julusian:
Closes #3497

This has the limitation of not supporting streaming for yaml files. And to preserve the pretty json formatting, we have to use the normal JSON.stringify, so when uncompressed will only stream when too large to do the simple way

## Summary by CodeRabbit

* **New Features**
  * Import files up to 1 GiB, including compressed files with decompressed-size limits.
  * Large JSON and compressed JSON exports can now be streamed for improved reliability.
  * Backup exports are written more safely, reducing the risk of incomplete files.

* **Bug Fixes**
  * Improved detection and validation of JSON, YAML, malformed, empty, and unsupported imports.
  * Oversized or corrupted files now return more consistent error messages.
  * Export failures are handled more reliably during downloads and file creation.


jswalden:
For what it happens to be worth -- not terribly much, obviously! -- `@jswalden/streaming-json` implements prettyprinting support in its streaming stringification.  (As a bonus, I'm sure the socket-security bot's dependency report on it would be stunning.  :sweat_smile:)

PR #4415, opened 2026-08-15
fix(variables): synchronize blinker ticks with epoch phase to prevent clock drift

### LINKED ISSUE #4292: [BUG] Multiple Blink Buttons not synchronous

### Describe the bug

using the blink function with the same arguments across multiple buttons creates buttons blinking asynchronously.

### Steps To Reproduce

In a fresh config
Create a simple button with a boolean based feedback
0 == 0 && blink (1000)
copy and paste the button one or more times

buttons will not blink synchronously

Image

### Expected Behavior

Expected synchronous blinking using same arguments 

```markdown
- Companion Version: 4.3.4
- OS: Windows 11
- Browser: Edge Version 149.0.4022.98 (Official build) (64-bit)
```

### Additional context

_No response_

snowyukitty:
## Summary
`VariablesBlinker` previously scheduled ticks using relative delays (`onPeriod` / `offPeriod`), which caused timer drift to accumulate over time. This caused multiple blink buttons with the same or harmonically related intervals to desynchronize over long runtimes, and caused newly registered intervals to start unaligned with existing intervals.

This PR:
1. Derives initial blinker state and first transition timing directly from `Date.now() % interval` relative to Unix epoch.
2. Recalculates each subsequent tick delay against the absolute epoch phase to eliminate drift accumulation.
3. Implements bounded synchronization recovery under slow listener callbacks without event-loop starvation.
4. Adds unit test coverage verifying phase alignment, asymmetric duty cycles, and slow callback recovery.

Fixes #4292

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blinker timing by aligning transitions with the current cycle phase and state.
  * Preserved correct behavior for continuously on or off configurations.
  * Improved recovery when slow callbacks cause timing delays.
  * Prevented unnecessary emissions and maintained accurate transition timing after catch-up.

* **Tests**
  * Added coverage for phase alignment, delayed callbacks, recovery behavior, and bounded catch-up processing.

### DIFF
--- a/companion/lib/Variables/VariablesBlinker.ts
+++ b/companion/lib/Variables/VariablesBlinker.ts
@@ -55,11 +55,15 @@ export class VariablesBlinker {
 
 		this.#logger.debug(`Starting new blinker interval: ${onPeriod}ms/${offPeriod}ms`)
 
+		const now = Date.now()
+		const phase = now % interval
+		const isCurrentlyOn = onPeriod > 0 && phase < onPeriod
+
 		const newEntry: BlinkingInterval = {
 			interval: interval,
 			onPeriod: onPeriod,
 			offPeriod: offPeriod,
-			lastProbed: Date.now(),
+			lastProbed: now,
 			name: {
 				variableId: `internal:__interval_${onPeriod}_${offPeriod}`,
 				label: 'internal',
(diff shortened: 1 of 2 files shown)


snowyukitty:
Closing after rechecking the issue discussion and contribution requirements. This changes the already shared blinker clock, while the maintainer-identified desynchronization is downstream in rendering, so it does not address the reported root cause. Sorry for the noise.

PR #4423, opened 2026-08-20
fix: make folding sidebar touch interactions reliable

### LINKED ISSUE #4361: [BUG] Sidebar not touch friendly

### Describe the bug

When the sidebar is dynamic and you click on an item, it goes back to the narrow state, even when you stay over the sidebar. This is good, but it doesnt work for touch entry. When you touch an item the sidebar stays expanded.

### Steps To Reproduce

_No response_

### Expected Behavior

Sidebar shoud minimize also after touching an item.

```markdown
- Companion Version: 5.0.2
- OS: Win 11
- Browser: Firefox
```

### Additional context

_No response_

mvanhorn:
In desktop folding mode, choosing a sidebar navigation item is meant to return the dynamic sidebar to its narrow state. `SidebarRoot` currently recognizes selections through a native `mouseup` listener, so touch input depends on browser-generated compatibility mouse events and behaves inconsistently in Firefox on Windows. The reporter has observed the sidebar remaining expanded, briefly expanding and collapsing, or staying collapsed after otherwise equivalent touches.

In desktop folding mode, completing a primary touch pointer interaction on a navigation link applies the deferred temporary narrow state and leaves route selection behavior intact.

## Summary by CodeRabbit

Update `SidebarRoot` to handle completed primary pointer interactions directly so mouse, touch, and pen navigation share one fold-after-selection path without relying on synthetic mouse events. Preserve the existing exclusions for secondary-button context menus, group toggles, footer controls, blank areas, and mobile-mode dismissal, including the deferred state update that avoids racing route navigation. Account for non-hover touch pointers re-entering a temporarily narrowed folding sidebar so the explicit narrow state does not prevent a subsequent touch from expanding and selecting another item.

Closes #4361

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sidebar interactions across touch, mouse, and pen input.
  * Mobile sidebar dismissal and narrowed-sidebar folding now respond more reliably.
  * Touch interaction temporarily expands a narrowed sidebar before navigation.
  * Secondary clicks and non-navigation controls no longer trigger unintended sidebar changes.

* **Tests**
  * Added comprehensive coverage for sidebar navigation, folding, touch expansion, and mobile behavior.

### DIFF
--- a/webui/src/Layout/Sidebar.tsx
+++ b/webui/src/Layout/Sidebar.tsx
@@ -648,9 +648,9 @@ function SidebarRoot({
 		if (mobileMode) setVisibleMobile(false)
 	}, [mobileMode])
 
-	// handle clicks in the sidebar for mobile mode and "unfolding" mode
-	const handleOnClick = useCallback(
-		(event: MouseEvent) => {
+	// handle completed pointer interactions in the sidebar for mobile mode and "unfolding" mode
+	const handlePointerUp = useCallback(
+		(event: PointerEvent) => {
 			const target = event.target
 			// note: middle-click currently opens the nav-link target in a new tab, so it makes sense to close the sidebar in that case.
 			// Only context-menu should leave the sidebar alone, since it is acting on the current sidebar, hence "event.button === 2".
@@ -679,6 +679,15 @@ function SidebarRoot({
 		[setNarrow, mobileMode, unfoldable]
 	)
 
+	// Touch pointers do not hover, so entering a temporarily narrowed folding sidebar must explicitly expand it.
+	// The subsequent pointerup will schedule the fold again if a navigation item is selected.
+	const handlePointerEnter = useCallback(
+		(event: React.PointerEvent<HTMLDivElement>) => {
+			if (narrow && event.pointerType === 'touch') setNarrow(false)
+		},
+		[narrow, setNarrow]
+	)
+
 	// if in "temporary narrow-mode" return to folding mode after the mouse leaves the sidebar
 	// note that in "permanent" narrow-mode, setNarrow is passed as a no-op, so this callback is active only when not in narrow-mode
 	const handleMouseLeave = useCallback(() => {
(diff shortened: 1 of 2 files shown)


Julusian:
As I just posted in the linked issue, I dont think im able to reproduce the issue this is trying to fix.

But with this fix, I am now finding that any tap on the sidebar is causing it to expand then collapse, which is not expected.

mvanhorn:
Thanks, and sorry for the regression. You are right about the tap behavior, and I think it points at something I got wrong.

The `pointerenter` handler was mine: touch pointers do not hover, so I added it to expand a temporarily narrowed sidebar on tap. But that same tap then fires `pointerup`, which schedules the fold, so a tap expands and immediately collapses. That is exactly what you are seeing, and it is not fixable by tuning the timing. The handler should not be there.

Removing it leaves the part of this PR I still believe in: switching the sidebar from `mouseup` to `pointerup` so touch and pen selections register at all, which is the actual mechanism behind the original report.

Before I push anything, though: you said you cannot reproduce the original issue. If it does not reproduce for you, I would rather you tell me whether this PR is worth keeping at all. I am happy either way. If you want it, I will push the `pointerenter` removal and you can re-test the tap behavior. If you would rather close it, that is a completely reasonable outcome and I will not be precious about it.

One caveat if you do want it: I could not run the sidebar tests locally, my clone has a broken jsdom `localStorage` that fails all six tests identically with and without the change, so CI would be the real signal.

Julusian:
Its hard to say.

It sounds like it solves something for you meaning it has value, but without my being able to reproduce it, it is going to be hard to merge as I cant be sure it does what it claims and cant be sure it doesnt break anything else.

mvanhorn:
Closing this. Julusian, you're right to be uneasy — I can't reproduce #4361 either. It's specific to Firefox on Windows relying on browser-generated compatibility mouse events, I don't have that combination, and the original report isn't mine. Asking you to merge a 227-line touch-handling change that neither of us can verify isn't a fair trade for the regression risk you already found.

The diagnosis in my earlier comment stands if anyone picks this up later: the pointerenter handler expands a temporarily-narrowed sidebar on tap, and the same tap's pointerup then schedules the fold, which is the expand-then-collapse you saw. Thanks for taking the time to build it and test it.

Issue #4324, opened 2026-07-15
Feature Request: Faster way to edit basic text of a button


tim-sparkstreet:
### Is this a feature relevant to companion itself, and not a specific module?

- [x] I believe this to be a feature for companion, and is not specific to a module

### Is there an existing issue for this?

- [x] I have searched for similiar existing issues

### Describe the feature

It would be very helpful for my company's workflows (and I'm sure many others who will soon be transitioning to v5.0.0+ and do any work that requires often changing button Text labels) to be able to quickly edit the "Button text string" of the default "Text" element on a button's "Style" tab.

Several possible options in (likely) increasing order of difficulty:

1. Automatically expand the "Text" element when creating a new button or selecting the "Style" tab of an existing button.

Currently when clicking on the "Style" tab, you have to "Select an element from the list above to edit its properties" which I feel like could instead default to automatically selecting and expanding the topmost element in the list, which in most cases would be the "Text" element. I imagine the vast majority of the time most people will be selecting "Text" to edit first anyway, and if not then they're going to have to mouse over and click a different element anyway so at worst it inconveniences nobody and at best saves a click for most.

This alone would halve the extra mouse moves/clicks needed from old behavior and I don't see a downside to implementing it.

2. Move the "Style" tab (or add an option to move) to be physically separated from the "Step", "Feedback", etc. tabs and place it where the "Notes" field currently is (and where it used to be in previous versions). New Buttons should also automatically open to "Step 1" like they used to instead of adopting the previously selected tab from an existing button.

Notes could become a tab with "Style", separating the visual configurations from the functional tabs. It could also just move down to sit in between "Style" and the "Step, etc." tabs.

The current behavior (in v5.0.0+) is that the first screen that opens up after creating a new button is whatever tab you were selected on with a different button, which can vary. I understand the rationale for this when switching between existing buttons (changing step behavior, feedback, or style on several buttons in sequence), but it doesn't seem intuitive for new button behavior.

Here's the old screen that comes up when creating a new button, which is the equivalent of now having "Style" and "Step 1" open at the same time, stacked vertically:
<img width="943" height="689" alt="Image" src="https://github.com/user-attachments/assets/162b9b8f-6c1f-48c4-a2da-14737091db4c" />

[...]

PR #4442, opened 2026-08-31
feat: pinned properties for the layered button style editor #4324

Julusian:
A new pitch to solve #4324 

I think the simple/advanced toggle was intended to solve this simpler. This replaces that with an ability to 'pin' any property, which adds it to this list of pinned properties as a quick access.  
By default we pin the subset of properties that matches the old drawing 

image

## Summary by CodeRabbit

* **New Features**
  * Layered button editors now support pinning individual properties in a dedicated Pinned view.
  * Added pin/unpin controls for supported elements, including nested elements and grouped property sections.
  * Added default pinned properties for supported element types.
  * Added an inline button style variant for compact controls.

* **Improvements**
  * Updated editor layout with clearer Pinned and Canvas entries and a more compact resizable divider.
  * Existing projects and imports now preserve and populate pinned-property settings.
  * Property panels now consistently use collapsible sections with expand/collapse-all support.

* **Documentation**
  * Updated button styling guidance for the Pinned view and pinning workflow.


krocheck:
Overall I like the approach here and that there's default pins per object type / layer.

But there's some aspects I think become confusing for new users.  E.g. because I can collapse the pin groups, by default I'll see all the layers.  It's not obvious now you click the layers in the sorter to get all of the properties.  Not sure how you solve that.  This is, in part, because the Pinned & Canvas buttons look like a self-contained tab group but they are in-fact grouped with the layer "buttons" as well.
image
I'll add that I'm not a fan that those buttons highlight red when selected but the layers only highlight a darker shade of gray.  Not that they need to highlight red when selected, but all this together makes it feel like a band-aid approach.

Otherwise the functionality the pin/unpin and how the groups show up is nice.  I'm wondering if this interface needs expand all / collapse all like actions has now (but lord knows where to put them).

I'm also not sure with this added pinning functionality and tab what the default when clicking between buttons should be.  Right now the behavior is that if I'm selected on a Text layer, when I click another button I get dropped into the top Text layer.  Ditto for box, etc.  We might want to always default to Pinned when clicking between buttons ... especially since it persistently remembers which groups in pinned are collapsed or not.

Separately ... I find myself accidentally clicking the add step + in this UI instead of the add layer + at times.  Also unclear how to solve that

Julusian:
>  It's not obvious now you click the layers in the sorter to get all of the properties. Not sure how you solve that. 

There were some shortcut links in those pinned headings but they appear to have gotten lost (or maybe I didnt like them for some reason)

>  This is, in part, because the Pinned & Canvas buttons look like a self-contained tab group 

Yeah im not sure what to do about this. Putting pinned in the stack of layers will be weird, and someone commented on slack this week that they didnt realise the canvas in the layer stack was selectable (prior to this).

I dont know how better to reflect that these buttons and the elements themselves are all selectbale and change the same content in the lower panel.

>  I'll add that I'm not a fan that those buttons highlight red when selected but the layers only highlight a darker shade of gray. Not that they need to highlight red when selected, but all this together makes it feel like a band-aid approach.

Sure. This followed the 'primary' button colour we have (eurgh that primary red), but I see no reason they couldnt be grey too.

> Otherwise the functionality the pin/unpin and how the groups show up is nice. I'm wondering if this interface needs expand all / collapse all like actions has now (but lord knows where to put them).

In some of the other panels I recently added the alt+click to expand/collapse all. probably should match that.

> I'm also not sure with this added pinning functionality and tab what the default when clicking between buttons should be. Right now the behavior is that if I'm selected on a Text layer, when I click another button I get dropped into the top Text layer. 

I'm hesitant to change that behaviour just yet, as that persistence was added in a 5.0 patch release. I dont know which behaviour is best, I can see that both are good at different times.

> Separately ... I find myself accidentally clicking the add step + in this UI instead of the add layer + at times. Also unclear how to solve that

Sounds like something for a larger UI/design overhaul ;)

PR #4450, opened 2026-09-03
feat: better custom preview aspect ratios

Julusian:
Replaces #4381

Shortcuts for more aspect ratios for different surfaces in the style editor.

There isn't enough space on the bar to show everything the user could end up with, and if that list becomes dynamic they will need some better way to identify them.

So instead, we add some options to the bottom of the popover:

image

This uses data sourced from the 'known' surfaces (the ones both connected and disconnected in the surfaces table).
To do this we are now storing the full surface layout definition they provide, which we can use for more in the future. So this will start working after Companion next sees those surfaces.  

So this will work for any connected surfaces that use different aspect ratios, not just official ones. Anything connected over satellite will appear here too.

## Summary by CodeRabbit

- **New Features**
  - Added surface layout and button-size information for connected and offline surfaces.
  - Added live updates for surface layouts and available button dimensions.
  - Added an aspect-ratio picker with presets, custom width/height values, and ratios derived from connected surfaces.
  - Surface layouts now persist with surface configuration.

- **Improvements**
  - Aspect-ratio controls are organized into a dedicated, more compact picker beneath the layered button preview.

### DIFF
--- a/companion/lib/Surface/Config.ts
+++ b/companion/lib/Surface/Config.ts
@@ -79,6 +79,7 @@ export function createOrSanitizeSurfaceHandlerConfig(
 		type: panel.info.description || 'Unknown',
 		integrationType,
 		gridSize: panel.gridSize,
+		layout: panel.surfaceLayout,
 	}
 
 	// Forget old values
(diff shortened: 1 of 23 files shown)


bryce-seifert:
Just saw this and wanted to say it looks great! Nice elegant solution.

